Issue #8, opened 2018-01-11
Toy Model


mortonjt:
Need a simple model that

1. Calculates the coordinates of the nodes given an `skbio.TreeNode` object.
2. Calculates all of the parent/child relationships
3. Assigns metadata (i.e. edge color and node size)
4. Outputs a node metadata `pd.DataFrame` and an edge metadata `pd.DataFrame`

See the [coords](https://github.com/biocore/gneiss/blob/master/gneiss/plot/_dendrogram.py#L89)  method in Gneiss.  It may make more sense to output the parent node rather than the children, since every node will have a parent (except for the root node).

htoo97:
Default layout: radial

Issue #7, opened 2018-01-11
Toy viewer


mortonjt:
Need to create a simple view model that

1. Accepts edge metadata `pd.DataFrame`
2. Accepts node metadata `pd.DataFrame`
3. Plots the results onto a canvas using matplotlib

YimengYang:
For now, it only plots the nodes and edges without any attribute like color. We might need to discuss the data format of attributes?

mortonjt:
For plotting colors, sizes, ..., I think it is reasonable to have those specified as a separate column.

So for instance, if you had a method named, `color_nodes` - that could be specified as follows

```python
def color_nodes(node_metadata, color_column='Disease_Type', ...):
   """
   Parameters
   ------------
   node_metadata : pd.DataFrame
      Contains the metadata + attributes, where the row names are the node ids
      and the column names are the attributes to be plotted
   color_column : pd.DataFrame
      The column that specifies the colors for each node
   ...
   """
   ...
```
Of course, we may have another name for the view method.  But the interface should be flexible enough, where it can specify the attributes to be plotted on the fly.

With respect to designing a visualization API, I think it makes sense to follow the standards followed in [Seaborn](https://seaborn.pydata.org/examples/grouped_boxplot.html).  We can discuss API implementation details in the future.

YimengYang:
I looked at Gneiss and is heatmap a view? Will view be an object?

PR #16, opened 2018-01-18
Toy model view

kwcantrell:
This is resolving issue #7 and #8


phyloviz/ToyModel.py
```diff
@@ -0,0 +1,353 @@
+import abc
+from collections import namedtuple
+from skbio import TreeNode
+import pandas as pd
+import numpy as np
+class ToyModel(TreeNode):
```
mortonjt: In order to reduce code redundancy, I suggest

1. Move this into model.py
2. Rename this to Model, rather than ToyModel.

phyloviz/ToyModel.py
```diff
@@ -0,0 +1,353 @@
...
+        
+        #stores (x,y) of all leaf nodes
+        points = []
+        
+          
+        #Dirty attempts to simulate the recurisive method found in Gniess
+        #Needs to be cleaned up
+        #Note: This function assumes a node can have n-children
+        #      I believe we are only going to use bifurcating trees
+        #      which means this code can be cleaned up quite a bit
```
mortonjt: Right, I'd suggest cleaning this up.  I think the [preorder](http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.preorder.html#skbio.tree.TreeNode.preorder) method will help simply this, as it can emulate a stack to perform depth-first search.

phyloviz/model.py
```diff
@@ -0,0 +1,415 @@
...
+                pass
+
+    else:
+        # return error message file format cannot be parsed
+        pass
+
+    pass
+
+
+class Model(object):
```
mortonjt: Let's move all of the class methods within `ToyModel` into here.

phyloviz/model.py
```diff
@@ -0,0 +1,415 @@
...
+        # viewcoords.
+        self.pan = np.array()
+
+        # Mask specific coordinates not to display.
+        # TODO: should this involve the resolution
+        # handling?
+        self.mask = np.array()
+
+    # Coordinate manipulation
+    def layout(self, layout_type):
```
mortonjt: We can move the contents within the coords method here.

phyloviz/view.py
```diff
@@ -0,0 +1,42 @@
+import pandas as pd
+import matplotlib
+import numpy as np
+import matplotlib.pyplot as plt
+
+def plot(node_metadata, edge_metadata):
```
mortonjt: Can we have this method accept an axes object?  Preferably something similar to the standards in seaborn (see the boxplot example [here](https://seaborn.pydata.org/generated/seaborn.boxplot.html#seaborn.boxplot))

The advantage of doing this is that we can enable multiple subplots -- we can embed trees alongside heatmaps, boxplots, scatter plots, and many other useful visualizations.

phyloviz/view.py
```diff
@@ -0,0 +1,42 @@
...
+       Contains all of the edge attributes.
+       Every row corresponds to a unique edge
+       and every column corresponds to an attribute.
+
+    """
+    # Plot nodes
+    col_node_x = node_metadata['x']
+    col_node_y = node_metadata['y']
+    # TODO: annotation on points
+    plt.scatter(col_node_x,col_node_y)
```
mortonjt: I suggest swapping this out with an [matplotlib Axes object](https://matplotlib.org/api/axes_api.html).  This will not only make the code more modular, but it will allow for multiple plots to coexist (plt is a global variable and should be avoided when making visualization libraries).

phyloviz/view.py
```diff
@@ -0,0 +1,42 @@
...
+    node_metadata : pd.DataFrame
+       Contains all of the species attributes.
+       Every row corresponds to a unique species
+       and every column corresponds to an attribute.
+       Metadata may also contain ancestors.
+    edge_metadata : pd.DataFrame
+       Contains all of the edge attributes.
+       Every row corresponds to a unique edge
+       and every column corresponds to an attribute.
+
```
mortonjt: I'd recommend returning an axes object here

phyloviz/view.py
mortonjt: Along the same lines, I'd suggest looking at the [boxplot API](https://seaborn.pydata.org/generated/seaborn.boxplot.html#seaborn.boxplot) to see how colors are specified.

For example, it would be extremely useful if the following API were available

```python
def plot(node_metadata, edge_metadata, 
              node_color=None, node_size=None, 
              node_alpha=None, edge_color=None, 
              edge_width=None, edge_alpha=None):
    """
    Parameters
    -----------
    node_metadata : pd.DataFrame
        Contains all of the species attributes.
        Every row corresponds to a unique species
        and every column corresponds to an attribute.
        Metadata may also contain ancestors.
    edge_metadata : pd.DataFrame
        Contains all of the edge attributes.
        Every row corresponds to a unique edge
        and every column corresponds to an attribute.
    node_color : str
        Name of column in `node_metadata` to plot node colors.  If None, all nodes will be colored black.
    node_size : str
        Name of column in `node_metadata` to resize the nodes.  If None, all nodes will be 10.
    node_alpha : str
        Name of column in `node_metadata` to specify transparency of nodes.  If None, the nodes won't be transparent.
    edge_color : str
        Name of column in `edge_metadata` to plot edge colors.  If None, all edges will be colored black.
    edge_size : str
        Name of column in `edge_metadata` to resize the nodes.  If None, all edge will have a width of 1.
    edge_alpha : str
        Name of column in `edge_metadata` to specify transparency of edges.  If None, all edges won't be transparent.
    """

```

PR #19, opened 2018-02-01
Toy model view

YimengYang:


.gitignore
```diff
@@ -99,3 +99,7 @@ ENV/
 
 # mypy
 .mypy_cache/
+<<<<<<< HEAD
```
mortonjt: watch out for this.

phyloviz/view.py
```diff
@@ -0,0 +1,141 @@
...
+    """ Updates the edge metadata of viewer object
+    by parsing the json object with index orientation
+    """
+    edge_json = request.get_json()
+    edge_metadata = pd.read_json(edge_json, orient='index')
+
+
+# Run Flask app
+if __name__ == '__main__':
+    app.run(host=LOCALHOST, port=VIEW_PORT, debug=True)
```
mortonjt: I suggest to create a separate webserver.py to house the controller see [here](https://github.com/jdereus/labman/blob/master/labman/gui/webserver.py)

phyloviz/view.py
```diff
@@ -0,0 +1,141 @@
...
+from flask import Flask
+from flask import request
+
+# Constants for REST API
+MODEL_PORT = '9001'
+VIEW_PORT = '9002'
+LOCALHOST = '127.0.0.1'
+
+# Set up REST API for view
+app = Flask(__name__)
```
mortonjt: Excited to see this in action!  May want to switch the restiful interface over to Tornado -- especially since Tornado can handle more requests simultaneously

mortonjt:
Looks like the .ipynb files are broken.  I'd suggest removing them from this pull request.
And the extra `<<<HEAD` line from the .gitignore needs to be removed.

Otherwise, I think this is good to merge, once the pep8 / flake8 errors are corrected!

Issue #41, opened 2018-07-06
Need to make reading meta-data file more general.


kwcantrell:

mortonjt:
One thing we should try to base the metadata format off of is the qiime2 format

https://docs.qiime2.org/2018.6/tutorials/metadata/

PR #47, opened 2018-07-25
Refactor with cli

kwcantrell:
Added some tests, refactored JS code, and added CLI


phyloviz/model.py
```diff
@@ -56,7 +54,9 @@ def read_internal_node_metadata(file_name):
        pd.Dataframe
 
     """
-    metadata = pd.read_table(file_name, skiprows=3, names=["Node_id", 'label'])
+
+    metadata = pd.read_table(file_name, skiprows=3)
```
mortonjt: TODO: need to make skiprows an option in the future

phyloviz/model.py
```diff
@@ -56,7 +54,9 @@ def read_internal_node_metadata(file_name):
...
+    metadata.rename(columns={metadata.columns[0]: "Node_id"}, inplace=True)
```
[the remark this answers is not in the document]
mortonjt: Actually, I'm not so concerned about this.  We will readdress this causes problems downstream

phyloviz/model.py
mortonjt: Let's make skiprows a CLI option for this PR before merging.

Will want to follow up on the format a little more closely for a followup PR: https://github.com/biocore/empress/issues/41

phyloviz/static/init_webpage.js
```diff
@@ -7,53 +7,49 @@ function initWebPage(edgeMetadata) {
...
 function extractEdgeInfo(edgeMeta) {
-  console.log("Start Extracting");
-  const extractedFieldsMultiDimensional = edgeMeta.map(function(edge) {
-    return [edge.px, edge.py, edge.branch_color, edge. x, edge.y,
+  let extractedFields = edgeMeta.map(function(edge) {
+    return [edge.px, edge.py, edge.branch_color, edge.x, edge.y,
             edge.branch_color];
   });
-  const extractedFieldsFlattend = [].concat.apply([], extractedFieldsMultiDimensional);
+  extractedFields = [].concat.apply([], extractedFields);
```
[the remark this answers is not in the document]
mortonjt: Don't forget this one

phyloviz/tests/test_controller.py
```diff
@@ -0,0 +1,41 @@
...
+# Distributed under the terms of the Modified BSD License.
+#
+# ----------------------------------------------------------------------------
+import os
+import filecmp
+from phyloviz.tests.test_handler import TestHandler
+from unittest import main
+
+
+class TestControll(TestHandler):
```
mortonjt: TestControll -> TestController

phyloviz/tests/test_controller.py
```diff
@@ -0,0 +1,41 @@
...
+            lines = f.readlines()
+
+        with open('model_html_exp.txt', 'w') as f:
+            lines = list(map(lambda x: x.lstrip(), lines))
+            f.writelines(lines)
+
+        self.assertTrue(
+            filecmp.cmp(
+                'model_html_res.txt',
+                'model_html_exp.txt'))
```
mortonjt: Nice test!

phyloviz/tests/test_controller.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_edge_handler(self):
```
mortonjt: Do you have a test for this?  Otherwise we'll want to raise an issue to remind ourselves

phyloviz/tests/test_model.py
```diff
@@ -1,263 +1,241 @@
...
+            0: [(332.627 - 750), (450 - 450), (37.5 - 750), (154.873 - 450)],
+            1: [(332.627 - 750), (450 - 450), (37.5 - 750), (745.127 - 450)],
+            2: [(1167.37 - 750), (450 - 450), (1462.5 - 750), (745.127 - 450)],
+            3: [(1167.37 - 750), (450 - 450), (1462.5 - 750), (154.873 - 450)],
+            4: [(750 - 750), (450 - 450), (332.627 - 750), (450 - 450)],
+            5: [(750 - 750), (450 - 450), (1167.37 - 750), (450 - 450)]},
+            index=['px', 'py', 'x', 'y']).T
+        self.tree.center_tree()
+        center_res = self.tree.edge_metadata[['px', 'py', 'x', 'y']].copy()
+        # center_res = center_res[:-1]
```
mortonjt: will want to remove this

phyloviz/model.py
```diff
@@ -15,12 +13,12 @@ def name_internal_nodes(tree):
     Returns
     -------
     """
-    print("start labelling nodes")
     # initialize tree with branch length
     for i, n in enumerate(tree.postorder(include_self=True)):
         if n.length is None:
             n.length = 1
-        if not n.is_tip() and n.name is not None:
+        # if not n.is_tip() and n.name is not None:
```
qiyunzhu: This comment can be removed?

phyloviz/model.py
```diff
@@ -290,20 +288,17 @@ def update_edge_category(self, attribute, category,
         updated version of edge metadata
         """
 
-        edgeData = self.edge_metadata
         if lower is not "":
-            edgeData[category] = edgeData[category].mask(edgeData[attribute] >
-                                                         float(lower),
-                                                         new_value)
+            self.edge_metadata[category] = self.edge_metadata[category].mask(
```
qiyunzhu: I think the Pythonic way of check if string is empty is not using `is not ''`. Instead, it should be `if xxx != ''`, or simply `if xxx`. Meanwhile, `is not` is for values like `None`. Not 100% sure @mortonjt .

phyloviz/model.py
```diff
@@ -343,8 +336,7 @@ def collapse_clades(self, sliderScale):
 
                 # if parent was visible
                 # add triangle coordinates to dataframe
-                if self.edge_metadata.loc[self.edge_metadata['Node_id'] ==
-                   node.parent.name, 'visibility'] is True:
+                if (self.edge_metadata.loc[self.edge_metadata['Node_id'] == node.parent.name, 'visibility'] is True):
```
qiyunzhu: Is there always only one row meeting this criterium? Is this `Node_id` not the index of the DataFrame? Also, a simpler way of doing this is: 
```
if self.edge_metadata.query('Node_id == "%s"' % node.parent.name)['visibility']:
```

phyloviz/static/init_webpage.js
```diff
@@ -7,53 +7,49 @@ function initWebPage(edgeMetadata) {
...
   for (let property in templateMetadata) {
-    if (templateMetadata.hasOwnProperty(property)) {
-      if (!($.inArray(property, exclude) >= 0)) {
-        let option = document.createElement("option");
-        option.text = property;
-        option.label = property; //TODO: check to see if property is numeric or categorical
-        x.add(option);
-      }
+    if (!($.inArray(property, exclude) >= 0)) {
+      let option = document.createElement("option");
```
qiyunzhu: A simpler way is (please test in case I am wrong):
```
x.add(document.createElement("option", {"text": property, "label": property}));
```

phyloviz/static/init_webpage.js
```diff
@@ -7,53 +7,49 @@ function initWebPage(edgeMetadata) {
...
 /*
- * Extracts the coordinates/other info from metadata and formats it for webgl
+ * Extracts the coordinates/color info from metadata and format it for webgl
  */
 function extractEdgeInfo(edgeMeta) {
-  console.log("Start Extracting");
-  const extractedFieldsMultiDimensional = edgeMeta.map(function(edge) {
-    return [edge.px, edge.py, edge.branch_color, edge. x, edge.y,
+  let extractedFields = edgeMeta.map(function(edge) {
+    return [edge.px, edge.py, edge.branch_color, edge.x, edge.y,
```
qiyunzhu: If these keys are in order, then this function can simply be `return edge.values()`.

phyloviz/tests/test_controller.py
```diff
@@ -0,0 +1,41 @@
...
+        with open('model_html_res.txt', 'w') as f:
+            f.write(response.body.decode('utf-8'))
+
+        # taken from https://codereview.stackexchange.com/questions/145126
+        # /open-a-text-file-and-remove-any-blank-lines
+        with open(os.path.join(os.path.dirname(__file__), '../tree_with_webgl.html'), 'r')as f:
+            lines = f.readlines()
+
+        with open('model_html_exp.txt', 'w') as f:
+            lines = list(map(lambda x: x.lstrip(), lines))
```
qiyunzhu: Why this line is here instead of two lines above?

phyloviz/tests/test_handler.py
```diff
@@ -0,0 +1,46 @@
...
+# ----------------------------------------------------------------------------
+import os
+from phyloviz.webserver import Application
+from tornado.testing import AsyncHTTPTestCase
+from phyloviz.model import Model
+
+
+class TestHandler(AsyncHTTPTestCase):
+    internal_metadata_file = os.path.join(
+        os.path.dirname(__file__),
```
qiyunzhu: I saw this method being used multiple times. Is it worth creating a variable for it?

phyloviz/tests/test_handler.py
```diff
@@ -0,0 +1,46 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2018--, empress development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# ----------------------------------------------------------------------------
+import os
```
qiyunzhu: if `os.path.join` and `os.path.dirname` are the only cases when `os` is used, then this can be simplified as `from os.path import join, dirname`.

phyloviz/tests/test_model.py
```diff
@@ -1,263 +1,241 @@
...
-                'y11': ['y11', 408.3850804246091, 240.10442497874831],
-                'y14': ['y14', 375.00926942577706, 153.15746472040379],
-                'y15': ['y15', 331.86106285543758, 219.59987626950374],
-                'y16': ['y16', 257.89956082792412, 247.99103687506513],
-                'y17': ['y17', 178.78464125253325, 252.13727411816777],
-                'y18': ['y18', 128.92768189743305, 190.56905677707087],
-                'y2': ['y2', 79.070722542332845, 129.00083943597397],
-                'y6': ['y6', 74.068217341096869, 368.43664502236788],
-                'y7': ['y7', 117.21642391143635, 301.99423347326797]
+                0: [np.NaN, np.NaN, True, 'a1'],
```
qiyunzhu: I typically saw it in small letters: `np.nan`. My impression could be biased.

phyloviz/tests/test_model.py
```diff
@@ -1,263 +1,241 @@
...
+        # hack to make time column's dtype to be float64 instead of object
+        merge_exp[['time']] = merge_exp[['time']].apply(pd.to_numeric)
+
+        merge_res = self.tree.edge_metadata[[
+            'clade', 'time', 'gram-positive', 'species']].copy()
+        assert_frame_equal(merge_exp, merge_res)
+
+    def test_center(self):
+        center_exp = pd.DataFrame({
+            0: [(332.627 - 750), (450 - 450), (37.5 - 750), (154.873 - 450)],
```
qiyunzhu: Are `750` and `450` some hard-coded values (e.g., dimension of viewport)? Maybe worth to define them as constant at the beginning.

phyloviz/tests/test_tree.py
```diff
@@ -0,0 +1,335 @@
...
+from phyloviz.tree import Tree
+
+
+class TestDendrogram(unittest.TestCase):
+
+    def mock_random_tree(self):
+
+        np.random.seed(0)
+        x = np.random.rand(10)
+        dm = DistanceMatrix.from_iterable(x, lambda x, y: np.abs(x - y))
```
qiyunzhu: I see that the only moment you used `DistanceMatrix` is here. Actually, `condensed_form()` generates a numpy array. I guess it is easier to bypass `DistanceMatrix` and directly generate the numpy array. Trivial though.

phyloviz/tests/test_tree.py
```diff
@@ -0,0 +1,335 @@
...
+        return TreeNode.read(['(((a:1,e:2)f:1,b:2)g:1,(c:1,d:3)h:2)i:1;'])
+
+    def setUp(self):
+        self.tree2 = self.mock_tree_from_nwk()
+        self.tree1 = self.mock_random_tree()
+
+    def test_cache_ntips_random_tree(self):
+
+        t = Tree.from_tree(self.tree1)
+        t._cache_ntips()
```
qiyunzhu: This sounds like an awesome feature, though I don't know what it does accurately.

phyloviz/tree_with_webgl.html
```diff
@@ -3,13 +3,13 @@
...
   <meta charset="UTF-16">
   <title>WebGL - Experiment</title>
-  <link rel="stylesheet" type="text/css" href="{{ static_url('style.css') }}">
-  <script type="text/javascript" src="{{ static_url('gl-matrix.js') }}"></script>
-  <script type="text/javascript" src="{{ static_url('jquery_min.js') }}"></script>
-  <script type="text/javascript"  src="{{ static_url('init_webgl.js') }}"></script>
-  <script type="text/javascript" src="{{ static_url('callbacks.js') }} "></script>
-  <script type="text/javascript" src="{{ static_url('drawing_loop.js') }} "></script>
-  <script type="text/javascript" src="{{ static_url('init_webpage.js') }} "></script>
+  <link rel="stylesheet" type="text/css" href="/static/style.css?v=a8f02de1862809b8f6a88ccc6d8e36af">
```
qiyunzhu: What are these hard-coded hexes?

phyloviz/webserver.py
```diff
@@ -28,10 +30,32 @@ def __init__(self):
         tornado.web.Application.__init__(self, handlers, **settings)
 
 
-if __name__ == '__main__':
+@click.command()
+@click.option('--tree_file', '-f', required=True,
```
qiyunzhu: I think the convention is `--tree-file` instead of `--tree_file`. `click` automatically translates the former to `tree_file` as a variable name.

phyloviz/tree_with_webgl.html
mortonjt: This is a good question actually.  @kwcantrell , any idea where this came from?

phyloviz/tests/test_tree.py
mortonjt: mah - I don't think this is a deal breaker.

phyloviz/tests/test_model.py
mortonjt: Agree with @qiyunzhu here.

phyloviz/static/init_webpage.js
kwcantrell: Looks like this does not work. taken from https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement. Looks like the only additional optional parameter is an object containing a single property named 'is'

phyloviz/static/init_webpage.js
kwcantrell: The input to this method (edgeMeta) contains other fields as well. This method is part of a sequence of calls where the other methods need additional information. So, I just wanted to extract those fields from edge.

phyloviz/tests/test_controller.py
kwcantrell: The two lines above read the text from "tree_with_webgl.html" and store the text in lines. Then line 27-28 create a new file called "model_html_exp.txt" and also strip the beginning of each line in lines of white space.  Would it be better if I use a different variable so its easier to see that lines 24-25 are separate from lines 27-28?

phyloviz/tests/test_model.py
kwcantrell: They are the coordinates of the root tree node. I'll create a constants.

phyloviz/tests/test_tree.py
kwcantrell: This was a method we had copied over from gneiss. It is not used anywhere. Maybe we should remove it?

phyloviz/tree_with_webgl.html
kwcantrell: For test_controller I was comparing line by line to see if the html response was the same as this file and it expands all references to there hex values. So, I copied the hex values over so that the test would pass. I changed the way I tested this function in the later pull requests. I'll go ahead and make the same change in this pull request as well.

phyloviz/tests/test_tree.py
mortonjt: Looks like this functionality is now unnecessary with the newest refactor.  Im ok with removing

mortonjt:
Looks good!!

PR #48, opened 2018-07-31
Grid labels

kwcantrell:


empress/model.py
```diff
@@ -2,9 +2,7 @@
 from skbio import TreeNode
```
qiyunzhu: I think the order should by numpy -> pandas -> skbio -> empress. Trivial. Just a coding style suggestion. See [here](https://github.com/google/styleguide/blob/gh-pages/pyguide.md), find the "3.13 Imports formatting" section.

PR #52, opened 2018-08-08
File structure

kwcantrell:
Just changed the file structure to make it more organized/similar to the way emperor structures its files.


mortonjt:
I think this is good as is, since these are structural changes.

PR #79, opened 2018-10-24
need to speed up pandas access

kwcantrell:


empress/model.py
```diff
@@ -571,26 +592,65 @@ def collapse_selected_tree(self):
...
+        collapse_amount = self.tree.tip_count - TIP_LIMIT
+        cur_level = 0
+        pq = [(clade.level, clade.tip_count, self.__get_tie_breaker_num(), clade)
+                for clade in self.tree.levelorder(include_self=False)]
+
+        for clade in pq:
+            if collapse_amount == 0:
+                    break
+            if (collapse_amount - clade[TIP_COUNT] >= 0
+                    and self.edge_metadata.loc[self.edge_metadata['Node_id']==clade[NODE].name, 'branch_is_visible'].values[0]):
```
kwcantrell: This check takes a long time

qiyunzhu: How about:
```
self.edge_metadata.query('Node_id == %s' % clade[NODE].name)['branch_is_visible'].values[0]
```

If that isn't faster, I guess converting the dataframe into a dict will accelerate this:
```
d_ = dict(self.edge_metadata[['Node_id', 'branch_is_visible']].values)
```

empress/model.py
```diff
@@ -571,26 +592,65 @@ def collapse_selected_tree(self):
...
+                if clade[TIP_COUNT] > 1:
+                    self.__collapse_clade(clade[NODE])
```
mortonjt: Is this where all of the time is being gobbled up?  It looks like every call to this performs a post-traversal across the entire tree.  For some reason, I thought we could compute all of the collapsed coordinates once during initialization and just query.  Will that not work here?

kwcantrell: That is what I thought but I timed all of the events and each collapse takes at most 0.01 seconds. Plus only about 20ish collapses occur. What do you mean by collapsed coordinates?

mortonjt:
Ok, if it is that fast, let's not worry about it for now

empress/model.py
kwcantrell: @qiyunzhu  a dictionary should definitely help. Let me try that.

mortonjt:
Note that pandas internally should be a dictionary, so Qiyuns first
suggestion may just suffice

kwcantrell:
Fixed it! I tried Qiyun's first suggestion but I pandas returned an error. So I just decided to create a set containing all of the collapsed nodes since I retrieve those nodes inside __collapse_clade() already. Now it is much faster.

qiyunzhu:
The error in my first suggestion might be that `query` returns a single-row dataframe instead of a series. To fix, you can add something like `.iloc[0]`. But as long as your solution worked, we are all set!

PR #92, opened 2018-11-21
Speedy pandas

kwcantrell:
Updated to_df to use list of list


kwcantrell:
Create pandas dataframe from list of list instead of dictionary

PR #111, opened 2019-08-27
Travis test

kwcantrell:


fedarko:
Hi @kwcantrell and @ElDeveloper! I'll be helping out with Empress, and as I understand it this PR is where I'll try to start contributing.

So I'll be completely honest: I've never actually gotten Karma to run successfully with a browser environment before. (The bulk of my JS testing experience is with [Puppeteer](https://github.com/GoogleChrome/puppeteer), and even that is mostly indirect -- I use [mocha-headless-chrome](https://github.com/direct-adv-interfaces/mocha-headless-chrome/) a lot.) It looks like getting Karma to load a HTML file is possible, albeit kind of a pain? 

I spent some time today playing around with the code on this PR (Kalen, this looks fantastic, by the way), trying out some stuff and seeing what worked. One quick solution that might be good is the [qunit-puppeteer package](https://github.com/davidtaylorhq/qunit-puppeteer) (it's the second thing listed under "Integration" [on QUnit's site here](https://qunitjs.com/plugins/)).

This package made it super easy to test Empress headlessly:

```bash
$ npm install -g qunit-puppeteer
$ # Note: qunit-puppeteer doesn't accept relative filepaths, for some reason.
$ # As far as I can tell you have to give it an absolute path like follows:
$ qunit-puppeteer file:///Users/mfedarko/Dropbox/Work/KnightLab/Empress/empress/tests/index.html

Running: {}

...............................

Time: 284ms, Total: 440, Passed: 440, Failed: 0
```

The number of assertions (440) and dots (31) match to the number of assertions and dots listed on the actual Qunit test page (what you get when you open up tests/index.html in a browser). I can also confirm that breaking tests shows up in the console output:
```bash
$ # Before running this, I changed `i < 3` to `i < 2` on line 37 of the byte array tests.
$ qunit-puppeteer file:///Users/mfedarko/Dropbox/Work/KnightLab/Empress/empress/tests/index.html

Running: {}

..........................F....

Module Failed: Byte Array
  Test Failed: Test ByteArray.seqUnique()
    Assertion Failed: Test: seqUnique()
      Expected: [object Object], Actual: [object Object]

Time: 287ms, Total: 440, Passed: 439, Failed: 1
```

### Upside
This solution would be really simple to integrate into this PR! In the short term, I think this is a reasonable way to get headless testing up and running on Travis.

### Downsides
It really would be better to use a testing program/framework that'd allow us to look at JS code coverage (this is [possible](https://github.com/biocore/qurro/blob/master/Makefile#L20) with, e.g., `nyc` and `mocha-headless-chrome`, but implementing that in this repo would require some messing around to switch from QUnit to Mocha).

Also, I think eventually using a non-Puppeteer solution (allowing us to also test e.g. Firefox headlessly, not just Chrome) would be ideal.

### Anyway

Let me know what you think! If going with qunit-puppeteer seems good to y'all, I can remove Karma and commit the requisite changes to .travis.yml and the related files tomorrow. Otherwise, I can try out some other solutions and see what I can get working.

Thanks!

ElDeveloper:
@fedarko, thanks for working on this. We initially wanted to use Karma 
because it was suggested as a replacement for phantomjs + a qunit 
script. I think using qunit-puppeteer would be a great solution. If you 
can submit a new PR including the changes to use qunit-puppeteer that 
would be great. Maybe we can chat about the downsides of moving to 
mocha, if there are any :).

fedarko:
@ElDeveloper Ok! I opened a new PR at #114 I agree that Karma sounds like a good idea, so I definitely think it's a good idea to keep the stuff in this PR's branch around—I'm just not immediately familiar with getting Karma running.

My hot take is that most JS testing frameworks (mocha, qunit, jasmine, ...) are pretty similar, so whatever works and is easy to use is ok by me ;)

ElDeveloper:
Wonderful, thanks so much. Yes, I agree with you that's also my take on 
how the different frameworks compare.

ElDeveloper:
Closing this in favor of #114

PR #112, opened 2019-08-28
Bp oc cache

kwcantrell:
Added cache for open/close


ElDeveloper:
@kwcantrell, looks like your changes from #111 are being included here, would you mind removing those?

ElDeveloper:
@kwcantrell, if you could remove the changes from #111 from this PR that would be great, then we can merge that.

kwcantrell:
I went ahead and removed the files

PR #116, opened 2019-10-21
Add JS linting/style-checking; add contributing info; various JS style/code tweaks

fedarko:
As discussed over email between @kwcantrell, @ElDeveloper, and myself:

- Adds a lot of `make` targets for conveniently running style-checking, auto-formatting, etc.
- Run linting/stylechecking for JS code on Travis using jshint and prettier, respectively
- Add a `make` target (`make githook`) that sets up a pre-commit hook that calls `make stylecheck`. This isn't necessary for development, but it saves some headaches from Travis builds failing at the very end from small style problems.
- Add a `CONTRIBUTING.md` file describing how to use all of the aforementioned `make` targets.
- Various changes to the JS code to shut up jshint. **This still fails**: there's an error [here](https://github.com/fedarko/empress/blob/67d8fdd49efe997ea4ddb18ea3fdc5c5ad190c79/empress/support_files/js/legend.js#L16-L17) where a variable is clobbered where I'm not super sure how to rewrite things. So I'm leaving this as is so that someone who's more familiar with that code can address it.
- Chuck the JS code through prettier (using `make jsstyle`) to autoformat it. It should look somewhat more consistent, now.

(The final two bullet points are responsible for most of the changes in this PR.)

Thanks!

PR #117, opened 2019-10-23
Remove prime fix

kwcantrell:
This fix will allow for any arbitrary number of sample categories to be colored. I also changed the way names are handled. Before, I assumed all branch names would be unique however, I noticed that some tree do not have unique branch names. So, I created an additional structure that takes in a name and returns a list of all branches associated with it. Also, I now use post-order positions as an index into treeData because names are no longer assumed to be unique.


kwcantrell:
I am merging #116  into this PR

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+        var keys = [];
+        for (var i = 0; i < names.length; i++) {
+            var name = names[i];
+            keys.push(...this._nameToKeys[name]);
+        }
+        return keys;
+    };
+
+    /**
+     * Remove the keys non unique keys in keys
```
ElDeveloper: Would you mind rephrasing this sentence?

empress/support_files/js/empress.js
```diff
@@ -171,26 +169,23 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
     Empress.prototype.thickenSameSampleLines = function(amount) {
+        var tree = this._tree;
+
         // the coordinate of the tree.
         var coords = [];
         this._drawer.loadSampleThickBuf([]);
 
-        // iterate throught the tree in preorder, skip root
-        for (var i = 2; i <= this._tree.size; i++) {
+        // iterate throught the tree in postorder, skip root
```
ElDeveloper: Does the skip root comment still apply?

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+            return otherSets;
+        };
+
+        // for each item in keys, only keep the unique ones
+        var items = Object.keys(keys);
+        for (var i = 0; i < items.length; i++) {
+            var itemKeys = new Set(keys[items[i]]);
+            var keysInOtherSets = unionOtherSets(i, keys, items);
+            result[items[i]] = [...difference(itemKeys, keysInOtherSets)];
+        }
```
ElDeveloper: If we aren't already, perhaps it would be worthwhile adding a dependency on underscore.js or another utility library like that. I know several libraries already include a "unique" function.

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+
+        return result;
+    };
+
+    /**
+     * Creates a color map for each categoy in items
+     *
+     * @param{Array} items List of categories
```
```suggestion
     * @param {Array} items List of categories
```

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+     * @param {String} color The chroma color map to use
+     * @param {Boolean} rbg if true than a webGL color map will be created
```
```suggestion
     * @param {Boolean} rbg if true then a webGL color map will be created
```

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+     *                      if false than a javascript color map will be created
```
```suggestion
     *                      if false then a JavaScript color map will be created
```

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+     * @param {Object} sampleObs The object containing which tree branches are
+     *                colored by which sample category
+     * @param {Object} keyInfo The object containing the information to be
+     *                 displayed in the sample legend
+     */
+    Empress.prototype.percentColoredBySample = function(sampleObs, keyInfo) {
+        // calculate relative tree size i.e. the subtree spanned by the samples
+        // iterate over tree using postorder
+
+        var i,
```
```suggestion
```

empress/support_files/js/empress.js
```diff
@@ -264,101 +259,182 @@ define(["Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+            relativeTreeSize = 0;
```
```suggestion
            var i, relativeTreeSize = 0;
```

empress/empress-biom.py
```diff
@@ -21,21 +21,29 @@
 tools.name_internal_nodes(empress_tree)
 print('cacluate coords')
 empress_tree.coords(4020, 4020)
+
```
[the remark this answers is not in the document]
kwcantrell: I ported and removed script

empress/support_files/js/empress.js
kwcantrell: js linter seems to complain if I declare these variables on the same line

empress/support_files/js/empress.js
kwcantrell: I agree. Perhaps we can look into this on monday

empress/support_files/js/empress.js
kwcantrell: It still applies. All branch information comes from the child node so the root technically doesn't have a branch associated with it.

ElDeveloper: Got it, thanks!

empress/support_files/js/empress.js
kwcantrell: resolved

ElDeveloper: Great, thanks!

PR #120, opened 2019-10-24
DOC: Add IBM acknowledgement

ElDeveloper:


kwcantrell:
Thanks for added this! I will also merge this into #117

PR #121, opened 2019-10-26
Qiime plugin

kwcantrell:
Fixed the issue. Just needed to filter out samples that were not in the biom table.

https://github.com/kwcantrell/empress/blob/3d96816e9b0f4a9db80bf9d7ccc25fbefd33ed35/empress/_plot.py#L74-L76


kwcantrell:
Also, I think we should merge this PR before #117

ElDeveloper:
Great, thanks @kwcantrell !

Issue #126, opened 2019-11-21
nan is not defined


antgonza:
Hello,

I was trying to use Empress for one of my datasets and the page wasn't loading without any errors or messages; however, after a while I finally got these error in the Chrome console:
```
require-2.1.22.min.js:19 ReferenceError: nan is not defined
    at empress.html:129
    at Object.execCb (require-2.1.22.min.js:30)
    at ea.check (require-2.1.22.min.js:18)
    at ea.<anonymous> (require-2.1.22.min.js:23)
    at require-2.1.22.min.js:8
    at require-2.1.22.min.js:24
    at x (require-2.1.22.min.js:7)
    at ea.emit (require-2.1.22.min.js:24)
    at ea.check (require-2.1.22.min.js:20)
    at ea.<anonymous> (require-2.1.22.min.js:23)
```

Any thoughts on how to fix?

kwcantrell:
@antgonza can you post the list of commands you used to run empress? I have not seen that error before so it would be helpful to know what commands were used.

antgonza:
The command was: 
```bash
qiime empress plot \
   --i-tree insertion-tree.qza  \
   --i-feature-table feature-table.qza  \
   --m-sample-metadata-file metadata.txt  \
   --m-feature-metadata-file silva-taxonomy.qza  \
   --o-visualization empress.qzv
```

ElDeveloper:
Resolved as part of #127

PR #127, opened 2019-11-21
fixing the problem with nans in metadata

esayyari:
This issue is related to transform python types to JSON readable objects, see #126

### DIFF
--- a/empress/support_files/templates/empress-template.html
+++ b/empress/support_files/templates/empress-template.html
@@ -86,7 +86,7 @@
         var tree = new BPTree(tArr, {{ names }});
 
         var canvas = document.getElementById('tree-surface')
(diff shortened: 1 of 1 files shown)


antgonza:
Thanks, testing with my data to see if it works ...

antgonza:
BTW, while I tests, should this have a tests to make sure it never happens again?

kwcantrell:
Thanks @esayyari for catching that!

kwcantrell:
@antgonza that would be a good idea. I do not believe we currently have any tests related to metadata.

antgonza:
That worked, now I have a visualization. On a separated question, is it possible to color by the feature metadata?

kwcantrell:
Great to hear! Visualization of feature metadata is only supported on the legacy build of empress. We decided to rebuild empress from the ground up and are currently working on bringing back those features.

Issue #144, opened 2020-03-27
Set a maximum value for line width; add more granularity; normalize to tree dimensions


fedarko:
In practice, increasing the line width by 1 is a really big increase. In my opinion, it would be desirable to cap the line width at ~3, and then make the interface support smaller increases (e.g. by 0.1, or even keeping things as integers but scaling them by 1/10 or something), so that increasing the line width doesn't feel so sudden. (When you get up to line width = 3, it becomes really hard to see what's going on in the tree, especially when you zoom in.)

Below are screenshots of line widths = 1, 2, 3 for the two layouts. For each of the layouts' sections, I kept the camera at the same place for all 3 screenshots.

## unrooted

Line width = 1:

![image](https://user-images.githubusercontent.com/4177727/77720513-63a0f000-6fa5-11ea-8821-eda9f4709325.png)

Line width = 2:

![image](https://user-images.githubusercontent.com/4177727/77720538-73203900-6fa5-11ea-94eb-2747672854ef.png)

Line width = 3:

![image](https://user-images.githubusercontent.com/4177727/77720559-7c110a80-6fa5-11ea-9eda-96388d051c08.png)

## rectangular

Line width = 1:

![image](https://user-images.githubusercontent.com/4177727/77720708-f0e44480-6fa5-11ea-9f3f-2fee653daa47.png)

Line width = 2:

![image](https://user-images.githubusercontent.com/4177727/77720718-f9d51600-6fa5-11ea-8e71-7e8eb415d1fd.png)

Line width = 3:

![image](https://user-images.githubusercontent.com/4177727/77720736-02c5e780-6fa6-11ea-8dff-cf30ab24ec41.png)

ElDeveloper:
Great suggestion @fedarko, increments of 0.1 sound like a good idea.

kwcantrell:
I think this issue mainly stems from how the line width are calculated. Currently, the line width are created by creating a rectangle of width 2*line_width (in tree space). So, a line width of 2 will appear smaller or larger depending on how much space is spanned by the tree.

fedarko:
[From](https://github.com/biocore/empress/pull/145#discussion_r401141659) @kwcantrell, emphasis mine:

[...]

Issue #74, opened 2018-10-11
Identify and document scalability benchmarks


mortonjt:
Empress needs to be run against a huge tree (> 1 million tips)

antgonza:
Just wondering if there are any updates on this issue; thank you.

antgonza:
Installed the [latest version of empress](https://github.com/biocore/empress/commit/e308177aff1be716052f66b3b7c8e9f3b1127cdf) and ran it on one of the large trees generated in Qiita within a 2020.2 Qiime2 conda environment;  the mapping file, feature-table and taxonomies from the [moving pictures dataset](https://qiita.ucsd.edu/study/description/550#) - only one dataset. 

Note that this is a tree was created over a year ago (we could generate even  larger today), is the 100bp fragments insertion tree and is ~8.8M tips:
```python
In [1]: from skbio import TreeNode
In [2]: tree = TreeNode.read('../insertion_tree.relabelled.tre')
In [3]: print(tree.count(tips=True))     
8830174
```

I generated the no-taxonomy, GG and Silva added empress qzv's to test, each takes ~3hrs to generate the qzv and it works just fine (no error messages). However, when I try to open them in https://view.qiime2.org/, the browser fails with:
<img width="1252" alt="qiime2-view-error" src="https://user-images.githubusercontent.com/2014559/77239445-6e701500-6ba0-11ea-9ae1-9c41b910f077.png">
and if I unzip the qzv and try to open the index.html or empress.html, I get:
<img width="651" alt="error-opening-directly" src="https://user-images.githubusercontent.com/2014559/77239450-8051b800-6ba0-11ea-8874-e674cdb40dc0.png">

Anyway, here are the [testing files](https://drive.google.com/drive/folders/1ixE4j2WeuZeHGGInKmDwmyGWABw3x3w9).

cc: @ElDeveloper

kwcantrell:
@antgonza I'm looking into this

fedarko:
Once we identify upper bounds for what sorts of data sizes Empress can comfortably visualize, we should document this clearly in the README so that e.g. users with billion-tip trees know that they probably want to consult another tool and/or a priest ._.

PR #145, opened 2020-03-28
Support rectangular (and arbitrarily configurable) layouts; minor typo fixes / detail additions in CLI

fedarko:
Screenshot on the tree `(((a:1,e:2)f:1,b:2)g:1,(c:1,d:3)h:2)i:100;`, where I configured things so that all tips belong to the same sample category:

image

There are a lot of changes in this PR (some of them are a lot simpler than they seem -- e.g. a bunch of JS files got slightly changed because the version of `prettier` Travis was using changed, and I had to rerun it on the JS files to get the builds to pass), so it might be easiest for us to walk through the PR together next week.

Thanks!


kwcantrell:
Thank you Marcus!

empress/support_files/js/empress.js
```diff
@@ -187,60 +419,70 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+                var x2 = this.getX(this._treeData[node]);
+                var y2 = this.getY(this._treeData[node]);
+                var point = VectorOps.translate([x1, y1], -1 * x2, -1 * y2);
+
+                // find angle/length of branch
+                var angle = VectorOps.getAngle(point);
+                var length = VectorOps.magnitude(point);
+                var over = point[1] < 0;
+
+                //find top left of box of think line
```
ElDeveloper: Looks like autocomplete got us again should be thick not think.

empress/support_files/js/empress.js
```diff
@@ -187,60 +419,70 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+                tL = [0, amount];
+                tL = VectorOps.rotate(tL, angle, over);
+                tL = VectorOps.translate(tL, x2, y2);
+
+                tR = [length, amount];
+                tR = VectorOps.rotate(tR, angle, over);
+                tR = VectorOps.translate(tR, x2, y2);
+
+                // find bottom point of think line
```
```suggestion
                // find bottom point of thick line
```

empress/support_files/js/side-panel-handler.js
```diff
@@ -167,10 +170,79 @@ define(["Colorer"], function(Colorer) {
...
+
+            // Initialize the label pointing to that radio button
+            // https://stackoverflow.com/a/15750291
+            lele.htmlFor = iele.id;
+            lele.innerHTML = layouts[i];
+
+            // Use of onclick based on
+            // https://www.dyn-web.com/tutorials/forms/radio/onclick-onchange.php.
+            // Long story short, either onchange or onclick works for most
+            // browsers, but IE 9 is inconsistent with onchange -- hence our
```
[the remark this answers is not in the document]
fedarko: Understood! Accommodating it here was pretty simple, but in the future I won't go out of my way to do so.

empress/tree.py
```diff
@@ -51,6 +51,24 @@ def from_tree(cls, tree, use_lengths=True):
...
+                )
+            if not n.is_root():
+                max_branch_length = max(n.length, max_branch_length)
+
+        if max_branch_length == 0:
+            raise ValueError(
+                "Tree must contain at least one non-root branch with "
+                "length > 0."
+            )
+
```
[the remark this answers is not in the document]
fedarko: The input that `empress.Tree.from_tree()` gets is a skbio `TreeNode`, which is mostly why I used `TreeNode` to do the traversals here. There's already a `TreeNode` traversal call being made in this function below these traversals, so these traversals seemed to match with "precedent."

I'd be down to switch these traversals to use `BP`, and/or to modify the empress `Tree` class to use `BP` rather than `TreeNode` internally in general. However, I'm not very familiar with the `BP` API (and switching things over would probably take a nontrivial amount of time), so my preference would be for making this a new issue rather than doing this change as part of this PR.

tests/python/test_tree.py
```diff
@@ -20,31 +20,169 @@ def setUp(self):
...
+            A tree whose nodes contain xattr and yattr attributes.
+
+        xattr : str
+            An attribute of nodes indicating the x-coordinates to check.
+            e.g. "x2" for the unrooted layout, "xr" for the rectangular layout
+
+        yattr : str
+            An attribute of nodes indicating the y-coordinates to check.
+            e.g. "y2" for the unrooted layout, "yr" for the rectangular layout
+
```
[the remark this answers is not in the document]
fedarko: I might be missing something here, but I don't see how these suggestions differ from the current section of documentation? Like, I copied both the red and the green blocks into separate files, and they had the same md5 hash :sweat_smile:

empress/support_files/js/empress.js
```diff
@@ -13,17 +13,30 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
      * @param {BIOMTable} biom The BIOM table used to color the tree
      * @param {Canvas} canvas The HTML canvas that the tree will be drawn on.
      */
-    function Empress(tree, treeData, nameToKeys, biom, canvas) {
+    function Empress(
+        tree,
+        treeData,
+        nameToKeys,
+        layoutToCoordSuffix,
+        default_layout,
```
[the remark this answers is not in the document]
fedarko: Done, thanks!

empress/tree.py
```diff
@@ -105,21 +131,195 @@ def coords(self, height, width):
...
+            height of the canvas
+
+        References
+        ----------
+        https://github.com/qiime/Topiary-Explorer/blob/master/src/topiaryexplorer/TreeVis.java
+            Description above + the implementation of this algorithm
+            derived from the Polar/Radial layout algorithm code.
+        """
+        pass
+
```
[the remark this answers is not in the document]
fedarko: Done! Looking forward to hopefully moving it back in soon ;)

fedarko:
OK, I think the requested changes so far should be mostly addressed. Please let me know if there's anything else you'd like me to change! Thanks guys!

empress/_plot.py
```diff
@@ -48,19 +48,29 @@ def plot(output_dir: str,
...
+        # Add coordinate data from all layouts for this node
+        for layoutsuffix in layout_to_coordsuffix.values():
+            xcoord = "x" + layoutsuffix
+            ycoord = "y" + layoutsuffix
+            tree_data[i][xcoord] = getattr(node, xcoord)
+            tree_data[i][ycoord] = getattr(node, ycoord)
+        # Also add vertical bar coordinate info for the rectangular layout
+        if not node.is_tip():
+            tree_data[i]["highestchildyr"] = node.highestchildyr
+            tree_data[i]["lowestchildyr"] = node.lowestchildyr
```
kwcantrell: By change do you know how much memory is being used to store all the layout? If its a lot, do you think it would be better to store the different layouts in a .txt file and then have JavaScript read from that file when switching layouts. Do worry about modifying the PR, this can be differed to a different issue.

fedarko: Compared to the current "main" version of Empress, the extra things being stored (on a per-node basis) are:

- rectangular layout coordinates for each node (`xr` and `yr` attributes)
- highest- and lowest- child y coordinates for each internal node (`highestchildyr` and `lowestchildyr` attributes)

So, it's a max of ~4 extra numbers per node (and that's an overestimate, since leaves will only require 2 extra numbers). That's pretty much it -- the only other "extra stuff" being stored in memory for the JS is stuff like `layout_to_coordsuffix` and `default_layout`, which should all be negligible in size compared to a big tree.

In general, adding additional layouts using this PR's functionality will require at minimum two numbers (a position) for each node in the tree, so I think (keyword = think :) that this should scale reasonably well. But I'd totally be down to explore alternate solutions in the future -- with trees with millions of nodes, I suppose "4 extra numbers per node" is a pretty nontrivial amount of memory!

empress/support_files/js/empress.js
```diff
@@ -163,19 +286,128 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
      * (i.e. features that are only in gut)
      *
      * @param {Number} amount - How thick to make branch
      */
-    Empress.prototype.thickenSameSampleLines = function(amount) {
+    Empress.prototype.thickenSameSampleLines = function (amount) {
+        // we do this because SidePanel._updateSample() calls this function
+        // with lWidth - 1, so in order to make sure we're setting this
+        // properly we add 1 to this value.
+        this._currentLineWidth = amount + 1;
```
kwcantrell: I'm adding this for #144 

It looks like the method has changed a bit so @fedarko correct me if this is wrong. Originally, I defined a line width of 1 to be equivalent to "0" i.e. the branches are just colored. This is why SidePanel._updateSample() would call lwidth - 1 since a "line_width" of 1 is really a line width of 0. 

Also, lines 436-9 is what I was referring to in my comment in #144 when I said line width are "created in tree space". 

To create the "thickened" lines, the goal is to create a rectangle (made up of two triangles for webgl) of width 2*lwidth  centered on the original branch. But if you notice, for example, the y component of the triangle is defined as this.getY(this._treeData[node]) + amount. Here, we are adding "amount" to the y-coord of the tree. This will result in rectangle (centered on the branch) of width 2*amount. However, because we are adding amount directly to the y-coord itself with out taking into account the camera/tree size, a line width of 2 will appear large/smaller depending on how large the tree is. For example, say we have tree t1 whose y-coords are in the range [-1000, 1000] and t2 whose y-coords are in the range [-100, 100] a line width of 2 in will appear 10x larger in t2 than in t1. To fix this, we need to first normalize "amount" so that a line width of 2 will proportionally take up the same amount of space in t1 as it would in t2.

tests/python/test_tree.py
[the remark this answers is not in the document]
kwcantrell: I t looks like lines 87 and 91 (white space lines) were deleted.

empress/support_files/js/empress.js
```diff
@@ -83,41 +96,111 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+        var numLines;
+        if (this._currentLayout === "Rectangular") {
+            // Leaves have 1 line (vertical), the root also has 1 line
+            // (horizontal), and internal nodes have 2 lines (both vertical and
+            // horizontal).
+            var leafCt = tree.numleafs();
+            numLines = leafCt + 1 + 2 * (tree.size - leafCt - 1);
+        } else {
+            numLines = tree.size - 1;
+        }
```
kwcantrell: It may be worthwhile to port this to a separate function to modularize some of the functionality of this method.

empress/support_files/js/empress.js
```diff
@@ -187,60 +419,70 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+                var y2 = this.getY(this._treeData[node]);
+                var point = VectorOps.translate([x1, y1], -1 * x2, -1 * y2);
+
+                // find angle/length of branch
+                var angle = VectorOps.getAngle(point);
+                var length = VectorOps.magnitude(point);
+                var over = point[1] < 0;
+
+                // find top left of box of thick line
+                tL = [0, amount];
```
kwcantrell: Again this comment is just to for #144.

Refer to https://github.com/biocore/empress/pull/145/files#r401141659. Here we again want to make sure "amount" is normalized.

empress/support_files/templates/side-panel.html
```diff
@@ -38,4 +38,12 @@
...
\ No newline at end of file
+</div>
+
+<!-- Layout options; shamelessly copied from above.
+     Stuff will be added to the #layout-div element below when
+     SidePanel.addLayoutTab() is called :)
+-->
+<button class="side-header collapsible">Layout</button>
+<div class="side-content control hidden" id="layout-div">
+</div>
```
kwcantrell: I think we should split side-panel.html into two separate files (i.e. tabs). One for sample coloring and one for "global" properties such layouts and default background/branch color. Same for SidePanel.js. Split SidePanel.js into SampleColorPanel.js and GlobalPropertiesPanel.js.

This isn't a pressing issue and can be moved to a separate issue but I think it will help contain separate functionality.

empress/tree.py
```diff
@@ -96,7 +119,15 @@ def update_geometry(self, use_lengths, depth=None):
...
-        """ Returns coordinates of nodes to be rendered in plot.
+        """ Computes the coordinates of nodes to be rendered in plot.
+
+        This runs multiple layout algorithms and saves all of the resulting
+        coordinates for each node, so that layout algorithms can be rapidly
+        toggled between in the JS interface.
+
+        Also adds on .highestchildyr and .lowestchildyr attributes to internal
+        nodes so that vertical bars for these nodes can be drawn in the
+        rectangular layout.
```
kwcantrell: Just curious, how long does this take? If it takes a while, do you think we should either let the user choose which which layouts they want instead of defaulting to calculate all. Alternatively, do you know if it would be possible to add some kind of progress bar?

kwcantrell:
This looks good! I added a few comments but nothing major.

ElDeveloper:
@kwcantrell, if all looks good to you feel free to merge when your comments are addressed.

tests/python/test_tree.py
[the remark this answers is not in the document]
fedarko: Ah whoops my bad, I completely missed that!!! Applying the suggestion now...

empress/support_files/js/empress.js
fedarko: Agreed on all counts! Having `thickenSameSampleLines()` take care of this stuff would be ideal. It might be a good idea to have each layout's tree dimensions passed from the Python to the JS Empress object, so that `thickenSameSampleLines()` can (based on whatever the current layout's dimensions are) adjust thicknesses accordingly.

I'll update #144 with these details. Thanks so much for the detailed explanation!

empress/support_files/js/empress.js
fedarko: Done, thanks! New function is called `computeNecessaryCoordsSize()`.

empress/support_files/templates/side-panel.html
fedarko: Agreed that this would make the side panel more maintainable, esp. as we start to add in more and more features. Opened #150.

empress/tree.py
fedarko: I can't say for sure how the unrooted layout scales, but I do know that the rectangular layout process should take `O(3n)` time for a tree with `n` nodes: it uses 3 full-tree traversals. (The first one of these traversals also checks the child y-positions for internal nodes to center them, but I don't think this should increase the runtime that much.)

I don't *think* precomputing layouts will be the primary limiting factor for huge trees, but I guess it's possible. My preference would be for allowing users to opt out of calculating all layouts, since I imagine for most users the up front time taken won't be a problem (and having to rerun empress a few times to generate a tree with the layouts they want might be cumbersome).

Adding a progress bar is totally possible -- an example of this is [Songbird](https://github.com/biocore/songbird/), which uses tqdm to display a progress bar. That being said I would caution against adding this in, since in practice the progress bar can be a bit annoying for "fast" jobs.

Let me know what you think -- IMO these are both ideas we should think about more, so it'd make sense to convert these into issues we can mull over. Tagging @ElDeveloper in case he has any thoughts on this.

fedarko: Oh, also linking #74 for reference.

fedarko:
I think I've addressed all of the comments by this point, but please either of you let me know if you'd like me to make any more changes :)

ElDeveloper:
@kwcantrell if all looks good to you feel free to merge this.

kwcantrell:
Great work everyone!

ElDeveloper:
Thanks both!

PR #146, opened 2020-03-28
Animation

kwcantrell:


empress/support_files/js/empress.js
```diff
@@ -284,7 +293,13 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
      */
     Empress.prototype._keepUniqueKeys = function(keys) {
         // get unique keys
-        var uniqueKeys = _.chain(keys)
+        var items = Object.keys(keys);
+        var i;
+
+        // TODO: This method current method to get the unique observations
+        // belonging to each sample category is slow. Refactoring it will lead
+        // to a nice speed boost.
```
```suggestion
        // TODO: The current method to get the unique observations
        // belonging to each sample category is slow. Refactoring it will lead
        // to a nice speed boost.
        // https://github.com/biocore/empress/issues/147
```

empress/support_files/js/legend.js
```diff
@@ -97,18 +97,9 @@ define([], function() {
...
             div.appendChild(component);
 
-            // total percentage of leafs
-            component = document.createElement("label");
-            component.classList.add("gradient-label");
-            component.innerHTML = this.__formatNumLabel(info[key].tPercent);
-            div.appendChild(component);
-            container.appendChild(div);
+            // TODO: add total percentage of tree colored by each category
+            //       old method was not correct.
```
ElDeveloper:
```suggestion
            // TODO: add total percentage of tree colored by each category
            // https://github.com/biocore/empress/issues/148
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        // show the following buttons
+        this.stopBtn.classList.remove("hidden");
+        this.pauseBtn.classList.remove("hidden");
+    };
+
+    /**
+     * Makes the prev/next/stop/resume buttons visible. This is the menu shown
+     * when user pauses the animation.
+     *
+     * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+
+        // show the following buttons
+        this.startBtn.classList.remove("hidden");
+    };
+
+    /**
+     * Makes the stop/pause buttons visible. This is the menu shown during the
+     * animation.
+     *
+     * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        this.resumeBtn = document.getElementById("animate-resume-btn");
+        this.prevFrameBtn = document.getElementById("animate-prev-btn");
+        this.nextFrameBtn = document.getElementById("animate-next-btn");
+    }
+
+    /**
+     * Makes the play button visible. This is the menu shown before user has
+     * started the animation.
+     *
+     * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+
+        // dont show next button on last frame
+        if (this.animator.onLastFrame()) {
+            this.nextFrameBtn.classList.add("hidden");
+        }
+    };
+
+    /**
+     * enables/disable the drop down menus. When the animation is playing,
+     * the drop down menus are disable in order to prevent user from changing
```
ElDeveloper:
```suggestion
     * Enables/disable the drop down menus. When the animation is playing,
     * the drop down menus are disabled in order to prevent user from changing
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+     * the gradient/trajectory during an animation.
+     *
+     * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+     */
+    AnimatePanel.prototype.__toogleSelects = function(disableStatus) {
+        this.colorSelect.disabled = disableStatus;
+        this.gradient.disabled = disableStatus;
+        this.trajectory.disabled = disableStatus;
+    };
+
+    /**
+     * Initializes GUI components/set up callback events
+     * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+
+        // copy options and add them to trajectory drop down menu
+        var options = this.gradient.innerHTML;
+        this.trajectory.innerHTML = options;
+
+        /**
+         * Event: triggers when user clicks on the hide branch checkbox.
+         * Sets hide parameter in animation state machine.
+         *
+         * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+            ap.animator.setAnimationParameters(
+                trajectory,
+                gradient,
+                cm,
+                hide,
+                lWidth
+            );
+
+            // start animation
+            ap.animator.startAnimation();
```
ElDeveloper: 🎥

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        };
+
+        /**
+         * Event: triggers when user clicks on pause button.
+         * Pauses the animation.
+         *
+         * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        this.pauseBtn.onclick = function() {
+            ap.__resumeOptions();
+            ap.animator.pauseAnimation();
+        };
+
+        /**
+         * Event: triggers when user clicks on resume button.
+         * Resumes the animation.
+         *
+         * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        this.resumeBtn.onclick = function() {
+            ap.__pauseOptions();
+            ap.animator.resumeAnimation();
+        };
+
+        /**
+         * Event: triggers when user clicks on stop button.
+         * Stops the animation and clears the state machine
+         *
+         * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+            ap.__toogleSelects(false);
+            ap.__startOptions();
+            ap.animator.stopAnimation();
+        };
+
+        /**
+         * Event: triggers when user clicks on previous button.
+         * Shows the previous frame in the animation
+         *
+         * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        this.prevFrameBtn.onclick = function() {
+            ap.animator.prevFrame();
+            ap.__resumeOptions();
+        };
+
+        /**
+         * Event: triggers when user clicks on next button.
+         * Shows the next frame in the animation.
+         *
+         * @return {null}
```
ElDeveloper:
```suggestion
```

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+        /**
+         * @type {Number}
+         * Stores the current timeframe that WILL BE DISPLAYED NEXT not the one
+         * that is currently shown on screen.
+         */
+        this.curFrame = null;
+
+        /**
+         * @type {number}
+         * The total duration of the animation
```
ElDeveloper:
```suggestion
         * The total duration of the animation in milliseconds.
```

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+         */
+        this.lWidth = null;
+    }
+
+    /**
+     * Sets the parameters for the animation state machine.
+     *
+     * @param {String} trajectory Metadata column to use for trajectory
+     * @param {String} gradient Metadata column to user for gradient
+     * @param {}String} cm The color map to use for the animation
```
ElDeveloper:
```suggestion
     * @param {String} cm The color map to use for the animation
```

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+        gradient,
+        cm,
+        hide,
+        lWidth
+    ) {
+        // used in closure
+        var animator = this;
+
+        this.gradientCol = gradient;
+        this.gradientSteps = this.empress.getUniqueSampleValues(gradient);
```
ElDeveloper: It would probably be nice if we could move this functionality out of the Empress class and into a method for the metadata/biom table.

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+        // draw new legend
+        this.legend.clearAllLegends();
+        this.legend.addColorKey(name, keyInfo, "node", false);
+
+        //draw tree
+        this.empress.resetTree();
+        this.empress._colorTree(obs, this.cm);
+        this.empress.thickenSameSampleLines(this.lWidth);
+
+        // // TODO: hide should be taken care of in empress state machine
```
ElDeveloper:
```suggestion
        // TODO: hide should be taken care of in empress state machine
```

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+        this.empress.setNonSampleBranchVisibility(this.hide);
+        this.empress.drawTree();
+        this.curFrame += 1;
+    };
+
+    /**
+     * The asynchronous animation loop. This method will continous create
```
ElDeveloper:
```suggestion
     * The asynchronous animation loop. This method will continously create
```

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+     * timeout events until animation is done or user pauses animation. If a
+     * frame is reached before it is ready, then a new timeout event will be
+     * created.
+     */
+    Animator.prototype.playAnimation = async function() {
+        // used in closure
+        var animator = this;
+
+        // Animation loop
+        var animationID = setTimeout(function loop() {
```
ElDeveloper: Looks like the animationID is not used for anything, right?

```suggestion
        setTimeout(function loop() {
```

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+     */
+    Animator.prototype.resumeAnimation = function() {
+        this.pause = false;
+        this.playAnimation();
+    };
+
+    /**
+     * Stops the animation and clears state machine parameters
+     */
+    Animator.prototype.stopAnimation = function() {
```
ElDeveloper: As I understand "stop" is more of a cancel action, right? Do you think it would be worthwhile renaming?

empress/support_files/js/legend.js
fedarko: It's still not really clear how the previous method was incorrect from looking at #148. Not urgent or anything, but describing the bug somewhere would be nice (if nothing else, so that someone unfamiliar with it like me doesn't accidentally make the same mistake again :P)

kwcantrell: @fedarko thanks for the suggestion, I'll add a description. But the tldr was that it calculated the summaries before the non-unique features were filtered out.

empress/support_files/js/biom-table.js
```diff
@@ -127,5 +138,83 @@ define([], function() {
...
+
+            // ignore missing data
+            if (cVal === "unknown" || (isNumeric && isNaN(cVal))) {
+                continue;
+            }
+            values.add(cVal);
+        }
+
+        // convert result to array and sort
+        values = [...values];
```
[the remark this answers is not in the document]
kwcantrell: That's a good point. I believe the spread operator will expand the entire values set inside [ ] before creating the array which means we will temporarily create two copies of values before it is converted into an array. Do you know if Array.from() is better? 

Also, would you know how many sample are typically in a study? values will never have more elements than the number of samples since values creates a set from the sample values.

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
+define(["Colorer"], function(Colorer) {
+    /**
+     *
+     * @class AnimationPanel
```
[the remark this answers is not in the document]
kwcantrell: Thanks, for catching this. AnimatePanel should be AnimationPanel. I'll update the PR to make this change.

empress/support_files/js/legend.js
fedarko: @kwcantrell sweet, thanks for letting me know!

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+     *
+     * @return {null}
+     * @private
+     */
+    AnimatePanel.prototype.__pauseOptions = function() {
+        // hide the following buttons
+        this.startBtn.classList.add("hidden");
+        this.resumeBtn.classList.add("hidden");
+        this.prevFrameBtn.classList.add("hidden");
+        this.nextFrameBtn.classList.add("hidden");
```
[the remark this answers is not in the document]
kwcantrell: @fedarko thanks for the suggestion! My goal with the GUI is to make things as intuitive and easy to use. One way I try to accomplish this is by only presenting options to the user that can be applied at that moment. Since the prev/next features can only be used when the animation is paused, I decided to hide those options unless the user pauses the animation. 

Instead of hiding those buttons, I can disable while the animation is running and enable them when the animation is paused. Also, I can start the animation is the pause state instead of the run state so users who wanted to manually step through the animation could do so without having to first pause the animation and then returning to the first frame.
 
@ElDeveloper how does Emperor handle this?

empress/support_files/js/animation-panel-handler.js
```diff
@@ -0,0 +1,257 @@
...
+        this.nextFrameBtn.classList.remove("hidden");
+
+        // dont show previous button on frame 1
+        if (this.animator.onFirstFrame()) {
+            this.prevFrameBtn.classList.add("hidden");
+        }
+
+        // dont show next button on last frame
+        if (this.animator.onLastFrame()) {
+            this.nextFrameBtn.classList.add("hidden");
```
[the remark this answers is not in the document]
kwcantrell: Nice catch! I'll update the PR to disable the prev/next button when on the first/last frame instead of hiding them.

empress/support_files/js/animator.js
kwcantrell: All the work for this function is done in the BIOMTable class. Empress is simply calling the function. Since BIOMTable is responsible for holding all of the metadata, I would like to limit the number of classes that have direct access to it. What do you think?

Also, if we move the Animator object into Empress like we discussed, we can expose the Animator UI through the Empress and essentially "encapsulate" it in Empress.

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+        var obs = frame.obs;
+
+        // draw new legend
+        this.legend.clearAllLegends();
+        this.legend.addColorKey(name, keyInfo, "node", false);
+
+        //draw tree
+        this.empress.resetTree();
+        this.empress._colorTree(obs, this.cm);
+        this.empress.thickenSameSampleLines(this.lWidth);
```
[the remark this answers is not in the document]
kwcantrell: Good catch! SidePanel (which probably should be renamed to SampleColorer or something like) should not be responsible for adjusting the line width as that operation should be handled by the thickenSameSampleLines (which also should be renamed since it is thickens all colored branches not same sample lines). 

I'll update #144 to include this.

empress/support_files/js/animator.js
kwcantrell: Correct, this was an artifact left over from a previous method I was exploring.

empress/support_files/js/animator.js
kwcantrell: Correct, "stop" is a cancel action. Would something like "clear" be better?

empress/support_files/js/animator.js
```diff
@@ -0,0 +1,397 @@
...
+     * Show the next timeframe. This method is only called when animation is
+     * paused and user presses the next button.
+     */
+    Animator.prototype.nextFrame = function() {
+        // curFrame is the next frame to draw
+        // make sure curFrame is not passed last frame
+        if (this.curFrame >= this.totalFrames) {
+            this.curFrame = this.totalFrames - 1;
+        }
+        this.drawFrame();
```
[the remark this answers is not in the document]
kwcantrell: I like the idea of having displaying some info of the "progress". For the time being, I updated the legend to show this information since it only required an extra line of code.

@ElDeveloper do you have any input on this topic?

empress/support_files/js/drawer.js
```diff
@@ -42,9 +42,12 @@ define(["jquery", "glMatrix", "Camera"], function($, gl, Camera) {
      */
     function Drawer(canvas, cam) {
         this.contex_ = canvas.getContext("webgl");
-        this.CLR_COL = 1;
         this.cam = cam;
         this.VERTEX_SIZE = 5;
+
+        // sets empress to night mode
+        // TODO: make this a customizable parameter
+        this.CLR_COL = 0;
```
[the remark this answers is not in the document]
kwcantrell: I switched it to match Emperors default background. However, I don't have a preference for the default color. @ElDeveloper any thoughts?

empress/support_files/js/biom-table.js
```diff
@@ -127,5 +138,83 @@ define([], function() {
...
+        var values = new Set();
+        var isNumeric = this._types[category] === "n";
+        for (var sample in this._samp) {
+            // grab next value in column
+            var cVal = this._samp[sample][category];
+
+            // ignore missing data
+            if (cVal === "unknown" || (isNumeric && isNaN(cVal))) {
+                continue;
+            }
```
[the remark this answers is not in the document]
kwcantrell: That is a good idea. For now, I update the PR so that it does not ignore missing data (missing data does not cause an error, it simply be handles it as another trajectory value). 

Would it be possible to notify the user of missing data through qiime2 when Empress is being initialized and create a .txt file with the missing columns/rows? We could also create toast messages when missing data is encountered for example during the animation, when a time-frame is shown with missing data.

empress/support_files/js/empress.js
```diff
@@ -265,12 +265,21 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function(
...
+        var keys = new Set();
+
+        // adds a key to the keys set.
+        var addKey = function(key) {
+            keys.add(key);
+        };
+
         for (var i = 0; i < names.length; i++) {
-            keys.push(...this._nameToKeys[names[i]]);
+            // most names have a one-to-one correspondence but during testing
```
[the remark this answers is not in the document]
kwcantrell: _nameToKeys map the names of nodes (i.e. the names that appear in the newick string) to the nodes in the BPTree. I'll update the comment to make this more clear.

empress/support_files/templates/animate-panel.html
```diff
@@ -0,0 +1,48 @@
...
+    <label for="animate-gradient">Gradient</label>
+    <label class="select-container">
+      <select id="animate-gradient"></select>
+    </label>
+  </p>
+  <p>
+    <label for="animate-trajectory">Trajectory</label>
+    <label class="select-container">
+      <select id="animate-trajectory"></select>
+    </label>
```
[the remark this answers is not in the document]
kwcantrell: Great suggestion!

empress/support_files/templates/animate-panel.html
```diff
@@ -0,0 +1,48 @@
...
+      <button id="animate-stop-btn">Stop Animation</button>
+    </p>
+    <p>
+      <button id="animate-pause-btn">Pause</button>
+    </p>
+    <div>
+      <p>
+        <button id="animate-resume-btn">Resume</button>
+        <button id="animate-prev-btn">Previous Frame</button>
+        <button id="animate-next-btn">Next Frame</button>
```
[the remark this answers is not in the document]
kwcantrell: I think visual icons would definitely give a nice touch. I looked into using UNICODE characters for icons but I could get all of the icons to show up correctly. I was going to look for images but I am not sure if that would violate any license/copy-write. I think we should defer this to a separate issue.

@ElDeveloper any thoughts?

empress/support_files/templates/side-panel.html
```diff
@@ -31,7 +31,7 @@
       </p>
       <p>
         <label for="sample-line-width">Line Width</label>
-        <input id="sample-line-width" type="number" value="1">
+        <input id="sample-line-width" type="number" value="4">
```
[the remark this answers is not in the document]
kwcantrell: This ties into the comment I posted on #144. The main issue comes from how line widths are calculate in empress. To create new lines, a rectangle is created (in tree space) whose width is 2*line_width. So, a line width of 4 can appear larger/smaller depending on how much space the tree spans. 

I set the default to 4 before #144 came up because the tree I, when testing the animation feature, a line width of 4 looked the best. I guess I forgot to switch it back.

empress/support_files/templates/animate-panel.html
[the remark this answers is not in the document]
fedarko: I'm fine with deferring this to an other issue, especially since icons could be used in other parts of the interface (e.g. the layout/coloring sections). In practice there are like infinity ways to do this, so it's really down to whatever you're most comfortable with -- we recently had to add icons for Qurro, and [Bootstrap Icons](https://icons.getbootstrap.com/#css) worked pretty well.

empress/support_files/js/animator.js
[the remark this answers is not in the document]
fedarko: this looks really cool!!!

empress/support_files/templates/side-panel.html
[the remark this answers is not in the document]
fedarko: Understood! That makes sense to me. We'll probably have to reevaluate stuff for #144, but until then I guess the defaults don't matter too much.

empress/support_files/js/animation-panel-handler.js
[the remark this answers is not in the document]
ElDeveloper: In emperor we disable the buttons and still leave them visible (instead of hiding them).

empress/support_files/js/animator.js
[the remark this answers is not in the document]
ElDeveloper: This looks perfect!

empress/support_files/js/biom-table.js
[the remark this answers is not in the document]
ElDeveloper: The number of samples can be up to hundreds of thousands. Not sure what method might be better these days, but I think Array.from might be better.

empress/support_files/js/drawer.js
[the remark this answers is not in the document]
ElDeveloper: Let's keep the light background until this can be user-modifiable.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
ElDeveloper: Thanks!

empress/support_files/templates/animate-panel.html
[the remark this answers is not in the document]
ElDeveloper: Let's defer to a separate PR. I've opened an issue: https://github.com/biocore/empress/issues/149

empress/support_files/templates/animate-panel.html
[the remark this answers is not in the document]
ElDeveloper: Great suggestion @fedarko, I suggest doing that now using the `title=` attribute.

empress/support_files/js/drawer.js
```diff
@@ -46,7 +46,7 @@ define(["jquery", "glMatrix", "Camera"], function($, gl, Camera) {
         this.VERTEX_SIZE = 5;
 
         // sets empress to night mode
```
fedarko, line 48:
```suggestion
        // sets empress to light mode
```

ElDeveloper:
Thanks everyone!

🎆 :shipit: 🚀 🎥

Issue #130, opened 2019-11-22
color by feature metadata


antgonza:
Currently is not possible to highlight via the GUI via feature metadata and this will be great.

Issue #142, opened 2020-03-26
Add JS tests covering more modules


fedarko:
Listing JS files currently in `empress/support_files/js/` and whether or not a `test-[filename].js` file already exists for the file:

- [ ] `biom-table.js`
   - (worth noting that `test-summary-helper.js` uses the JS BIOM Table implementation a lot)
- [x] `bp-tree.js`
- [x] `byte-array.js`
- [x] `camera.js`
- [ ] `colorer.js`
- [ ] `drawer.js`
- [ ] `empress.js`
- [ ] `legend.js`
- [ ] `side-panel-handler.js`
- [x] `summary-helper.js`
- [ ] `vector-ops.js`

I imagine we'd want to cover all or at least most of these files eventually. Adding tests for more complex modules that rely on page state (e.g. `empress.js`) may be tricky, but should be doable. (And on the other side of things, some files, e.g. `vector-ops.js`, hopefully shouldn't require much effort to test -- since they just expose utility functions that can be unit-tested in isolation.)

One more note, somewhat related: adding JS code coverage via CodeCov would also help make identifying untested portions of the code easier (I rely on digging through coverage a lot, esp when testing JS code). That being said, this might require us to use a different test runner than `qunit-puppeteer`, since I don't think that it currently outputs JS test coverage information.

PR #154, opened 2020-04-08
"Match" table, tree, and sample metadata, and verify that things seem ok

### LINKED ISSUE #139: Empress' JS gets stuck in the loading animation when none of the samples in the table have any metadata
empressscreenshot

In this test data, there are two samples: `S1` and `S2`, and every feature in the tree is present in one of these samples. When neither `S1` nor `S2` are in the metadata, this error occurs; when at least one of the two samples have metadata, the Empress visualization loads (i.e. the loading animation stops and the visualization becomes interactive).

this (and related errors re: "matching" between datasets) should be solvable by adding in validation code that explicitly checks that:

- [ ] features in the tree are all present in the BIOM table
- [ ] all samples in the table are also present in the metadata
  - (we could relax this and just filter to samples that are shared by the table and the metadata, so long as at least 1 sample is shared in this way; this is what Qurro does)
- [ ] In the future, that all features in the tree have feature metadata (however, Empress currently doesn't use feature metadata so this check won't be super helpful now)

Ideally, "problematic" or impossible cases would cause an error when the user tries to create a visualization from the command line, rather than later on in the JS side of things.

I know that Qurro has code to do similar validation [here](https://github.com/biocore/qurro/blob/b9613534b2125c2e7ee22e79fdff311812f4fefe/qurro/_df_utils.py#L255), and there are plenty of other packages from the lab that do similar things (e.g. Songbird [here](https://github.com/biocore/songbird/blob/8e107782508cfa1532c59aa5a60bae1317f1783e/songbird/util.py#L132)). The main time-consuming task here will just be adding lots of tests to make sure that whatever validation code we add behaves as intended.

fedarko:
Closes #139. Now, the sort of case described there will trigger the following error:

image

We will still need to match/validate feature metadata (e.g. taxonomy) when we add support for that in to Empress (#130) -- I haven't done that in this PR, but I can add it in if you'd like. (Wasn't sure how we'd prefer to handle feature metadata, so I haven't included this here.)

Also changed in this PR: I moved the QZAs/etc. needed for the moving pictures example back into the repository (in `docs/moving-pictures/`), and added a `make docs` command that'll re-generate the `empress-tree.qzv` visualization linked from the README. We could make running `make docs` part of the Travis build in the future if desired, as a de facto "integration test" (of course, this would require us to have QIIME 2 installed on Travis, which would slow down builds by a few minutes).

Thanks! Let me know if you'd like to go over this over Zoom or whatever sometime.


empress/_plot.py
```diff
@@ -30,23 +30,37 @@ def plot(output_dir: str,
          sample_metadata: qiime2.Metadata,
          feature_metadata: qiime2.Metadata = None) -> None:
 
+    # 1. Convert inputs to the formats we want
+
     # TODO: do not ignore the feature metadata when specified by the user
     if feature_metadata is not None:
         feature_metadata = feature_metadata.to_dataframe()
 
+    sample_metadata = sample_metadata.to_dataframe()
```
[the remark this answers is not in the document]
fedarko: IMO the main bottleneck with huge sample metadata in Empress will be in having the browser load it into memory, rather than in the python side of things.

That being said, yeah, it would make sense to avoid or delegate the `qiime2.Metadata` -> `pd.DataFrame` conversion if possible. I'm hesitant to make changes to this here because this code isn't part of this PR, but happy to open an issue if you'd like.

empress/_plot.py
```diff
@@ -30,23 +30,37 @@ def plot(output_dir: str,
...
+
     # create/parse tree
     tree_file = str(tree)
     # path to the actual newick file
     with open(tree_file) as file:
         t = parse_newick(file.readline())
+    empress_tree = Tree.from_tree(to_skbio_treenode(t))
```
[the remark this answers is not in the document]
fedarko: Yeah, I agree that we could make this more efficient. @ElDeveloper raised similar concerns in [this comment a while back](https://github.com/biocore/empress/pull/145#discussion_r399712873) -- I'll open an issue so we can discuss this further.

empress/_plot.py
```diff
@@ -83,24 +97,18 @@ def plot(output_dir: str,
...
-    # The original intent is to signal what
-    # columns are discrete/continous.
+    # data (i.e. NaN and "unknown") and also determines sorting order. The
+    # original intent is to signal what columns are discrete/continuous.
     # type of sample metadata (n - number, o - object)
-    sample_data_type = sample_metadata \
-        .to_dataframe().filter(feature_table.index, axis=0) \
-        .dtypes \
-        .to_dict()
+    sample_data_type = sample_metadata.dtypes.to_dict()
```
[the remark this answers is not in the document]
fedarko: Yeah, reducing the amount of conversions will make things more efficient (and likely easier to read). I'm going to lump this with #155 if that's ok.

fedarko:
Thanks for the review @antgonza!

Addressing your three comments:

> adding/running flake8 to the review the code, currently is being installed but it not being used to review the code

This is already being done -- see [the Travis log for this PR](https://travis-ci.org/github/biocore/empress/builds/672321288) as of writing. The `make stylecheck` directive (all the way down at the bottom of the log) is running `flake8` to check the python code, as well as some JS libraries to check that half of the codebase.

> adding the coverage batch in the README

This is a really good idea, and will help a lot with figuring out where the "holes" are in our test suite (e.g. #142). Opened an issue for this in #156.

> if you have the qza files in the repo, perhaps worth adding a call to generate the empress.qzv at the end of the build?

I'm down to add this, but as mentioned this will automatically increase build times by at least a few minutes (installing QIIME 2 takes some time). @ElDeveloper / @kwcantrell what do you think?

ElDeveloper:
>> if you have the qza files in the repo, perhaps worth adding a call to generate the empress.qzv at the end of the build?
>I'm down to add this, but as mentioned this will automatically increase build times by at least a few minutes (installing QIIME 2 takes some time). @ElDeveloper / @kwcantrell what do you think?

In general I agree, however I would suggest testing the QZV generation via Python. And installing QIIME2 should be fine. We can test the plugin like it's done in q2-diversity:

https://github.com/qiime2/q2-diversity/blob/master/q2_diversity/tests/test_beta_correlation.py

fedarko:
Ok, I've added an Artifact API test for the QZV generation. Currently the test just checks that the QZV was generated without errors, although we can of course add more detailed things like HTML checks in the future (that might be sort of out of scope of this PR, though).

Let me know what you think -- it's exciting to get this code formally tested!

empress/_plot.py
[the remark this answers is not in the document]
ElDeveloper: @antgonza do you know of other ways to handle metadata that would be more memory efficient? This is roughly the same way that we handle metadata in Emperor and other plugins. Like @fedarko says, the real problem is likely going to be with the feature metadata.

tests/python/test_integration.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # If the test was successful, we'll save the output QZV to this path
+        # during tearDown().
+        self.output_path = os.path.join(prefixdir, "empress-tree.qzv")
+
+    def test_execution(self):
+        """Just checks that the visualizer at least runs without errors."""
+        self.result = self.plot(self.tree, self.table, self.md, self.fmd)
+        # TODO check details of viz -- likely by digging into the index HTML of
+        # self.result.visualization, etc.
```
ElDeveloper, line 55: I would suggest doing a very "rough" check to ensure the HTML is written. Something like:

```
self.assertTrue(isinstance(self.result, q2.Visualization)
```

empress/tools.py
```diff
@@ -35,3 +39,165 @@ def read(file_name, file_format='newick'):
...
+    sf_sample_metadata = sf_sample_metadata_t.T
+
+    # Check that at least 1 sample is shared between the s. metadata and table
+    if sf_sample_metadata.shape[0] < 1:
+        raise DataMatchingError(
+            "No samples are shared between the sample metadata file and the "
+            "feature table."
+        )
+    # Report to user about any dropped samples from s. metadata and/or table
+    print_if_dropped(
```
ElDeveloper, line 187: Any motivation for using print statements instead of using `warnings.warn`? I think a warning here would make more sense, plus testing is fairly straight-forward with `assertWarnsRegex`.

empress/tools.py
```diff
@@ -35,3 +39,165 @@ def read(file_name, file_format='newick'):
...
+
+    table: pd.DataFrame
+        Representation of the feature table. The index should describe feature
+        IDs; the columns should describe sample IDs.
+
+    sample_metadata: pd.DataFrame
+        Sample metadata. The index should describe sample IDs; the columns
+        should describe different sample metadata fields' names.
+
+    feature_metadata: pd.DataFrame or None
```
ElDeveloper, line 120:
```suggestion
    tree: empress.tree.Tree
        The tree to be visualized.
    table: pd.DataFrame
        Representation of the feature table. The index should describe feature
        IDs; the columns should describe sample IDs.
    sample_metadata: pd.DataFrame
        Sample metadata. The index should describe sample IDs; the columns
        should describe different sample metadata fields' names.
    feature_metadata: pd.DataFrame or None
```

empress/tools.py
```diff
@@ -35,3 +39,165 @@ def read(file_name, file_format='newick'):
...
+
+    DataMatchingError
+        If any of the following conditions are met:
+            -No features are shared between the tree and table.
+            -No samples are shared between the sample metadata and table.
+
+    References
+    ----------
+
+    This function was based on match_table_and_data() in Qurro's code:
```
ElDeveloper, line 148:
```suggestion
    (table, sample_metadata): (pd.DataFrame, pd.DataFrame)
        Versions of the input table and sample metadata filtered such that:
            -The table only contains samples also present in the sample
             metadata.
            -The table only contains features also present in the tree.
            -The sample metadata only contains samples also present in the
             table.

    Raises
    ------
    DataMatchingError
        If any of the following conditions are met:
            -No features are shared between the tree and table.
            -No samples are shared between the sample metadata and table.

    References
    ----------
    This function was based on match_table_and_data() in Qurro's code:
```

empress/tools.py
```diff
@@ -35,3 +39,165 @@ def read(file_name, file_format='newick'):
...
+         The name of the other DataFrame which caused these items to be
+         dropped. For example, if we're checking to see if samples were
+         dropped from the sample metadata file due to to samples not being
+         in the table, df_name could be "sample metadata file" and
+         filter_basis_name could be "table".
+
+    References
+    ----------
+
+    This function was adapted from Qurro's source code:
```
ElDeveloper, line 86:
```suggestion
    ----------
    df_old: pd.DataFrame (or pd.SparseDataFrame)
         "Unfiltered" DataFrame -- used as the reference when trying to
         determine if df_new has been filtered.
    df_new: pd.DataFrame (or pd.SparseDataFrame)
         A potentially-filtered DataFrame.
    axis_num: int
         The axis in the DataFrames' .shapes to check. This should be either
         0 or 1, but we don't explicitly check for that.
    item_name: str
         The name of the "thing" described by the given axis in these
         DataFrames. In practice, this is either "sample" or "feature".
    df_name: str
         The name of the DataFrame represented by df_old and df_new.
    filter_basis_name: str
         The name of the other DataFrame which caused these items to be
         dropped. For example, if we're checking to see if samples were
         dropped from the sample metadata file due to to samples not being
         in the table, df_name could be "sample metadata file" and
         filter_basis_name could be "table".

    References
    ----------
    This function was adapted from Qurro's source code:
```

empress/tools.py
```diff
@@ -35,3 +39,165 @@ def read(file_name, file_format='newick'):
...
+    # the whole tree (that being said the Empress UI supports just showing
+    # features in the table, anyway)
+    ff_table = table.loc[tree_and_table_features]
+
+    # Report to user about any dropped samples from table
+    print_if_dropped(table, ff_table, 0, "feature", "table", "tree")
+
+    # Match table and sample metadata
+    sample_metadata_t = sample_metadata.T
+    sf_ff_table, sf_sample_metadata_t = ff_table.align(
```
ElDeveloper, line 173: Does this mean that samples in the feature table not present in the metadata are going to be dropped? If so, I would suggest only doing this if the user explicitly asks for this, for example with a flag `--ignore-missing-samples` or something along those lines.

tests/python/test_integration.py
```diff
@@ -0,0 +1,70 @@
+# ----------------------------------------------------------------------------
```
ElDeveloper, line 1: This is amazing, thanks for putting this @fedarko!

empress/_plot.py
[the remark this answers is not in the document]
antgonza: Good question! I was thinking in not loading the full  sample metadata dataframe into memory but use the qiime2 methods to retrieve what you need; for example: 
Metadata.get_column
MetadataColumn.to_series
etc 

However, not sure if this will help the footprint or not ...

tests/python/test_integration.py
fedarko: Added a check for this, thanks!

empress/tools.py
fedarko: I don't have a particularly good motivation for this -- printing was just how I did this in Qurro :)

It's possible to get `warnings.warn()` to go to stdout, right? I'd want to make sure that (if the user passes `--verbose` to Empress) they see these messages show up.

empress/tools.py
fedarko: Yes, this means that the default behavior of Empress right now is "filtering" the table and sample metadata to just the shared samples.

Just to check, would you be ok with the following solution:

- By default, if any samples in the table have no metadata (or any samples in the metadata are not in the table), raise an error explaining the situation
- If the user passes a `--filter-missing-samples` flag or something, do the current behavior (filter to samples shared btwn. table and metadata, so long as there's at least 1 such sample)

This would be slightly different from Emperor's `--ignore-missing-samples` flag, hence the slightly different name to avoid confusing users.

We could also implement an analogue to what Emperor's `--ignore-missing-samples` flag does, where we allow for extra samples in the metadata to *not be in the table* by default, but raise errors that need to be manually overridden when the table contains samples not in the metadata. However I'm not sure this would be super useful, because samples without metadata are kind of useless in the current visualization IMO? Unlike in Emperor, where those samples are still "represented" in the visualization. The first solution seems more intuitive to me for Empress' utility.

...Sorry for the rambles -- we can talk more about this over a call if you'd prefer :)

empress/_plot.py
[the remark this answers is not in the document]
ElDeveloper: Gotcha, my understanding is that `Metadata` parses and loads the whole mapping file into memory.

empress/tools.py
ElDeveloper: Gotcha. Yes, it is possible to see warnings with the `--verbose` flag.

empress/tools.py
ElDeveloper: Thanks for the explanation @fedarko.

> We could also implement an analogue to what Emperor's --ignore-missing-samples flag does, where we allow for extra samples in the metadata to not be in the table by default, but raise errors that need to be manually overridden when the table contains samples not in the metadata.

The motivation is to always show the users what samples are lacking metadata, and do so clearly. For example in Emperor when a sample does not have metadata and the user selectes to `--ignore-missing-samples` these samples are padded with "placeholder metadata" that shows what is missing. For example all columns in the metadata for those samples would have "This sample has no metadata". I think it's better to implement something analogous to `--ignore-missing-samples`.

fedarko: Got it. I can implement something analogous to what Emperor does -- will take a bit of extra time, but shouldn't be too bad :) Just to double check, this means that samples that are in the metadata but not in the table are dropped by default? (I can set things up to warn the user about these samples but still not do anything.)

Agreed that consistency here would be good, especially between the two Emp[eror|ress] tools :) As a heads up, Qurro does things a slightly different way: it filters its inputs so that all the samples that remain are the "matches" between the table and metadata, outputting warnings/print messages about these operations as needed. (...If I could go back a year and change things, I'd probably make this more similar to how Emperor works. Sorry!)

ElDeveloper: Yes, in general the metadata and tree can be considered a supersets of the feature table, but not the other way around. The flow should be:

```
if there's tips in the table that are not present in the tree:
    raise an error and allow to override with "filter missing features" flag
    # when the user sets the "filter missing features" flag, these features
    # are removed from the table.
if there's samples in table not present in the metadata:
    raise an error and allow to override with "ignore missing samples" flag.
    # when the user sets the "ignore missing samples" flag these sample's
    # metadata is padded with a message "This sample has no metadata".
```

For reference, here's how [things are done in Emperor](https://github.com/biocore/emperor/blob/master/emperor/core.py#L306).

Regarding Qurro, there's always a chance for a new release :) :shipit:

fedarko: Thanks for the detailed writeup! This makes things clearer. I'll try to get to this soon. Will also see if I can add https://github.com/biocore/qurro/issues/296 in for the next Qurro release ;)

fedarko:
@ElDeveloper Sorry for the wait! I've updated the matching behavior to more closely resemble Emperor's, and I've added decently thorough tests that check the related cases. 

Some related screenshots, for fun:

### One of the possible errors

image

### Some possible warnings (shown when you use the `--verbose` flag)

image

### Placeholder metadata in practice

image

These changes should take care of things -- at least until we add in feature metadata support to Empress. (And feature metadata is an optional argument to `match_inputs()`, so we shouldn't have to modify these tests at all when we add in feature metadata matching in the future. (Knock on :evergreen_tree:, though.))

Thanks!

empress/tools.py
```diff
@@ -35,3 +45,186 @@ def read(file_name, file_format='newick'):
...
+    # below, so they'll cancel out. (I hope that makes sense?)
+    dropped_sample_ct = sample_metadata.shape[0] - sf_sample_metadata.shape[0]
+    if dropped_sample_ct > 0:
+        warnings.warn(
+            (
+                "{} sample(s) in the sample metadata were not present in the "
+                "table. These sample(s) have been removed from the "
+                "visualization."
+            ).format(dropped_sample_ct),
+            DataMatchingWarning
```
ElDeveloper, line 228: In general we do not want to warn users that there are more samples in their metadata than there are their data tables/trees. I don't think we should warn when this happens as this is a rather frequent scenario.

fedarko: Ok, latest commit should address this. Thanks!

ElDeveloper:
Thanks so much @fedarko!

PR #159, opened 2020-04-24
Handle categorical colorschemes correctly; add tests for colorer.js; fix a tiny bug

### LINKED ISSUE #137: "Classic QIIME Colors" are incorrectly assigned
The hex codes [here](https://github.com/biocore/empress/blob/0f4c98d348ade522474408692551859f99eda10e/empress/support_files/js/colorer.js#L34-L59) match up with Emperor as far as I can tell, but when you use this color scheme in Empress the colors look pretty different -- some of the colors are shades of grayish brown which (esp. when line widths are all 1) are pretty hard to distinguish from uncolored branches.

issue1

Ideally it would be good to have the colors be consistent with Emperor, since I imagine this is probably the "default" color scheme people using Empress will try first.

Also, if possible, it would be good for colors to be consistent across different amounts of fields -- e.g. when you select `barcode-sequence` as the field to color by instead of `body-site`, the first few "Classic QIIME" colors after red are different. For discrete color schemes, at least, this behavior seems counterintuitive to me (although I can totally understand how it'd be useful for quantitative colorschemes like viridis). Looks like the origin of this behavior is [in this function](https://github.com/biocore/empress/blob/0f4c98d348ade522474408692551859f99eda10e/empress/support_files/js/empress.js#L323-L338)?

issue2

fedarko:
Closes #137.

### Summary
This PR changes `colorer.js` so that all colorschemes are treated as categorical, replacing the old behavior of treating all colorschemes as quantitative ([details here](https://github.com/biocore/empress/issues/137#issuecomment-601980132)). **If desired, I can modify this PR (or we can open an issue for later) to support toggling between categorical / quantitative colorschemes;** I wasn't sure if we wanted to focus on that for now, or how the UX for that functionality would work, which is why I'm making this PR sooner rather than later :)

### Other stuff added in this PR

- Fairly comprehensive tests for `colorer.js` (#142).
  - The `addColorsToSelect()` method is still untested (I don't _think_ there are any JS tests yet that create HTML elements for use in tests; I wanted to hold off on testing this for now so I could pick your brains on how best to do this), but the rest of the functions in the class are covered.

- Fixed a minor bug in how RGB colors were computed.
  - Long story short, `Colorer.getColorRGB()` produces an array of `[r, g, b]`, where each of the three  elements in the array is a number in the range `[0, 1]`. As an intermediate step, the code was generating an array of elements in the range `[0, 255]`, but was then converting these to `[0, 1]` by dividing by `256` instead of by `255`.
  - This introduced a small amount of imprecision, since e.g. `#ffffff` would've been converted to `[255, 255, 255]` and then to `[0.996, 0.996, 0.996]` instead of to `[1, 1, 1]`.
  - [At first I fixed this by changing the denominator to `255` instead of `256`](https://github.com/biocore/empress/blob/49941123338ef6f571d208a15b5611dc829ed9ac/empress/support_files/js/colorer.js#L32-L39), and then while I was writing this PR message out I wondered "wait does chroma already have something that does all of this" and then it turned out that yeah it does in [chroma's `color.gl()`](https://gka.github.io/chroma.js/#color-gl) method, so now I feel silly :sweat_smile: **Edit: see comment below about this method**
  - Anyway, now that `getColorRGB()` is tested, we shouldn't need to worry about this bug again.

#### Default QIIME colorscheme, coloring by body site

image

#### Accent colorscheme, coloring by barcode sequence

This demonstrates the "looping" feature, where (when the number of categories for a metadata field is greater than the number of colors in the palette) things "loop" around.

image

#### Viridis colorscheme, coloring by barcode sequence

This demonstrates that quantitative colorschemes are currently treated the same way as categorical  colorschemes, which results in funky behavior -- the "looping" looks kind of off here. (The reason Viridis loops after 9 "items" is that this is just how the underlying [`chroma.brewer["Viridis"]`](https://gka.github.io/chroma.js/#chroma-brewer) array is structured.)

image

This behavior should likely be changed at some point, either by removing the quant color schemes for now, adding some UI elements to toggle between categorical/quantitative color scales and adjusting the code accordingly, or some third option that I feel like I'm forgetting right now.

image

Thanks!


empress/support_files/js/colorer.js
```diff
@@ -1,56 +1,50 @@
 define(["chroma"], function (chroma) {
-    // class globals closure variables
```
ElDeveloper, line 2: @fedarko, when given `N` categories and a sequential or diverging color palette the Colorer should provide `N` unique colors that are interpolated from the provided color palette. Chroma has the ability to [do this](https://github.com/biocore/emperor/blob/master/emperor/support_files/js/color-view-controller.js#L415).This would make it such that for viridis the color values didn't repeat, but for discrete color palettes the colors repeated.

tests/test-colorer.js
```diff
@@ -0,0 +1,112 @@
+require(["jquery", "chroma", "Colorer"], function ($, chroma, Colorer) {
+    $(document).ready(function () {
+        module('Colorer');
+        test("Test that default QIIME colors are correct", function () {
+            // I copied this in from https://github.com/biocore/emperor/blob/659b62a9f02a6423b6258c814d0e83dbfd05220e/emperor/support_files/js/color-view-controller.js#L624,
+            // so this lets us guarantee that (at least in terms of the default
+            // discrete color values) Emperor and Empress are consistent >:)
```
ElDeveloper, line 7: :shipit: 🥳

empress/support_files/js/colorer.js
fedarko: @ElDeveloper Got it. My worry with this is that categories are currently sorted inconsistently throughout Empress: see the legend below, where `84` is included [at the end of `categories`](https://github.com/biocore/empress/blob/319af68654f8592f3a8c0b569ae85985bd70e9bb/empress/support_files/js/empress.js#L627) (resulting in it getting assigned the "highest" color), but the category "sorting" in the legend doesn't reflect this, resulting in something that would look weird even if we interpolate based on the color palette. (This problem also impacts discrete color schemes, but it's less noticeable there.)

image

I can still make sequential/diverging palettes work well for categorical data using the Chroma code from Emperor you linked (thanks!), and I can try to fix this issue to make sorting at least consistently lexicographic throughout Empress, but my concern is that lexicographic sorting is incorrect for numeric data, which could lead to users misinterpreting the resulting color schemes.

```python
>>> a = ["0", "84", "112", "140", "168"]
>>> sorted(a)
['0', '112', '140', '168', '84']
>>> sorted([int(x) for x in a])
[0, 84, 112, 140, 168]
```

My preference would be for just completely excluding sequential/diverging color schemes for categorical data, and only allowing these color schemes for quantitative data... but I know Emperor does things differently, so if we need to be consistent that's ok.

ElDeveloper: Thanks for thinking about this edge case. I had entirely forgotten how big of a nightmare this is. The solution in Emperor is to sort categories with a `naturalSort` implementation [we wrote](https://github.com/biocore/emperor/blob/master/emperor/support_files/js/util.js#L12). This makes sure numbers are sorted as numbers, and if there's a mixed type then numbers appear at the beginning of the list.

I like the idea of automatically using continuous categories with numeric data, however I've found that it's better to let users "cast the data on the fly". Here's a demo of how that is handled in Emperor:

continuous

When the user selects to cast data but there are numbers and missing data, all "non-numeric" values are colored in gray. All of this is only possible if a continuous or diverging colormap is selected.

For this PR I would suggest porting over the naturalSort function and using that to preprocess data before assigning colors to the values. The "on the fly casting" can be left for a future PR. The important here right now is to have usable colors, and we are almost there.

fedarko: Thanks for the thorough explanation! This sounds good to me. I'll get to work on porting `naturalSort` over to Empress and using it as you recommended. :shipit:

ElDeveloper: 👍 thanks, not sure if it would be overkill but we might also want to port the tests? Up to you. https://github.com/biocore/emperor/blob/master/tests/javascript_tests/test_util.js

fedarko: Agreed, porting the tests is a good idea -- shouldn't be too much extra effort and it'll be good to make sure my copypasta is correct :) Thanks @ElDeveloper!!!

ElDeveloper: Thanks!

fedarko:
just for reference, ran into a weird bug that I actually introduced a few days ago: this was caused by calling `chroma.color.gl()`, which actually produces an array with 4 components (red, green, blue, opacity). My test code didn't see a problem with this due to the way it was written, and everything seemed to work properly... but this messes up the drawing in Empress, because the drawing code mixes coordinates and colors in the same array (with the pattern `[x, y, r, g, b, ...]`), so having more than exactly 3 components for each color causes weird stuff like this:

image

... Anyway the bug is fixed now, and the tests have been updated accordingly :sweat_smile: All that's left on this PR should just be doing the interpolation when seq / diverging colormaps are selected, I think.

fedarko:
OK, requested changes have been made! Sorry for the delay. Now, Empress does Emperor-esque interpolation if a sequential / diverging color map is selected and the selected metadata field contains at least 2 numeric values. `naturalSort()`, `splitNumericValues()`, and their corresponding tests have been ported over from Emperor to a new `util.js` module for Empress.

Scaling when a field has all numeric values:
image

Scaling when a field has _partially_ (but at least 2) numeric values: (non-numeric values are displayed at the start of the list and are colored the same shade of gray as with Emperor.)
image

When a field has less than 2 numeric values, the user is shown an alert message...
image
... and then all colors are set to the Emperor gray color:
image

To support these changes, the `colorer` codebase has been pretty significantly overhauled. One of the benefits of this was it let us remove `Empress._assignColor()` entirely, which previously worked by creating a new `Colorer` object twice every time `colorBySampleCat()` was called -- now `empress.js` only makes one of these objects at a time (using the same object to produce both the RGB and Hex mappings), so stuff should be faster.

This comment is already getting too long, but I should mention -- I updated the chroma.js version in the codebase to v2.1. The previous version looked to be ~4 years out of date (its copyright range ended in 2015) -- there were small differences in the hex colors returned between the two versions for some of the interpolated values (like, we're talking off by ~1/255ths of a hex color, so not really noticeable but enough to break the tests). I think this is attributable to small precision fixes by the chroma.js team [in v1.3.3](https://github.com/gka/chroma.js/blob/master/CHANGELOG.md#133), which going by GitHub releases was released in late 2017. **It would probably be a good idea to update the chroma.js version [used in Emperor](https://github.com/biocore/emperor/blob/master/emperor/support_files/vendor/js/chroma.min.js) / other tools** to avoid this problem.

I think that's it for now! Things aren't perfect (for example, `util.naturalSort()` is called multiple times throughout the code -- I think it would be faster and more foolproof to just sort a category's values once and then pass the sorted array throughout the various drawing functions), but this should be a pretty decent improvement (and #137 should be close-able now at least :)

empress/support_files/js/colorer.js
```diff
@@ -1,56 +1,166 @@
...
+     * this will set every value with Colorer.NANCOLOR and set
+     * this.notEnoughNumericValues to true. Otherwise, this will properly
+     * scale numeric values according to the color map (while setting any
+     * non-numeric values to Colorer.NANCOLOR).
      */
-    Colorer.prototype.getColorRGB = function (color) {
-        return this.__colorer(color)
-            .rgb()
-            .map((x) => x / 256);
+    Colorer.prototype.assignScaledColors = function () {
```
ElDeveloper, line 88: @fedarko it looks like this is merging two different types of behavior. Lets call them continuous coloring mapping (**ccm**), and continuous coloring without mapping (**ccwm**). We don't really want these two together. To mimic emperor we would want these separately.

First, we want for sorted arrays to be able to be placed on the continuous gradient with non-numeric sorted values at the beginning and then number values sorted at the end (**ccwm**). In this case non-numeric values are assigned colors other than gray.

Second, we also want to be able to map each value to a color in the scale, and group all non-numeric values as a single "NA" value (**ccm**). In this case we want to be able to see a gradient colorbar and not the individual values, as well as a single element with a gray box (this is what happens when you click the continuous values checkbox in Emperor).

The portion of code you are missing for **ccwm** is here:
https://github.com/biocore/emperor/blob/master/emperor/support_files/js/color-view-controller.js#L463

The version you've implemented is closer to **ccm**, but would be missing the part where a colorbar is drawn and all non-numeric values are lumped into one. To implement that, feel free to reuse from [`color-view-controller.js`](https://github.com/biocore/emperor/blob/master/emperor/support_files/js/color-view-controller.js).

Let me know if you would rather have a quick zoom call since this is kinda complicated eitherway.

fedarko: Thanks for the information and for taking the time to talk with me about this!

For now, as discussed, I've implemented the **ccwm** option (basic ordinal scaling). This assigns values to colors from a sequential / diverging color map by using the first sorted value as the "lowest" value in the color map, using the last value as the "highest" value in the color map, and just interpolating between:

```js
    Colorer.prototype.assignOrdinalScaledColors = function () {
        var interpolator = chroma
            .scale(chroma.brewer[this.color])
            .domain([0, this.sortedUniqueValues.length - 1]);

        for (var i = 0; i < this.sortedUniqueValues.length; i++) {
            var val = this.sortedUniqueValues[i];
            this.__valueToColor[val] = interpolator(i);
        }
```

In practice this looks like:

image

Non-numeric and numeric values are treated in a way that is consistent with Emperor's expected behavior:

image

The work done on the **ccm** option remains untouched in the `color-numeric-scaling` branch [here](https://github.com/fedarko/empress/tree/color-numeric-scaling), for reference.

I also realized that my changes to `Colorer` broke the animation functionality, so included in this round of changes are fixes to `animator.js` accordingly. (Tagging @kwcantrell -- I *think* the animation looks correct, but I'd like to get your feedback on Monday to make sure I'm not breaking anything.)

fedarko:
one last thing, fixed a small corner case where all values in a field are the same. This results in a single color being used, and to match Emperor the color that's used is the "first" from a given palette:

image

tests have been added accordingly

ElDeveloper:
This looks great, thanks @fedarko!

Issue #141, opened 2020-03-23
Add a marker to show where the root is for rooted trees


ElDeveloper:
We should include and document a marker for where the root of the tree is located. This should be added after some of the layout changes make it into the repo.

fedarko:
May be worth considering the "hide non-sample features" option, and how that interacts with this. As part of the #139 fixes, we will be able to safely assume that the root / a descendant of it is always present in at least one sample, right?

fedarko:
Updates on this, for reference: #160 added a special marker for the root but it wasn't selectable\*, and as of b1ce0be0d01dbe39b70960a4705b55a8874acfcb (for #169) the root node is now both marked and selectable. So when the feature metadata branch is merged in, I think this can be closed?

\* (Also I'm a little ashamed to admit that I just spent like 15 minutes digging through the code and I still don't know where this marker is actually being set.)

PR #160, opened 2020-05-04
Merge pull request #30 from kwcantrell/node-hover

kwcantrell:
Node hover

A couple this to note for this PR. 
Currently, only nodes with ids in the newick tree will get a visible nodes associated with them. However, you can still select nodes that do not have a visible node as seen here.
nodes

selecting none visible node:
select

As you can see from the image above, I assign a generic name to all unnamed nodes (i.e. EmpressNodei) and the default behavior is to display this id. For unnamed nodes, would it be better to hide the id or show a blank?

@fedarko, what version of the js lint are you using? I pull the most recent master branch but it appears we are still using different versions.


fedarko:
For reference, I'm using prettier v2.0.1 and jshint v2.11.0.

empress/support_files/css/empress.css
```diff
@@ -312,6 +312,21 @@ label:hover .tooltip {
...
+.hover-select-container:after {
+    position: absolute;
+    content: "\203A";
+    transform: rotate(90deg);
+    color: gray;
+    font-size: 13.5pt;
+    right: 12px;
+    top: 1px;
+    pointer-events: none;
+}
```
fedarko, line 328: I'd suggest changing the color of this select to make it stand out more (or alternately changing the background of the hover box to a shade of gray similar to what's used in the side panels). Up to you, though.

empress/support_files/js/canvas-events.js
```diff
@@ -0,0 +1,168 @@
...
+            transVec = gl.vec3.fromValues(curPos[0], curPos[1], curPos[2]);
+            transMat = gl.mat4.create();
+            gl.mat4.fromTranslation(transMat, transVec);
+            gl.mat4.multiply(drawer.worldMat, transMat, drawer.worldMat);
+
+            // draw tree
+            drawer.draw();
+        };
+
+        var mouseClick = function(e) {
```
fedarko, line 121: (wasn't sure where to put this so I'm mentioning it here)

It doesn't seem like it's possible to click on the root node, even though it's marked in the display. It would be nice (IMO) to make it clickable, just for consistency's sake if nothing else. (Tagging #141 for reference, and in case doing this would be too much extra work for right now.)

empress/support_files/js/canvas-events.js
```diff
@@ -0,0 +1,168 @@
...
+        this.empress = empress;
+        this.drawer = drawer;
+        this.canvas = canvas;
+
+        // the current mouse coordinates in screen space
+        this.mouseX = null;
+        this.mouseY = null;
+        this.mouseMove = false;
+
+        this.selectedNode = new SelectedNodeMenu(this.empress, this.drawer);
```
fedarko, line 12: Suggest renaming this `selectedNodeMenu` or something instead -- without context I assumed this was referring to an object describing the currently selected node

empress/support_files/js/canvas-events.js
```diff
@@ -0,0 +1,168 @@
...
+                var nY = empress.getY(closeNode);
+                var screenSpace = drawer.toScreeSpace(nX, nY);
+                nX = screenSpace.x;
+                nY = screenSpace.y;
+                xDist = e.clientX - nX;
+                yDist = e.clientY - nY;
+                var screenDist = Math.sqrt(xDist * xDist + yDist * yDist);
+                if (screenDist < epsilon) {
+                    selectedNode.setSelectedNode(closeNode);
+                    empress.drawTree();
```
fedarko, line 157: Do you think it might be feasible to make the search menu call this two-line block if the user enters in an exact node ID? This could be a useful way to highlight specific tips on the tree that the user knows of beforehand.

empress/support_files/js/select-node-menu.js
```diff
@@ -0,0 +1,166 @@
...
+            var val = selectMenu.sel.value;
+            selectMenu.addMenuColumn(val, "s");
+            selectMenu.showNodeMenu();
+        };
+        this.addBtn.onclick = click;
+    };
+
+    SelectedNodeMenu.prototype.addMenuColumn = function(colName, type) {
+        this.cols.push({ col: colName, type: type });
+    };
```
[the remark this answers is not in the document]
kwcantrell: Thanks for catching that! In terms of the UI do you think it would be better to remove metadata fields from the drop down menu once they are added or just ignore them if they are already added?

empress/support_files/css/empress.css
```diff
@@ -250,6 +250,8 @@ label:hover .tooltip {
     background-color: white;
     opacity: 0.75;
     padding: 20px;
+    max-width: 500px;
+    overflow: auto;
 }
```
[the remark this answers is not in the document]
kwcantrell: Thanks, I set a max-width but for some reason not a max height. Do you know if there are any properties that can be used to scroll a table when it extends beyond the window but still within the max-height bounds? Or is that something that has to be done through javascript? ((I not too familiar with css)

empress/support_files/templates/empress-template.html
```diff
@@ -33,6 +33,15 @@
     <!-- hover box -->
     <div class="hidden" id="hover-box">
       <table id="hover-table"></table>
+      <div>
+        <p>
+          <br><b>Select a category to add to table</b>
```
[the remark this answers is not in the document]
kwcantrell: Good to know!

empress/support_files/js/empress.js
```diff
@@ -174,13 +184,37 @@ define(["underscore", "Camera", "Drawer", "Colorer", "VectorOps"], function (
...
+     * @return {Array}
+     */
+    Empress.prototype.getNodeCoords = function() {
+        var tree = this._tree;
+        var coords = new Float32Array(tree.size * 5);
+        var coords_index = 0;
+
+        for (var i = 1; i <= tree.size; i++) {
+            var node = this._treeData[i];
+            if (!node.name.includes("EmpressNode")) {
```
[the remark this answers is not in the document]
kwcantrell: It probably would be a good idea to handle this edge case. Do you think it would be better to handle this issue in the python or JS side?

@ElDeveloper do you have any thoughts?

empress/support_files/js/biom-table.js
```diff
@@ -210,5 +237,55 @@ define([], function () {
...
+     *
+     * @param{Array} samples A list of sample ids
+     * @param{String} cat The category to count
+     *
+     * @return{Object}
+     */
+    BIOMTable.prototype.getSampleValuesCount = function(samples, cat) {
+        var result = {};
+        for (var i = 0; i < samples.length; i++) {
+            var sVal = this._samp[samples[i]][cat];
```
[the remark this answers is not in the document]
kwcantrell: Good point, I'll change both to fVal (short for field value).

empress/support_files/css/empress.css
kwcantrell: We change the hover box to gray to match the side panels.

empress/support_files/js/canvas-events.js
kwcantrell: I'll have to look a bit into this issue. It looks like you can click the root node in radial layout but not in the rectangular layout. I think the issue arises because we introduce an artificial "root" node in the rectangular layout so we I scan the tree I do see that node (since I use the BPTree object).

empress/support_files/js/canvas-events.js
kwcantrell: That is a good suggestion! I think it would be nice to be able to manually search for nodes by ID!

empress/support_files/js/empress.js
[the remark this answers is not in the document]
ElDeveloper: I don't think it's a pressing thing to address, it would probably be easy to add a check for named notes on the python side of things i.e. if any node is labeled EmpressNode then raise a warning/error? This could also be a more improbably string like `__en__1__`, or something like that?

empress/support_files/js/select-node-menu.js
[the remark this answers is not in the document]
fedarko: Ooh, that's a tough one. I think either approach would make sense, so my instinct here is to say that whatever's easiest is probably best for now :)

For the long term, I think removing added fields from the dropdown menu would be a good option, but it would be best to accompany that with a way to remove fields from the table (which would add them back into the dropdown menu). But this is just my perspective -- tagging @ElDeveloper in case he has any thoughts on this.

empress/support_files/css/empress.css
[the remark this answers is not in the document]
fedarko: I _think_ it might be possible to finagle something like that using CSS media queries that take into account the window size and the hoverbox position (see e.g. [here](https://www.w3schools.com/cssref/css3_pr_mediaquery.asp)), but that seems like it might be a pain.

An idea that might be easier (and more intuitive for users) is having the JS code that displays the hoverbox check to see if it would go outside the screen, and if so alter the position of the hoverbox to stay inside the screen. (This would function analogously to how, when you "right click" on most windowing systems, the menu that pops up always remains on the screen even if it'd go off the screen normally.) I think [this stack overflow post](https://stackoverflow.com/a/8897628/10730311) might be helpful with doing this check, but LMK if you want to talk more about this.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: :+1: on just raising an error if the user has nodes named `EmpressNode` in their tree.

empress/support_files/js/canvas-events.js
fedarko: Oh, huh, that's funky. No worries -- as mentioned this isn't a huge issue.

empress/support_files/js/select-node-menu.js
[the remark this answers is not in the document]
ElDeveloper: > For the long term, I think removing added fields from the dropdown menu would be a good option

^ I think removing added fields from the dropdown menu would be the best starting point. 👍

ElDeveloper:
@kwcantrell, just a heads up, there's a few conflicts with the master branch. Likely due to the recent changes that came in from #159

kwcantrell:
Fixed merge conflicts, added tests, and changed the scroll behavior of the table. I just need to finish the autocomplete features.

tests/test-biom-table.js
```diff
@@ -0,0 +1,248 @@
...
+
+            teardown: function() {
+                this._obs = null;
+                this._samp = null;
+                this._types = null;
+                this.biomTable = null;
+            }
+        });
+
+        test('Test getObjservationUnionForSamples', function() {
```
ElDeveloper, line 69:
```suggestion
        test('Test getObservationUnionForSamples', function() {
```

empress/support_files/js/canvas-events.js
```diff
@@ -0,0 +1,214 @@
...
+        var newX = 0,
+            newY = 0;
+
+        // moves tree as long as mouse is pressed down
+        var moveTree = function (e) {
+            // set move flag
+            events.mouseMove = true;
+
+            // grab mouse location
+            var center = $(window).width() / 2;
```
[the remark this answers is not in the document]
kwcantrell: Thanks for catching this. This was old code I had copied.

kwcantrell:
Changes made:

- Selected nodes now appear larger to make them standout more.

- An aggregated table is created when user searches for an id that matches multiple nodes. When this happens a warning message is also displayed below the ID. Also, all match nodes are highlighted.
Screenshot from 2020-05-20 07-34-29

- A brief description was added to the table.

- If a user searches for an invalid id then the background color of the quick-search bar becomes red.

- As the user types in the quick-search bar, a dynamic autocomplete menu is created.

empress/support_files/js/canvas-events.js
```diff
@@ -163,51 +191,206 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
...
+     */
+    CanvasEvents.prototype.autocomplete = function (ids) {
+        // need for use in closures
+        var autocompleteContainer = this.autocompleteContainer;
+        var quickSearchBar = this.quickSearchBar;
+        var canvasEvents = this;
+        var searchBtn = this.quickSearchBtn;
+
+        var createClickEvent = function (e) {
+            var nodeId = this.getElementsByTagName("input")[0].value;
```
ElDeveloper, line 221: Can you name the element so that we only query for the element we need not for the first input element in the document?

empress/support_files/js/select-node-menu.js
```diff
@@ -96,78 +138,194 @@ define([], function () {
...
+        }
+        // create object that will map fields to all of their possible values
+        var field,
+            fieldValues,
+            fieldValue,
+            fieldsMap = {},
+            i,
+            j,
+            k;
+        for (i = 0; i < this.fields.length; i++) {
```
ElDeveloper, line 171:
```suggestion
        for (var i = 0; i < this.fields.length; i++) {
```

empress/support_files/js/select-node-menu.js
```diff
@@ -96,78 +138,194 @@ define([], function () {
...
+            field = this.fields[i];
+            var possibleValues = this.empress._biom.getUniqueSampleValues(
+                field
+            );
+            for (j = 0; j < possibleValues.length; j++) {
```
ElDeveloper, line 176:
```suggestion
            for (var j = 0; j < possibleValues.length; j++) {
```

empress/support_files/js/select-node-menu.js
kwcantrell: I had defined i,j, and k at the start of the function because jshint was giving me issues when I defined them inside the for-loops you suggested.

empress/support_files/js/select-node-menu.js
kwcantrell: See above comment

empress/support_files/js/select-node-menu.js
ElDeveloper: Thanks so much, my bad!

empress/support_files/js/canvas-events.js
kwcantrell: I removed the input and named the div container instead. The input was just used to store the suggested node id however, the same thing can be accomplished by setting the div container id to the suggested node id.

kwcantrell:
@ElDeveloper and @fedarko thanks for the feed back. I believe I have addressed all of your comments.

empress/support_files/js/canvas-events.js
ElDeveloper: Perfect!

ElDeveloper:
Thanks @kwcantrell, this looks great. Also thanks @fedarko for the reviews!

PR #162, opened 2020-05-25
Split the screen in half

kwcantrell:


ElDeveloper:
Thanks for the help @kwcantrell, let's close this for the time being. That being said, the code is representative of what we need to split the screen in half.

Issue #152, opened 2020-04-06
Figure out what's up with internal nodes being colored but children not colored during animation


fedarko:
![image](https://user-images.githubusercontent.com/4177727/78597139-d1e57e00-7801-11ea-8526-1cdeb8ba17c1.png)

May be an artifact of weird input data or of animation/coloring stuff.

fedarko:
I think I've identified the cause of this. As far as I can tell, this has to do with `Empress._projectObservations()`:

https://github.com/biocore/empress/blob/6e4bbd3cc53a79494ebe05b8db174086f02169a4/empress/support_files/js/empress.js#L629-L649

For reference, my understanding of how this function works is that it accepts an Object  (`obs`) mapping categories (e.g. `body_site`) to nodes in the tree present within these categories. This then goes through the tree in a post-order traversal, expanding `obs` by defining each node N's parent P to be present in the same group that N is in, if any. This will obviously not create unique assignments for nodes that are present in multiple categories / have descendants present in multiple categories, so there's a call to `util.keepUniqueKeys()` at the end that filters each category to just those nodes unique to it.

**The problem with this approach** is that it does not, as far as I can tell, consider "not in any of the categories" to itself be a category. So if you have a simple tree like

```
      + --- B
      |
A --- + 
      |
      + --- C
```

where B is in category "1" and C isn't in any categories, then A will still be considered by `_projectObservations` to be unique to category "1":

```
      + === B
      #
A === + 
      |
      + --- C
```

I guess the way to move forward with this bug is first acknowledging -- this actually is a bug, right? I can see how you might make an argument for this being "desired" behavior in some cases (my personal opinion is that this is a bug, but I might be biased because I just spent a while tracking this down :).

If we agree that this is a bug, then I'm not sure how it would be best to fix this. We could maintain a separate data structure during the postorder traversal of "nodes that aren't in any category" and use that analogously to an entry in `obs`, I guess? Not sure.

In any case, we should definitely add tests for this function so this problem doesn't happen again. My personal vote is for porting it over to the `BPTree` module, which unlike `empress.js` already has tests set up (so adding more tests shouldn't be too much work ... hopefully.)

kwcantrell:
Good catch @fedarko! You are correct, _projectObservations does not take into account tips that do not belong to a unique category. 

[...]

PR #174, opened 2020-06-15
Add Empress class and add support for OrdinationResults in q2-empress

ElDeveloper:
This PR moves over the code from _plot.py into a new class `Empress`. In addition this allows support for visualizing ordinations and phylogenies simultaneously. Here's an animated gif of this working with brush selection (related to https://github.com/biocore/emperor/pull/766).

empress-pr

The example above shows how selecting a group of samples of different colors shows the unique branches per color (similar to UniFrac comparing groups). When a single sample or group is selected, all the features for that group are highlighted in Empress.

`qiime empress plot` now accepts an optional ordination object.
The majority of the code for the Empress class was inspired by the Emperor object.

### DIFF
--- a/empress/core.py
+++ b/empress/core.py
@@ -46,9 +46,9 @@ class Empress():
             ``ordination`` object, should have an index set and it should match
             the identifiers in the ``ordination`` object.
         feature_metadata: pd.DataFrame, optional
-            DataFrame object with the metadata associated to the features in
-            the ``table`` and ``tree`` objects, should have an index set and it
-            should match the feature identifiers.
+            DataFrame object with the metadata associated to the names of
+            tips and/or internal nodes in the  ``tree`` object, should have an
+            index set and it should match at least one of these nodes' names.
         ordination: skbio.OrdinationResults, optional
             Object containing the computed values for an ordination method in
             scikit-bio. Currently supports skbio.stats.ordination.PCoA and
--- a/empress/core.py
+++ b/empress/core.py
@@ -9,6 +9,7 @@
 from empress.tree import Tree
 from empress.tools import name_internal_nodes, match_inputs
 
+import warnings
 import pkg_resources
 import os
 import pandas as pd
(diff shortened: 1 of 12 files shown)


empress/support_files/js/empress.js
```diff
@@ -546,10 +546,58 @@ define([
...
+
+        for (group in observationsPerGroup) {
+            obs = Array.from(observationsPerGroup[group]);
+
+            // convert hex string to rgb array
+            var rgb = [
+                parseInt(group.slice(0, 2), 16) / 255,
+                parseInt(group.slice(2, 4), 16) / 255,
+                parseInt(group.slice(4), 16) / 255,
+            ];
```
fedarko, line 593: Suggest delegating this to chroma.js rather than doing it manually -- `chroma(color).gl().slice(0, 3);` is how the `Colorer` module in Empress converts hex to RGB.

Of course the code is still correct as is, so this doesn't matter too much...

empress/core.py
```diff
@@ -0,0 +1,335 @@
...
+        emperor_div = '\n'.join(html[39:44])
+
+        # main js script for emperor
+        emperor_require_logic = '\n'.join(html[45:-3])
+
+        # once everything is loaded replace the callback tag for custom JS
+        with open(SELECTION_CALLBACK_PATH) as f:
+            selection_callback = f.read()
+        emperor_require_logic = emperor_require_logic.replace(
+            '/*__select_callback__*/', selection_callback)
```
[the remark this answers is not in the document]
ElDeveloper: I guess technically there's a slim chance this could happen. In general, I think we'll want to make it possible for the `Emperor` class to be able to render the custom code automatically and not need to do these string manipulations directly. This should be part of improving that API.

empress/support_files/js/empress.js
ElDeveloper: Good call!

ElDeveloper:
Thanks @fedarko, all changes should be addressed.

wasade:
this looks so cool....

kwcantrell:
Looks Good! 

I think this is outside of the scope of this PR but if the user had colored the empress tree using the sample coloring feature and then selects nodes in the emperor plot, the color by check box remains checked.
emperor
 
It would be nice to either reset the sample coloring panel or show the update button like you get when you change a coloring category as shown below. Otherwise if the user wants to recolor the tree again they would have to uncheck and then recheck the color by check box.
emperor2

empress/core.py
```diff
@@ -0,0 +1,335 @@
...
+        --------
+        empress.core.Empress.copy_support_files
+        """
+        main_template = self._get_template()
+
+        # _process_data does a lot of munging to the coordinates data and
+        # _to_dict puts the data into a dictionary-like object for consumption
+        data = self._to_dict()
+
+        # need to do something about low and high
```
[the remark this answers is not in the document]
fedarko: @ElDeveloper ... I still don't know what this comment means :P

fedarko:
Thanks @ElDeveloper! Agreed with Daniel and Kalen that this is awesome.

I tried this out locally and have played around with it -- I have a few remaining comments (if we could address them today that'd be awesome, if not then I guess we can get to them later).

1. I had to update my Emperor install to get this to work (which makes sense). However, it might be worth adding a check at some point that looks at the Emperor version and verifies that it's new enough to work with this? If an outdated version is in use (and we should keep in mind that 99.99% of people using these tools will have an "outdated" version of Emperor since the newest changes haven't been released yet), then we should raise an error or something explaining this. (We could even impose a minimum Emperor version in setup.py?)

2. Using a biplot as `--i-pcoa` works, but what happens is that the created version of Emperor doesn't seem to have a limit on the number of arrows shown (see below). If possible, changing the Emperor defaults on "number of arrows shown" to match the q2-Emperor defaults would be nice. If this wouldn't be ok, then we should probably expose a `--p-number-of-arrows` or something for Empress.

plot

3. Based on [this thread](https://github.com/biocore/emperor/pull/766#discussion_r439093981), I tried selecting samples from a parallel plot to view in Empress -- but it didn't seem like the samples were passed to Empress this time. Is this expected?

4. When I view the Emperor plot in 2-D and right-click drag to translate the display around, I can't seem to move the display up or down -- I can only move it left or right. Weirdly, when I add back Axis 3 (making the plot 3-D), the problem is still here -- but it sort of goes away when you tilt the plot along Axis 1. There seems to be some sort of "resistance" to moving the plot around that is somehow dependent upon the direction Axis 3 is pointing. My guess is that something is messing with the camera in Emperor. (This is a bizarre, unimportant problem, so it may be worth shelving and addressing later.)

5. (Not within the scope of this PR, but just for reference) I should also point out that the #152 bug is present here as well (which makes sense since the JS code also uses `_projectObservations()`), so once that is fixed this should highlight unique sections of the tree properly:

image

kwcantrell:
@ElDeveloper so you think we should still propagate the color up the tree if an internal node has children that do not belong to any samples (i.e. we will still get the current behavior for tips that are not in the feature table) and don't propagate the color when an internal node has children that belong to multiple samples? I think this option is better. However, we should distinguish uncolored tip that belong to multiple samples vs tips that aren't in the feature table. So, I think we should add all tips that belong to multiple categories to a "non-unique" category and assign them a color.

empress/support_files/js/selection-callback.js
```diff
@@ -0,0 +1,33 @@
...
+empress.colorSampleGroups(colorGroups);
+
+// 3 seconds before resetting
+setTimeout(function () {
+    empress.resetTree();
+    empress.drawTree();
+
+    samples.forEach(function (sample) {
+        sample.material.emissive.set(0x000000);
+    });
```
kwcantrell, line 30: I think it might be better to trigger the reset operation when the updates either the emperor plot or empress tree. That way the user explore the empress tree after the selected nodes in the emperor plot.

ElDeveloper: While I agree that this would be best, we can't really do that at the moment without linking additional empress and emperor callbacks (or am I missing something?).

ElDeveloper:
@fedarko:

1) I've added installation instructions to the README, which is I think the best we can do for the time being.

2) I've added the argument as is from q2-emperor.

3) Turns out this is a bug that I can't quite fix with empress. This needs to be fixed in Emperor: https://github.com/biocore/emperor/issues/771

4) Thanks for opening an issue in emperor https://github.com/biocore/emperor/issues/770

5) Should be addressed by Kalen's PR.

empress/support_files/js/selection-callback.js
kwcantrell: Okay I'll open up an issue.

ElDeveloper:
@kwcantrell, and @fedarko thanks so much you two! 🚀

PR #180, opened 2020-06-17
Circular Layout

kwcantrell:


kwcantrell:
Travis is failing because data_to_render has been modified. @ElDeveloper how easy would it be to update test_core.py? Did you get the values for DICT_A by creating an empress plot and then extracting the values?

fedarko:
Haven't gone over the PR super closely yet, but since it looks like this isn't in here -- I suggest adding test(s) to test_tree.py that verify that the computed coordinates for each node are positioned as expected (can be for a really simple tree, e.g. [as is done here](https://github.com/biocore/empress/blob/36df44c523132ca9fd39491cebf4c0b162683bec/tests/python/test_tree.py#L182-L191)).

Mainly, this would verify that the changes made w/r/t the scaling factor recently don't mess up the way nodes are represented -- each node's length in the circular layout space (just `sqrt(dx^2 + dy^2)` should be proportional to its branch length.

tests/test-circular-layout-computation.js
```diff
@@ -0,0 +1,99 @@
...
+
+            teardown: function() {
+            }
+        });
+
+        test('Test Circular Layout Arc Computation', function() {
+            // converting result to Set makes validation eaiser since
+            // getObjservationUnionForSamples uses a Set and then converts the
+            // Set to an array. The convertion does not keep order so converting
+            // the result back to a Set makes validation easier.
```
ElDeveloper, line 72: I think this comment is out of date.

```suggestion
            // converting result to Set makes validation eaiser since
            // getObjservationUnionForSamples uses a Set and then converts the
            // Set to an array. The convertion does not keep order so converting
            // the result back to a Set makes validation easier.
```

empress/support_files/js/vector-ops.js
```diff
@@ -76,10 +76,58 @@ define([], function () {
...
+     *
+     * @param {Number} x1
+     * @param {Number} y1
+     * @param {Number} x2
+     * @param {Number} y2
+     * @param {Number} amount - Thickness of the box to be drawn
+     *
+     * @return {Object} corners - Contains keys tL, tR, bL, bR
+     */
+    function computeBoxCorners(x1, y1, x2, y2, amount) {
```
ElDeveloper, line 95: Can you add some tests for this function?

empress/support_files/js/empress.js
```diff
@@ -413,24 +489,28 @@ define([
...
-            this.getX(this._treeData[node]) - amount,
-            this._treeData[node].lowestchildyr,
-        ];
-        var bR = [
-            this.getX(this._treeData[node]) + amount,
-            this._treeData[node].lowestchildyr,
-        ];
+        var corners = {
+            tL: [
+                this.getX(this._treeData[node]) - amount,
```
ElDeveloper, line 494: Shouldn't this be amount * 0.5? So that the amount is split evenly between the two sides or am I misunderstanding?

kwcantrell: My intention behind "amount" was that it specified a "level" of thinness instead of a specific amount of thinness. In this context, amount is the number the user selected in the "line-width" box. I played around with different modifiers and decided on this scheme. I guess the variable name is a bit miss leading, maybe "level" would be better?

ElDeveloper: I see, either changing the name or adding a comment would be good.

empress/support_files/js/vector-ops.js
kwcantrell: I added tests for vector-ops.js and fixed a bug in computeBoxCorners.

tests/test-circular-layout-computation.js
kwcantrell: This is a comment from an older test file. I copied one of the test files to use as a template and must have not erased all the code in it.

tests/test-vector-ops.js
```diff
@@ -0,0 +1,95 @@
+require(["jquery", "VectorOps"], function($, VectorOps) {
+    $(document).ready(function() {
+        // Setup test variables
+        // Note: This is ran for each test() so tests can modify bpArray without
+        // effecting other test
```
ElDeveloper, line 5:
```suggestion
```

ElDeveloper:
Thanks @kwcantrell and @fedarko!!!

Issue #181, opened 2020-06-17
Trim down CSS


fedarko:
As we discussed recently -- there is a lot of stuff in the ~900-line `empress.css` file that is currently unused (e.g. ~50 lines of code styling modal dialogs, none of which are used in the current version of Empress). It would be good to remove a lot of this stuff, both to simplify maintenance and to avoid nasty side effects (e.g. the CSS impacting Emperor's CSS, as is the case with a checkbox somewhere in Emperor's controls).

Related to #168.

fedarko:
Something within this issue to address -- would be good to remove the extra vertical scrollbar that is shown during Empress + Emperor plots. The reason for this scrollbar is *probably* attributable to something hiding within the CSS.

Issue #169, opened 2020-06-12
UI tweaks for node selection


fedarko:
- [x] It would be nice to allow users to select the root node of the tree. The fact that it isn't selectable might confuse some users. (Also, I think showing the user a table of sample presence info based on the root would help users see firsthand "oh, it's showing all of the samples in the table because this is the root of the tree, and all of the tips are descendants of the root.")

- [ ] If possible, the node selection menu should be positioned next to the selected node, even if it's a duplicate-ID node. I've noticed that sometimes I'll zoom into the tree, click on a node, and not see anything until I zoom out and see the table on another side of the screen:

![image](https://user-images.githubusercontent.com/4177727/84451256-6952bd00-ac07-11ea-9455-a613a408304a.png)

- [x] Regarding the bottom text in the selection box with `For tip nodes: ...` and `For internal nodes: ...` -- these shouldn't be shown at the same time; instead, the JS code should alter the selection box HTML accordingly in `SelectedNodeMenu.showLeafNode()` or `SelectedNodeMenu.showInternalNode()`.

- [x] Some of the UI fixes to the sample presence info table (e.g. making the sample metadata values actual header cells, rather than putting both values and counts in the same cell) are only done in `showLeafNode()` and not `showInternalNode()`. This is my bad -- ideally, the table-generating code from `showLeafNode()` should be encapsulated in another function that's called by both methods.

_Leaf_:
![image](https://user-images.githubusercontent.com/4177727/84451981-6c4ead00-ac09-11ea-8c64-33e619559f44.png)

_Internal node_:
![image](https://user-images.githubusercontent.com/4177727/84451970-61941800-ac09-11ea-86d4-b81c031e7005.png)

- [x] When I click on a node to select it, I have a hard time getting rid of the green circle that appeared above the node. Closing the node selection menu (or, I think equivalently, just clicking on the display but outside of the menu) should remove the green circle(s) since that node is no longer "selected".

I'd be happy to address some / all of these when the feature metadata stuff is worked out. Just recording these here for reference.

PR #183, opened 2020-06-18
Add feature metadata support, and some UI / code fixes

### LINKED ISSUE #129: Error when parsing Silva taxonomies
This is more an FYI as this was previously reported in QIIME2: https://forum.qiime2.org/t/qiime-taxa-filter-table-error/3947/10

If you try to use Silva based taxonomies you get this error:
```bash
There was an issue with viewing the artifact merged_seqs-taxonomy-silva.qza as QIIME 2 Metadata:

  CategoricalMetadataColumn does not support values with leading or trailing whitespace characters. Column 'Taxon' has the following value: 'D_0__Bacteria;D_1__Gemmatimonadetes;D_2__Gemmatimonadetes;D_3__Gemmatimonadales;D_4__Gemmatimonadaceae;D_5__Gemmatimonas;D_6__uncultured bacterium '
```

### LINKED ISSUE #130: color by feature metadata

### LINKED ISSUE #141: Add a marker to show where the root is for rooted trees

### LINKED ISSUE #168: Removing scrollbars on screen borders
One of the recent PRs added scrollbars (both vertical and horizontal) to the screen borders. Scrolling to zoom in/out on the tree also moves the scrollbars (and the legend box?) which is somewhat distracting. If it's possible, I think it would be good to restore the old behavior of fixing the legend in place (and not having scrollbars). I think this should be doable by fixing the CSS.

image

image

fedarko:
## Feature metadata support

The main thing in this PR is that Empress now makes use of feature metadata (closes #130). In the JS code, this is split up into _tip metadata_ and _internal node metadata_.

  - If the input feature metadata contains a "Taxonomy" column, this column will be split up into multiple columns by `;`s. This enables easy coloring of nodes by each taxonomy level (e.g. `Level 1` (Kingdom), `Level 2` (Phylum), etc.)

  - Feature metadata can be used to color the tree in two ways:
    - Method 1: **`Only use tip-level metadata?` is checked (default).** This just colors tips using their feature metadata, propagating things upward as is done for sample metadata coloring. Internal node metadata (if present) is ignored.

      image
      _Even though I assigned a few of the internal nodes fake "taxonomies" for testing purposes, this info isn't shown._

    - Method 2: **`Only use tip-level metadata?` is unchecked.** This colors all nodes in the feature metadata (tips and internal nodes alike) by their values, without doing any propagation.

      image
      _Here, we can see the internal nodes' feature metadata. All nodes with ID `0.915` have the Level 1 value `duplicate`, and the `root` has the Level 1 value `I'mTheRootLol`. Even though internal nodes can have duplicate names, they can still have feature metadata -- as this screenshot shows, this just means that internal nodes with the same name have the same feature metadata._

    - These methods are configurable using a checkbox (as @ElDeveloper suggested). Changing the value of this checkbox changes a "description" below the checkbox describing how feature metadata coloring works.

  - Feature metadata is also displayed (for any node that has it) in the node selection menu.
    image

    - In support of this, the root node is now selectable. I believe this is (in combination with the #160 changes) is sufficient to close #141, and this addresses 1 of the 5 tweaks in #169.
      
      image
      _Note that the sum across each of the rows in the sample presence info is 34 for the root node. This makes sense, because the moving pictures dataset (which I'm using here) has exactly 34 samples._

## UI / style improvements

  - Cleaned up the sample presence info table in the selection menu: this table (along with the feature metadata table added in this PR) are now rendered more conventionally, rather than having header and data information contained in the same cell but separated by `<p>`s.

  - Made the layout of the selection menu a bit more variable based on what the selected node is -- addresses one of the things in #169. Probably easiest to just show a before/after on the same node (before is from the prototype branch from a few weeks back)

    image
    image

  - Removed an old line in the CSS (see #181) that changed the cursor to `pointer` even on `text` and `number` inputs (and on `select`s). This makes it clearer to the user that you can type in these elements (since now the cursor is free to change to a normal [`I`-beam pointer](https://en.wikipedia.org/wiki/Cursor_(user_interface)#I-beam_pointer) where natural, e.g. on the `Search by node ID...` input)

  - Fixed the scrollbar issue because it was mildly inconveniencing me during testing and it was either that or throw my laptop out the window ;) (closes #168)

## Code maintenance

  - Abstracted a lot of code in `side-panel-handler.js` so that it could be reused between the sample and feature metadata coloring. Similarly, also abstracted some code in `select-node-menu.js` for reuse between leaf and internal node menus. (This addresses one of the things in #169.)
    - (hopefully this makes testing easier also)

  - Added SILVA annotation tests (closes #129)

  - Because we weren't actually doing it before (??? this may have been my fault), I added validation code that raises an error if tip IDs in the tree are not unique, or if tip IDs overlap with internal node IDs. Duplicate internal node IDs do not cause an error but do cause a warning.
    -  While I was at it, also added some quick validation code that checks to see if any of the node names in the tree start with `EmpressNode` and raises an error if so (this is another thing we have not been checking). I also updated the JS to use `.startsWith("EmpressNode")` rather than `.includes("EmpressNode")` to more faithfully match how these names are assigned.

  - Renamed `tools.name_internal_nodes()` to `tools.fill_missing_node_names()`, because this function 1) was not limited to internal nodes in the first place and 2) all internal nodes do not necessarily not have node names.

    - Also, I didn't realize it until this PR, but apparently that function added branch lengths of 0 for nodes without branch lengths... I didn't realize this until just now, which is why I added extra [branch length validation code](https://github.com/biocore/empress/blob/36df44c523132ca9fd39491cebf4c0b162683bec/empress/tree.py#L57-L75) back in March for the rectangular layout PR, which made this behavior of `name_internal_nodes()` unused (since any nodes with missing branch lengths would raise an error before the code would get to this point). In this PR I just chopped this behavior out of `name_internal_nodes()` (it was the easiest course of action since the validation code I added was tested, while this behavior of `name_internal_nodes()` was not), but we can relax the validation code later on down the line if desired.

## TLDR

... I think that's most of the major stuff here. (There's other stuff that wound up in this PR, for example some of the early work I did on circular layouts [#180] -- this is because I started working on feature metadata using the same branch as for circular layout work.)

I know this comment wound up being long, so I'm happy to meet to walk through this as needed.

Also, should mention that the JS code in this PR should be better tested -- getting things to this point took until now, so I would probably need at minimum another day to add in tests where I can.

Thanks!


empress/support_files/js/select-node-menu.js
```diff
@@ -150,16 +266,32 @@ define([], function () {
...
+            name,
+            this.empress._featureMetadataColumns,
+            this.empress._intMetadata,
+            this.fmHeader,
+            this.fmTable
+        );
+
+        // 2. Compute sample presence information for this node.
+        // (NOTE: this does not prevent "double-counting" samples, so the
+        // aggregation for duplicate names should be fixed.)
```
kwcantrell, line 293: Maybe we should only show sample presence information for just this node instead of the aggregation of all nodes with the same name and also display a warning message. (This would be outside the scope of this PR so after we decide how we want to address this, we would create an issue)

empress/taxonomy_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def split_taxonomy(feature_metadata):
+    """ Attempts to find a taxonomy column and split it into taxonomic levels.
+
+    If one of the columns in the feature metadata DataFrame (ignoring case)
+    is named "Taxon" or "Taxonomy", this will return a new DataFrame where
+    the column in question is removed and replaced with S + 1 new columns,
+    where S is the maximum number of semicolons present in any feature's
+    taxonomy.
```
kwcantrell, line 20: Are taxonomies always separated by a semicolon or is that something we are enforcing? if we are enforcing, do you think it would be better to allow the user to pass a deliminator as an input into Empress instead of forcing them to edit their metadata file?

empress/taxonomy_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    Raises
+    ------
+    TaxonomyError
+        If any of the following conditions are met:
+            1. Multiple possible "taxonomy columns" are present in the input
+               DataFrame: for example, both a "Taxon" and "Taxonomy" column are
+               present, or both a "taxon" and "Taxon" column are present.
+            2. A taxonomy column is present in the input DataFrame, and there
+               is already at least one column in the DataFrame that starts with
+               "Level".
```
kwcantrell, line 49: Does "Level" always refer to taxonomy?

empress/support_files/js/empress.js
```diff
@@ -663,25 +798,112 @@ define([
...
+                var fmVal = fmRow[cat];
+                if (_.has(uniqueValueToFeatures, fmVal)) {
+                    uniqueValueToFeatures[fmVal].push(tipID);
+                } else {
+                    uniqueValueToFeatures[fmVal] = [tipID];
+                }
+            });
+        });
+
+        var thisEmpress = this;
```
ElDeveloper, line 847: For consistency let's refer to this as sp or scope whatever is used in the codebase. Unless that's of course `thisEmpress`.

empress/support_files/js/empress.js
```diff
@@ -663,25 +798,112 @@ define([
...
+        // Do upwards propagation only if the coloring method is "tip"
+        // TODO / NOTE: _projectObservations() sets the .inSample property of
+        // features that are colored with metadata. This is "wrong," in the
+        // sense that samples don't really have anything to do with feature
+        // metadata coloring, but I don't *think* this will impact things
+        // because I think resetTree() should be called before any other
+        // coloring operations would be done. However, would be good to test
+        // things -- or at least to rename a lot of these coloring utilities
+        // to talk about "groups" rather than "samples", esp. since I think
+        // animation has the same problem...
```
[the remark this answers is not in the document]
ElDeveloper: Thanks for pointing this out.

empress/taxonomy_utils.py
```diff
@@ -0,0 +1,151 @@
+import warnings
```
ElDeveloper, line 1: Any chance you can add the copyright header for this and other files?

empress/taxonomy_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    # The entries in this tuple should only be lowercase
+    VALID_TAXONOMY_COLUMN_NAMES = ("taxon", "taxonomy")
+    lowercase_col_names = [str(c).lower() for c in feature_metadata.columns]
+
+    # See if there is a "taxonomy column", and do some related validation on
+    # column names
+    invalid_level_columns_present = False
+    tax_col_index = None
+    tax_col_name = None
+    for col, i in zip(lowercase_col_names, range(len(lowercase_col_names))):
```
ElDeveloper, line 60: Enumerate is the more pythonic way to do this:

```suggestion
    for i, col in enumerate(lowercase_col_names):
```

empress/core.py
```diff
@@ -107,9 +104,13 @@ def __init__(self, tree, table, sample_metadata,
...
 
         if self.ordination is not None:
+            # Note that tip-level metadata is the only "feature metadata" we
+            # send to Emperor, because internal nodes in the tree should not
+            # correspond to features in the table (and thus to arrows in a
+            # biplot).
             self._emperor = Emperor(
                 self.ordination, mapping_file=self.samples,
-                feature_mapping_file=self.features,
+                feature_mapping_file=self.tip_md,
```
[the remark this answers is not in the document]
fedarko: Yeah, `self.tip_md` and `self.int_md` are instantiated in `_validate_data()`, which is called before this block. I agree that it's sort of unclear, though -- do you think renaming `_validate_data()` to `_validate_and_match_data()` or something similar would make sense?

empress/core.py
```diff
@@ -233,6 +234,18 @@ def _to_dict(self):
         # Convert sample metadata to a JSON-esque format
         sample_data = self.samples.to_dict(orient='index')
 
+        # Convert feature metadata, similarly to how we handle sample metadata
+        if self.tip_md is not None or self.int_md is not None:
+            # We can just use self.tip_md.columns, since both the tip and
+            # internal node metadata should have identical columns
+            feature_metadata_columns = list(self.tip_md.columns)
```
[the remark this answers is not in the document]
fedarko: That case (only internal node metadata) is actually tested -- see [here](https://github.com/biocore/empress/pull/183/files#diff-677835dbe5d1a3456510983a9e74713aR372). It ends up working out because, even if `self.tip_md` is empty, it's still a DataFrame -- so it has the same columns as `self.int_md` (since these are just different subsets of the original feature metadata).

The code should probably be more explicit about why this works, though...

empress/taxonomy_utils.py
```diff
@@ -0,0 +1,151 @@
...
+            # https://stackoverflow.com/a/46018922/10730311
+            nonlocal max_sc_count
+            row_sc_count = fm_row.loc[tax_col_name].count(";")
+            max_sc_count = max(max_sc_count, row_sc_count)
+
+        # Figure out maximum number of semicolons in any feature (this works
+        # on a per-row basis even though axis="columns", because each call
+        # to validate_semicolon_count() gets a Series of columns which is
+        # technically a row I guess)
+        feature_metadata.apply(find_max_semicolon_count, axis="columns")
```
ElDeveloper, line 110: I think the following would be equivalent:

```suggestion
    feature_metadata[tax_col_name].str.count(';').max()
```

empress/taxonomy_utils.py
```diff
@@ -0,0 +1,151 @@
...
+        # Our use of result_type="expand" means that tax_levels will be a
+        # DataFrame with the same index as feature_metadata but with one column
+        # for each taxonomic level (in order -- Kingdom, Phylum, etc.)
+        tax_levels = feature_metadata.apply(
+            split_taxonomy_col, axis="columns", result_type="expand"
+        )
+        # Assign human-friendly column names: Level 1, Level 2, ...
+        tax_levels.columns = [
+            "Level {}".format(i) for i in range(1, len(tax_levels.columns) + 1)
+        ]
```
ElDeveloper, line 145: I think the following snippet would use pandas' API a bit better, let me know what you think.

```suggestion
        max_tags =  feature_metadata[tax_col_name].str.count(';').max()
        
        # add 2 for the trailing name and the non-zero index start
        levels = pd.DataFrame(index=feature_metadata.index, dtype='str',
                              columns=['level %d' % i for i in range(1, max_tags + 2)])
        
        for i, names in enumerate(df.Taxon.str.split(';')):
            levels.iloc[i, :len(names)] = names
        
        levels.fillna('Unspecified', inplace=True)

```

empress/tools.py
```diff
@@ -216,4 +236,32 @@ def match_inputs(
...
+    # features present as tips / internal nodes in the tree
+    tip_metadata = None
+    int_metadata = None
+    if feature_metadata is not None:
+        # Split up taxonomy column, if present in the feature metadata
+        ts_feature_metadata = taxonomy_utils.split_taxonomy(feature_metadata)
+        fm_ids = set(ts_feature_metadata.index)
+
+        # Subset tip metadata
+        fm_and_tip_features = fm_ids & tip_names
```
ElDeveloper, line 250: Being that fm_ids is an Index object you don't need to convert into a set to intersect a set, you can call the intersection method:

```
fm_ids.intersection(tip_names)
```

empress/tools.py
```diff
@@ -216,4 +236,32 @@ def match_inputs(
...
+        tip_metadata = ts_feature_metadata.loc[fm_and_tip_features]
+
+        # Subset internal node metadata
+        internal_node_names = set([
+            n.name for n in tree.non_tips(include_self=True)
+        ])
+        fm_and_int_features = fm_ids & internal_node_names
```
ElDeveloper, line 257: Same as above.

tests/python/test_taxonomy_utils.py
```diff
@@ -0,0 +1,304 @@
...
+                        "s__uniformis"
+                    ),
+                    "k__Bacteria; p__Firmicutes; c__Bacilli"
+                ],
+                "Confidence": [0.95, 0.8, 0, 1]
+            },
+            index=["f1", "f2", "f3", "f4"]
+        )
+
+    def check_basic_case_worked(self, split_fm):
```
ElDeveloper, line 41:
```suggestion
    def test_check_basic_case_worked(self, split_fm):
```

tests/python/test_taxonomy_utils.py
```diff
@@ -0,0 +1,304 @@
...
+        self.assertCountEqual(split_fm.columns, [
+            "Level 1", "Level 2", "Level 3", "Level 4", "Level 5", "Level 6",
+            "Level 7", "Confidence"
+        ])
+        # ...Next, check the index -- no features should've been dropped (that
+        # isn't even a thing that split_taxonomy() does, but let's be safe :P)
+        self.assertCountEqual(split_fm.index, ["f1", "f2", "f3", "f4"])
+
+        # Finally, let's check each row individually. This is kinda inelegant.
+        assert_series_equal(
```
ElDeveloper, line 58: Any reason not to use assert_frame_equal? It would avoid repeating the columns, etc.

tests/python/test_taxonomy_utils.py
```diff
@@ -0,0 +1,304 @@
...
+
+        # Do all the in-depth testing of split_fm in a utility func (we'll
+        # reuse this code a bit later)
+        self.check_basic_case_worked(split_fm)
+
+    def test_split_taxonomy_rows_with_no_semicolons(self):
+        funky_fm = self.feature_metadata.copy()
+        funky_fm.loc["f1", "Taxonomy"] = "birds aren't real"
+        funky_fm.loc["f2", "Taxonomy"] = "theyve been drones"
+        funky_fm.loc["f3", "Taxonomy"] = "all along :O"
```
ElDeveloper, line 167: 👍

tests/python/test_taxonomy_utils.py
```diff
@@ -0,0 +1,304 @@
...
+        split_fm = tax_utils.split_taxonomy(funky_fm)
+
+        # Notice that f4's taxonomy is still there -- so each feature will have
+        # 3 levels
+        self.assertCountEqual(split_fm.columns, [
+            "Level 1", "Level 2", "Level 3", "Confidence"
+        ])
+
+        # Check each row individually
+        assert_series_equal(
```
ElDeveloper, line 177: Same as above regarding assert_frame_equal. If there are other instances consider changing this.

empress/tree.py
```diff
@@ -54,8 +59,26 @@ def from_tree(cls, tree, use_lengths=True):
...
+        tip_names = []
+        internal_node_names = []
         max_branch_length = 0
         for n in tree.postorder(include_self=False):
+            if n.name is not None:
+                # NOTE: This should eventually be taken out when
+                # fill_missing_node_names() is refactored. However, for now,
+                # this makes sure that users can't accidentally break things by
+                # naming nodes identical to our default names for missing nodes
+                if n.name.startswith("EmpressNode"):
```
ElDeveloper, line 74: Is this something we could easily do with the balanced parenthesis structure instead of treenode?

ElDeveloper: Would just make for this to go faster.

empress/core.py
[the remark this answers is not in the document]
fedarko: (Update, I just went ahead and renamed the function. Marking as resolved, feel free to reopen this.)

empress/core.py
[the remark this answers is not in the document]
fedarko: Update, added comment explaining the situation.

empress/core.py
```diff
@@ -233,6 +234,18 @@ def _to_dict(self):
...
+            tip_md_json = self.tip_md.to_dict(orient='index')
+            int_md_json = self.int_md.to_dict(orient='index')
```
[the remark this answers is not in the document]
fedarko: This is a really good point -- the code was not clear about this. Added some clarifying comments and changed the `if` statement to just check `self.features`, which gives the same results and is a simpler way of phrasing this.

empress/plugin_setup.py
```diff
@@ -63,8 +63,10 @@
             'be automatically filtered out of the visualization.'
         ),
         'feature_metadata': (
-            'Feature metadata. Not currently used for anything, but will '
-            'be soon.'
+            'Feature metadata. Can be used to color nodes (tips and/or '
+            'internal nodes) in the tree. Features described in the metadata '
+            'that are not present in the tree will be automatically filtered '
+            'out of the visualization.'
```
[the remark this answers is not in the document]
fedarko: I think that's correct -- this is doing things the other way around, filtering out features described by the metadata but not present in the tree. This is analogous to how sample metadata is handled (here and in Emperor IIRC), where samples described by the metadata but not present in the table are automatically filtered out.

empress/support_files/js/empress.js
```diff
@@ -302,6 +350,62 @@ define([
                     coords.set(color, coords_index);
                     coords_index += 3;
                 }
+            } else if (this._currentLayout === "Circular") {
```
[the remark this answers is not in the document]
fedarko: IMO, it would probably make more sense to merge in #180 first. This PR includes some of the early circular layout code from when I was working on that, and I think it makes more sense to push all of the circular layout code in at once rather than to push in half of the circular layout now and the other half when #180 is in.

tests/python/test_taxonomy_utils.py
fedarko: Going to push back on this -- `check_basic_case_worked()` isn't actually a unit test, it's a utility function that's called by unit tests in this class. The current naming is deliberate, although I can make the comment here clearer if you'd like (I can see how it seems ambiguous as is).

tests/python/test_taxonomy_utils.py
fedarko: hahaha I'm glad someone caught this :P

tests/python/test_taxonomy_utils.py
fedarko: Nothing dramatic, the main motivation for writing things this way was that testing each feature one at a time seemed less overwhelming then making one really big DataFrame and passing that to `assert_frame_equal()`. Constructing that big DF would be kind of cumbersome, since as far as I can tell this would involve something like

```python
pd.DataFrame({
    "Level 1": ["k__Bacteria", "k__Bacteria", "k__Bacteria", ...],
    "Level 2": ["p__Bacteroidetes", "p__Proteobacteria", "p__Bacteroidetes", ...],
    ...
})
```
... where each feature's "ground truth" data is split across multiple lists, rather than being added in a single operation. I suspect that adding / removing features from the test is also easier with the series-testing method. However, this is totally a subjective preference, so I can rewrite these to use `assert_frame_equal()` if you'd like.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
ElDeveloper: 👍

tests/python/test_taxonomy_utils.py
ElDeveloper: Makes sense.

empress/core.py
```diff
@@ -233,6 +234,18 @@ def _to_dict(self):
         # Convert sample metadata to a JSON-esque format
         sample_data = self.samples.to_dict(orient='index')
 
+        # Convert feature metadata, similarly to how we handle sample metadata
+        if self.tip_md is not None or self.int_md is not None:
```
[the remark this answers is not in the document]
fedarko: The latter -- if one is not `None` the other will also not be `None`. @kwcantrell also noticed this problem -- I modified this section of the code to be more clear about what's going on.

empress/support_files/js/empress.js
fedarko: Did some digging and there isn't really a standard in the codebase for this:

```bash
$ grep -ri "= this;" empress/support_files/js/*
empress/support_files/js/animation-panel-handler.js:        var ap = this;
empress/support_files/js/animator.js:        var animator = this;
empress/support_files/js/canvas-events.js:        var canvasEvents = this;
empress/support_files/js/canvas-events.js:        var canvasEvents = this;
empress/support_files/js/empress.js:        var thisEmpress = this;
empress/support_files/js/select-node-menu.js:        var selectMenu = this;
empress/support_files/js/side-panel-handler.js:        var panel = this;
empress/support_files/js/side-panel-handler.js:        var sp = this;
empress/support_files/js/side-panel-handler.js:        var sp = this;
empress/support_files/js/side-panel-handler.js:        var sp = this;
empress/support_files/js/side-panel-handler.js:        var sp = this;
```

However, `var emp` is sometimes used to hold a reference to an `Empress` object (outside of `empress.js`, however) --

```bash
$ grep -ri "var emp" empress/support_files/js/*
empress/support_files/js/canvas-events.js:        var empress = this.empress;
empress/support_files/js/canvas-events.js:        var empress = this.empress;
empress/support_files/js/select-node-menu.js:        var emp = this.empress;
empress/support_files/js/select-node-menu.js:            var emp = this.empress;
empress/support_files/js/select-node-menu.js:        var emp = this.empress;
```

... so I updated this line accordingly to use `var emp = this;` (so it's slightly more consistent now).

ElDeveloper: Yikes, well we definitely need to fix that. https://github.com/biocore/empress/issues/185 Thanks for the investigation.

empress/support_files/js/empress.js
```diff
@@ -663,25 +798,112 @@ define([
...
+        if (method === "tip") {
+            fmObjs = [this._tipMetadata];
+        } else if (method === "all") {
+            fmObjs = [this._tipMetadata, this._intMetadata];
+        } else {
+            throw 'F. metadata coloring method "' + method + '" unrecognized.';
+        }
+
+        var uniqueValueToFeatures = {};
+        _.each(fmObjs, function (mObj) {
```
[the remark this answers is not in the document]
fedarko: Most of my testing for this branch has been just the moving pictures tree. Hm, that's interesting -- I'll do some performance testing tomorrow and see what I can find. (I doubt that this will be a big bottleneck because feature metadata, at least for 16S datasets, probably isn't going to cover *all* of the tips -- just those that are present in the table. But it's worth checking.)

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: Thanks! Opened #186 so we can keep track of this TODO.

empress/support_files/js/select-node-menu.js
fedarko: Gotcha. Opened #187 to keep track of this.

empress/taxonomy_utils.py
fedarko: It's pretty common convention (at least in my experience) -- definitely, for QIIME 2 users, I think the vast majority of datasets will use semicolons for this.

My vote is for now to just stick with semicolons, and we can maybe revisit in the future if requested. But I'll defer to Yoshiki / Erfan on what to do for this.

empress/tools.py
fedarko: Wow, apparently I've been doing extra work here for a really long time :D I'll come back and fix these tomorrow...

empress/taxonomy_utils.py
ElDeveloper: Semicolons are an ok assumption. We shouldn't enforce them. If someone is using a different delimiter the taxonomy string will in principle not be validly formatted for QIIME2. The beauty of a type and format system 🎆

tests/python/test_taxonomy_utils.py
ElDeveloper: Fair enough, this method is fine. Just thought I would check.

Just for reference you can also build a DataFrame in a more deterministic way:

```
pd.DataFrame(index=['first', 'second', 'third'],
    data=[[':)', ':|'], [':O', ':0'], [':D', 'D:']], columns=['one', 'two'])
```

This way everything is ordered the same way always, unlike with a dict 👍

empress/tree.py
fedarko: Maybe -- I'm not really familiar with the BP library, so I'm not sure how we'd do this through that.

empress/support_files/css/empress.css
```diff
@@ -277,53 +276,54 @@ label:hover .tooltip {
     border-color: transparent lightgray transparent transparent;
```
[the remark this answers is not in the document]
fedarko: Sounds good. I've changed everything from using `header-*` to using `menu-*`, so that things are consistent. (Some of the things which were previously named `hover-table-*` but weren't actually within a table, like `hover-table-notes`, I've changed to say `menu-box-notes`, for example.)

empress/support_files/css/empress.css
```diff
@@ -889,6 +883,26 @@ button.close:hover {
     background-color: gray;
 }
 
+#hover-table-warning {
```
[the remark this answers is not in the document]
fedarko: Wound up going with `menu-box-warning` instead, but I used your general suggestion :)

empress/support_files/js/empress.js
```diff
@@ -484,55 +585,89 @@ define([
...
+                        this.getX(this._treeData[parent]),
+                        this.getY(this._treeData[node]) - amount,
+                    ],
+                    bR: [
+                        this.getX(this._treeData[node]),
+                        this.getY(this._treeData[node]) - amount,
+                    ],
+                };
+                this._addTriangleCoords(coords, corners, color);
+            } else if (this._currentLayout === "Circular") {
```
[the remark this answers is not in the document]
fedarko: Ok, I'll keep an eye out during the merge

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: As an update, I tried this out on the HMP dataset you sent me back in May and didn't notice any performance problems when doing feature metadata coloring. I think it's safe to use underscore here for now -- if things start breaking down specifically w/r/t feature metadata, we can come back to this.

empress/support_files/js/vector-ops.js
```diff
@@ -76,10 +76,57 @@ define([], function () {
...
+        // https://stackoverflow.com/questions/2917175/return-multiple-values-in-javascript#comment2969172_2917186
+        return { tL: tL, tR: tR, bL: bL, bR: bR };
+    }
+
     return {
         getAngle: getAngle,
         magnitude: magnitude,
         rotate: rotate,
         translate: translate,
+        computeBoxCorners,
```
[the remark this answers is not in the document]
fedarko: Huh, I don't know why I wrote this like that. Yeah, this seems right to me.

empress/support_files/templates/empress-template.html
```diff
@@ -135,6 +142,10 @@ <h1 id="hover-table-node-id"></h1>
         var sPanel = new SidePanel(document.getElementById('side-panel'),
                                    empress, legend);
         sPanel.addSampleTab();
+        if ({{ feature_metadata_columns }}.length > 0) {
```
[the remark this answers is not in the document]
fedarko: Done, good catch!

empress/taxonomy_utils.py
fedarko: Done, added + updated it for this and all other python files. (doesn't seem like any of the html / js / css have copyright headers although we could add those there if desired, so i'm holding off on that for now.)

empress/taxonomy_utils.py
fedarko: We're making the simplifying assumption here that this is the case. In practice, I'm sure people could have feature metadata lying around where "Level" refers to something besides taxonomy ... which is why this error exists, to make sure that corner case won't break Empress. However, I doubt that this will be a big problem (most users will probably just have a f.m. file with "Taxon"/"Taxonomy"/etc. and "Confidence"/"Consensus"/etc.), and if it does wind up causing trouble for some users then we can handle this by using e.g. `TaxLevel` or something as a new prefix.

tests/python/test_taxonomy_utils.py
fedarko: Gotcha, thanks. Yeah in the interest of time I'll leave this as is, but if I have to write future tests of this ilk I'll use your method :)

empress/taxonomy_utils.py
fedarko: Wow, now I feel silly for writing all of that out :P Thanks for the heads up!

empress/taxonomy_utils.py
fedarko: This is a more concise way of doing things, but I ran into some problems trying to implement this code and make it pass the tests -- it doesn't strip out internal whitespace around levels (so the output has things like ` p__Firmicutes`), and after fixing that for some reason `names` is being set as `nan` for some rows during the enumerate loop. (Also, I think my original code is somewhat easier to read, although I am likely biased since I wrote it...)

Since the original code works, for the sake of time I'm going to suggest we move on for now? I can revisit this in the future if you'd like.

empress/taxonomy_utils.py
ElDeveloper: Outside fo the scope for this PR so no need.

empress/taxonomy_utils.py
ElDeveloper: Let's touch base on slack, I think it shouldn't be that far off to strip trailing and leading spaces. Trying `str.strip()` might do it.

ElDeveloper:
Thanks so much @fedarko! This looks great!

PR #189, opened 2020-06-19
fixed #152

### LINKED ISSUE #152: Figure out what's up with internal nodes being colored but children not colored during animation

kwcantrell:
This fixes #152. One thing to note is that the signature for _projectObservations has now changes. It takes one more parameter that, if set true, will insert a "non-unique" group into the returning object.


empress/support_files/js/util.js
```diff
@@ -1,12 +1,16 @@
 define(["underscore"], function (_) {
     /**
      * Remove all non unique keys
+     * Note: keys are refering to empress._treeData (i.e. postorder position of
+     *       tree nodes starting at 1)
      *
-     * @param {Object} keys An object containing multiple lists of keys
+     * @param {Object} keys Maps groups to sets of keys.
+     * @param {Set} removeAll Set of keys to remove regardless of whether or not
+     *              the are unique.
```
fedarko, line 9:
```suggestion
     *              they are unique.
```

empress/support_files/js/empress.js
```diff
@@ -793,16 +796,16 @@ define([
             obs[category] = this._namesToKeys(obs[category]);
         }
 
+        // assign internal nodes to approperiate category based on its children
+        obs = this._projectObservations(obs, true);
+
         // assign colors to categories
+        categories = util.naturalSort(Object.keys(obs));
```
fedarko, line 803: for some reason github won't let me comment on the actual line, but as long as we're sorting `categories` down here we shouldn't need to sort it in the first place earlier in this function (on the third line of `colorBySampleCat` looking at the latest commit in this PR). should be able to just replace that earlier line with `var categories = Object.keys(obs);`, and avoid the extra ~O(nlogn) cost :)

empress/support_files/js/empress.js
```diff
@@ -888,7 +891,7 @@ define([
         // to talk about "groups" rather than "samples", esp. since I think
         // animation has the same problem...
         if (method === "tip") {
-            obs = this._projectObservations(obs);
+            obs = this._projectObservations(obs, false);
```
fedarko, line 894: yeah again not sure why this has false for addNonUnique

tests/test-util.js
```diff
@@ -201,5 +201,59 @@ require(["jquery", "util"], function ($, util) {
...
+                "a" : new Set([2]),
+                "b" : new Set([6]),
+                "c" : new Set([7]),
+                "d" : new Set([10, 11, 12]),
+                "non-unique" : new Set([1,3,4,5])
+            }
+            var result = util.keepUniqueKeys(keys, new Set());
+            // qunit does not have a way to directly compare Set. So, first,
+            // each set has to be converted into an Array
+            var groups = ["a", "b", "c", "d", "non-unique"];
```
fedarko, line 222: Suggest adding something here that verifies that these are the _only_ keys aka groups in `keepUniqueKeys()`' output. I think you can probably use `util.naturalSort()` on `Object.keys(result)` and compare that with `groups` to verify that these two arrays are exactly equal.

tests/test-empress.js
```diff
@@ -0,0 +1,98 @@
...
+        // Note: This is ran for each test() so tests can modify bpArray without
+        // effecting other test
+        module('Empress' , {
+            setup: function() {
+                // tree comes from the following newick string
+                // ((f,(e,d)c)b,a)root;
+                var tree = new BPTree(
+                    new Uint8Array([1, 1, 1, 0, 1, 1, 0, 1, 0, 0, 0, 1, 0, 0]));
+                var treeData = {
+                    7: {
```
fedarko, line 13: (sorry if this is showing my balanced parentheses ignorance) but it isn't clear which tip IDs correspond to which node names in the newick tree. would you mind adding this information somehow? a comment underneath the newick string above would probably be easiest. i can infer that `2` and `3` are probably `e` and `d` (maybe not in that order), but not sure about the others.

(i'm drawing out this example on paper, hence the confusion)

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
-     * Finds the branches that are unique to each category in obs
+    /*
+     * Projects the groups in obs up the tree.
      *
-     * @param {Object} observations grouped by categories
+     * This function performs two distinct operations:
+     *      1) Removes the non-unique observations from each group in obs
+     *         (i.e. performs an 'exclusive or' between each group) and, if
+     *         addNonUnique is true, then places all non-unique observations
+     *         in a "non-unique" group.
```
fedarko, line 910:
```suggestion
     *         addNonUnique is true, then places all non-unique observations
     *         (that are tips in the tree) in a "non-unique" group.
```

fedarko: (wasn't clear to me at first that non-unique internal nodes are never included in the output `non-unique`, no matter what.)

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
+        // Note: if all children belong to "non-unique" should we project that
+        //       up the tree? Current behavior does not project it up the tree
+        //       because branches in this group may be elements of different
+        //       groups. For example "b1" may belong to skin and oral while
+        //       "b2" may belong to gut and skin.
+        for (var elem of result["non-unique"]) {
+            if (!tree.isleaf(tree.postorderselect(elem))) {
+                result["non-unique"].delete(elem);
+            }
+        }
```
fedarko, line 975: I don't know if this is intentional or not, but it looks like a tree where none of the leaves are `non-unique` (or, equivalently, all of the leaves are unique) still will cause `non-unique` to be included in the legend:

image

This is pretty confusing, or at least it was to me.

I'd suggest adding a check somewhere to see if `non-unique` is empty and, if so, remove it from the result anyway. (Or you could defer that to elsewhere in the code, I'll defer to you on where you think it'd be best to put this logic)

PS lmk if you want me to send you this data to help with testing

empress/support_files/js/empress.js
```diff
@@ -969,14 +1022,14 @@ define([
...
-        }
+        // // calculate total and relative percentages in each group
+        // var sampleCategies = Object.keys(sampleObs);
+        // for (i = 0; i < sampleCategies.length; i++) {
+        //     var category = sampleCategies[i];
+        //     var branchesInCategory = sampleObs[category].length;
+        //     keyInfo[category].tPercent = branchesInCategory / this._tree.size;
+        //     keyInfo[category].rPercent = branchesInCategory / relativeTreeSize;
+        // }
     };
```
fedarko, line 1033: I know it's kinda outside the scope of this PR, but we should really just delete this entire function (and the one line where it's called in `colorBySampleCat()`. Its only function right now is slowing down the visualization. (It'll remain in the version history, so we can add it back at some point in the future if desired.)

tests/index.html
```diff
@@ -66,13 +67,13 @@
          'Camera', 'Colorer', 'BiomTable', 'SummaryHelper', 'util', 'Empress',
          'testBPTree', 'testByteTree', 'testBIOMTable', 'testCamera',
          'testColorer', 'testSummaryHelper', 'testUtil',
-         'testCircularLayoutComputation', 'testVectorOps'],
+         'testCircularLayoutComputation', 'testVectorOps', 'testEmpress'],
```
fedarko, line 70: Super exciting that this is being added :D

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
+                var represented = false;
+                for (j = 0; j < categories.length; j++) {
+                    if (obs[categories[j]].has(i)) represented = true;
+                }
+                if (!represented) notRepresented.add(i);
+            }
+        }
 
         // assign internal nodes to appropriate category based on children
         // iterate using postorder
```
fedarko, line 945:
```suggestion
        // iterate using postorder
        // Note that, although we don't explicitly iterate over the
        // root (at index tree.size) in this loop, we iterate over all its
        // descendants; so in the event that all leaves are unique,
        // the root can still get assigned to a group.
```

(I had to double-check this to make sure it worked properly in this case, so this should save future work for other people with the same question :)

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
-                        obs
-    */
-    Empress.prototype._projectObservations = function (obs) {
-        var tree = this._tree;
-        var categories = Object.keys(obs);
+     *      2) Assigns internal nodes to a group if all of its children belong
+     *         to the same group.
+     *
+     * Note: All tips that are not passed into obs are considered to belong to
+     *       a "not-represented" group.
```
fedarko, line 916:
```suggestion
     *       a "not-represented" group, which will be omitted from the
             returned version of obs.
```

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
+     *       Also, if addNonUnique is true, then a "non-unique" object will
+     *       be inserted into the return object.
+     *
+     * @param {Object} obs Maps cateogires to a set of observations (i.e. tips)
```
fedarko, line 920:
```suggestion
     * @param {Object} obs Maps categories to a set of observations (i.e. tips)
```

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
+     * @param {Boolean} addNonUnique If true, and then insert the "non-unique"
+     *                               group into the returning object.
+     * @return {Object}
```
fedarko, line 923: should describe what the returned version of `obs` will be like (ok to not bother for now, this is minor)

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
-        return obs;
+        var result = util.keepUniqueKeys(obs, notRepresented);
+
+        // remove all "non-unique" branches that aren't tips
+        // Note: if all children belong to "non-unique" should we project that
+        //       up the tree? Current behavior does not project it up the tree
+        //       because branches in this group may be elements of different
+        //       groups. For example "b1" may belong to skin and oral while
+        //       "b2" may belong to gut and skin.
+        for (var elem of result["non-unique"]) {
```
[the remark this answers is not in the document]
kwcantrell: We should probably throw an error if "non-unique" is an actual group name (which in this context are values in the metadata files). The reason being, whatever group names are in result, are the name that will be displayed in the legend. That or we will have to rework the way legend.js functions.

empress/support_files/js/empress.js
```diff
@@ -897,34 +900,84 @@ define([
...
+            i,
+            j;
+
+        // find "non-represented" tips
+        // Note: the following uses postorder traversal
+        for (i = 1; i < tree.size; i++) {
+            if (tree.isleaf(tree.postorderselect(i))) {
+                var represented = false;
+                for (j = 0; j < categories.length; j++) {
+                    if (obs[categories[j]].has(i)) represented = true;
```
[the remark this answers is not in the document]
kwcantrell: Good point!

empress/support_files/js/empress.js
```diff
@@ -730,7 +730,10 @@ define([
         }
 
         // project to ancestors
-        observationsPerGroup = this._projectObservations(observationsPerGroup);
+        observationsPerGroup = this._projectObservations(
+            observationsPerGroup,
+            false
```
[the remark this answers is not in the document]
kwcantrell: If addNonUnique is set to true, _projectObservations() will add another group element to its return object which is the object used to color the tree. However, a lot of the coloring function created their coloring maps based on the groups that were passed into _projectObservations(). This resulted in multiple functions breaking because when they iterated through the groups in the return object from _projectObservations() to color the tree, they would usedthe group names to index into the color map and since "non-unique" was added after the color map was created, this resulted in multiple errors.

The two areas were addNonUnique is set to false are 1) feature based metadata coloring (which doesn't have non-unique branches) and 2) selecting nodes in emperor (which although can contain non-unique branches, it doesn't display a legend so, it wont be obvious to the user what the color representing non-unique branches means since the other colors will match emperor sample colors).

tests/test-empress.js
kwcantrell: woops that was an old comment. originally I used the letters in treeData but I forgot that empress uses the postorder position of a node as its key into _treeData so I changed the tree to just use the postorder position as its name to make things easier.

empress/support_files/js/empress.js
kwcantrell: thanks, for some reason I thought I explained that in the description.

empress/support_files/js/empress.js
kwcantrell: This is intentional (or at least was when I first designed empress). I wanted to show the user all of the possible values a node could take. However, I also think this is somewhat confusing. The reason this happens is because the data structure passed to _projectObservations is the same data structure being used to fill the legend. So, what ever groups are in that data structure are the things that get passed to the legend. This means that we also get the same behavior for any group that doesn't have any unique branches (for example if all observations in group a were also seen in group b then group a will still show up in the legend despite no branches being color for group a).

However, I think we should defer this issue as I believe it will be outside the scope of the PR

empress/support_files/js/empress.js
kwcantrell: I agree. This was an old (and broken) function that is not being used anywhere.

tests/index.html
kwcantrell: don't get too excited yet! I still need to add tests for all of the other functions in empress :)

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: Thanks for the explanation! That makes this a lot clearer.

I suspect we may want to develop things such that we don't need these special cases. We can discuss it at today's meeting, but I think one simple way to handle this is setting the (default) "non-unique" color to be the same as the (default) tree node color -- this would make it _a lot_ clearer that non-unique internal nodes and non-unique tips are colored the same way because they both aren't "unique" to a category. I suppose this would then mean that "non-represented" tips/nodes might need to be colored a different "technical" color, to quote @ElDeveloper -- maybe, by default, non-unique nodes could be colored gray and non-represented nodes could be colored black? We can hash out the details together.

(Even though there aren't non-unique tips in feature metadata coloring, the internal nodes are (basically always) still going to be non-unique, anyway -- consistency here is important, IMO. And for the Emperor-selection highlighting, we should probably handle things consistently (and maybe show a legend anyway). Of course, making that pull up a legend is probably outside this PR's scope, so we can open an issue for it :)

tests/index.html
fedarko: Fair -- getting the infrastructure set up is usually the hardest part, so I'm still excited anyway ;)

empress/support_files/js/util.js
```diff
@@ -1,12 +1,16 @@
 define(["underscore"], function (_) {
     /**
      * Remove all non unique keys
+     * Note: keys are refering to empress._treeData (i.e. postorder position of
```
fedarko, line 4:
```suggestion
     * Note: keys are referring to empress._treeData (i.e. postorder position of
```

tests/test-empress.js
fedarko: Awesome, thanks for changing this!

tests/test-empress.js
```diff
@@ -0,0 +1,102 @@
...
+            var groups = ["g1", "g2", "g3"];
+            for (var i = 0; i < groups.length; i++) {
+                var group = groups[i];
+                var expectedArray = Array.from(expectedResult[group]);
+                var resultArray = Array.from(result[group]);
+                deepEqual(resultArray, expectedArray);
+            }
+
+            var columns = Object.keys(result);
+            deepEqual(columns, groups);
```
fedarko, line 74: Not super important, but do you know if `deepEqual([1,2,3], [3,2,1])` passes or fails? I ask because I think the order of `Object.keys()` isn't necessarily constant, so I suspect these tests (or other tests based on these ones) might fail eventually due to the columns being out-of-order.

I would suggest calling `util.naturalSort()` on `columns` and `groups`, just so you know that the order should be the same.

tests/test-empress.js
```diff
@@ -0,0 +1,102 @@
...
+                "g3" : new Set([6]),
+            };
+            var result = this.empress._projectObservations(obs);
+
+            var groups = ["g1", "g2", "g3"];
+            for (var i = 0; i < groups.length; i++) {
+                var group = groups[i];
+                var expectedArray = Array.from(expectedResult[group]);
+                var resultArray = Array.from(result[group]);
+                deepEqual(resultArray, expectedArray);
```
fedarko, line 70: See comment below re: Object key ordering. I'd _expect_ `Array.from()` on a set to not give the same orderings consistently, meaning we'd have to sort the result and expected arrays first to verify they're the same with `deepEqual`, but ... it looks like this works? So IDK.

tests/test-util.js
```diff
@@ -201,5 +201,57 @@ require(["jquery", "util"], function ($, util) {
             deepEqual(split.numeric, ["1.0", "0.0", "-3.0"]);
             deepEqual(split.nonNumeric, ["0.0.0", "boaty"]);
         });
+
+        test("Test keepUniqueKeys without removeAll", function() {
+            var keys  = {
+                "a" : new Set([1,2,3,4]),
+                "b" : new Set([3,4,5,6]),
+                "c" : new Set([1,3,4,5,7]),
```
fedarko, line 209:
```suggestion
                "a" : new Set([1, 2, 3, 4]),
                "b" : new Set([3, 4, 5, 6]),
                "c" : new Set([1, 3, 4, 5, 7]),
```
(just changing the spacing to be consistent)

tests/test-empress.js
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            var columns = Object.keys(result);
+            deepEqual(columns, groups);
+        });
+    });
+});
```
fedarko, line 102: These tests look really fantastic. Having this stuff tested is going to make life so much easier!!!

This may be asking a bit much, but would it be possible to add in two extra corner-case tests to verify that these are correctly handled? These could both be on the same simple tree used for the other tests.

1. All of the tips are unique to the same group -> all internal nodes, including root (`7`), should be assigned to that group
2. All of the tips are not present in any group (i.e. `obs` is empty) -> no tips or internal nodes should be assigned to a group

IMO the first test would be especially useful to have written down so we don't have to worry about it in this context, since off-by-1 errors missing the root have come up here in the past :)

tests/test-empress.js
fedarko: ... Coming back to this, it may be worth setting up a simple util function that just takes two arrays, calls `util.naturalSort()` on them, and then verifies that the sorted arrays match using `deepEqual()`.

tests/test-util.js
```diff
@@ -201,5 +201,57 @@ require(["jquery", "util"], function ($, util) {
...
+                var resultArray = Array.from(result[group]);
+                deepEqual(resultArray, expectedArray);
+            }
+        });
+
+        test("Test keepUniqueKeys with removeAll", function() {
+            var keys  = {
+                "a" : new Set([1,2,3,4]),
+                "b" : new Set([3,4,5,6]),
+                "c" : new Set([1,3,4,5,7]),
```
fedarko, line 234:
```suggestion
                "a" : new Set([1, 2, 3, 4]),
                "b" : new Set([3, 4, 5, 6]),
                "c" : new Set([1, 3, 4, 5, 7]),
```

tests/test-util.js
```diff
@@ -201,5 +201,57 @@ require(["jquery", "util"], function ($, util) {
...
+                "a" : new Set([2]),
+                "b" : new Set([6]),
+                "c" : new Set([7]),
+                "d" : new Set([10, 11, 12]),
+            }
+            var result = util.keepUniqueKeys(keys, new Set());
+            // qunit does not have a way to directly compare Set. So, first,
+            // each set has to be converted into an Array
+            var groups = ["a", "b", "c", "d"];
+            for (var i = 0; i  < groups.length; i++) {
```
fedarko, line 222:
```suggestion
            for (var i = 0; i < groups.length; i++) {
```

tests/test-empress.js
```diff
@@ -66,15 +66,15 @@ require(['jquery', 'BPTree', 'Empress', "util"], function($, BPTree, Empress, ut
             for (var i = 0; i < groups.length; i++) {
                 var group = groups[i];
                 var expectedArray = Array.from(expectedResult[group]);
-                var resultArray = Array.from(result[group]);
+                var resultArray = util.naturalSort(Array.from(result[group]));
```
fedarko, line 69: I was thinking more along the lines of calling `util.naturalSort()` on both the expected array and result array... but I tested this and it still works, somehow ???, so whatever -- not a huge deal. JavaScript confuses me sometimes :P

PR #192, opened 2020-06-20
BUG: Problem updating color in circular layout

ElDeveloper:
Setting the color and switching between layouts would work just fine.
However, setting the color once in the circular layout would make the
browser tab crash. Seems to be related to to the line width not being parsed
correctly.


kwcantrell:
Interesting, I think their might have been some kind of "race" condition where the line width input box was invalid and was not set before the update function was called.  So, the line width value was undefined. I'll be able to confirm once I get #142.

ElDeveloper:
Thanks @kwcantrell!

PR #193, opened 2020-06-22
ENH: Update to newest Emperor API

ElDeveloper:
Depends on:

https://github.com/biocore/emperor/pull/773

Fixes integration with Parallel plots.


fedarko:
This looks good to me. I think I've found a few problems with the Emperor / Empress integration (namely: if you make a bunch of Emperor selections rapidly, then make a final selection, then that last selection gets cut short because I think the timeouts from the previous one activate early), but that is out of the scope of this PR so we can save it for later I guess.

ElDeveloper:
Thanks @fedarko!

PR #198, opened 2020-06-22
Update screenshot and QZV

fedarko:
Just a little something for the first release :)


kwcantrell:
Thanks @fedarko. We may also might want to update our README to reflect some of the changes regarding emperor/empress integrations.

PR #206, opened 2020-06-24
Use "var scope = this;" consistently throughout JS code

### LINKED ISSUE #185: Make consistent use of scoping variable name
When we assign `this` to another variable, it seems like we don't do this consistently. It would be good if we could consistently use `var scope = this` or `var self = this`.

fedarko:
Fixes #185.


empress/support_files/js/side-panel-handler.js
```diff
@@ -256,7 +256,7 @@ define(["underscore", "Colorer"], function (_, Colorer) {
             // </p>
             // The <p> breaks lines in a way that looks nice, and the
             // label/input just define the radio buttons as is normal in HTML:
-            // see https://www.w3schools.com/tags/att_input_type_radio.asp.
+            // see https://www.w3schools.com/tags/att_input_type_radio.ascope.
```
ElDeveloper, line 259: False positive 😄

ElDeveloper:
Thanks @fedarko

empress/support_files/js/side-panel-handler.js
fedarko: NOOOO I can't believe this, serves me right for not being thorough enough :sweat_smile:

Let me fix that...

fedarko:
Ok, false positive has been fixed -- should be good now :)

kwcantrell:
Thanks @fedarko, this looks good. I think we can go ahead and merge this.

ElDeveloper:
Thanks @kwcantrell and @fedarko!

PR #207, opened 2020-06-25
Fix overflowing bugs

### LINKED ISSUE #173: Add vertical overflow handling (e.g. a scrollbar) to the side panel
When you open up all of the sub-panels, the animation panel goes off the screen:

image

(There are scrollbars on the screen due to #168, but these only "scroll" the legend and not the side panel.)

As we add more functionality to Empress, this will become more of a problem.

### LINKED ISSUE #205: Put node name / ID in a scrollbar once it gets too long
This is mainly a problem for when features are exact sequences (e.g. sOTUs of lengths >= 150), since this can cause funky formatting of the node selection menu like:

image

It may be best to wrap the `ID: TACTATATACACAGT...` header in a `<div>` with a max width and `overflow-x: scroll;` so that the menu doesn't look off.

Happy to take this on, since it will probably end up tying into #181.

fedarko:
Fixes #173 and fixes #205.

GIF showing the now-slightly-fancier side panel (#173) --
scrollbar

GIF showing properly-overflowing long IDs (#205) --
idoverflow


kwcantrell:
Thanks @fedarko this looks good. I think we can merge this if @ElDeveloper agrees.

ElDeveloper:
Agree with @kwcantrell, looks great! Thanks @fedarko

Issue #167, opened 2020-06-11
Support toggling node circles on/off?


fedarko:
Not urgent or anything, but this could be a nice option for generating paper figures.

Issue #211, opened 2020-06-25
Set root marker to a non-circle shape, and allow recoloring based on root node's color


fedarko:
1. Justification for the "shape" thing is described in detail in the thread starting [here](https://github.com/biocore/empress/issues/143#issuecomment-649089045). Making this its own issue so I don't forget about that :)

2. Regarding the "recoloring" thing: this is due to a very minor issue that I doubt most users will run into. When the root node has a unique color (i.e. when coloring by feature metadata using both internal node / tip metadata, and the root has a unique value for the selected feature metadata field), this color is included in the legend as expected:

![image](https://user-images.githubusercontent.com/4177727/85769439-d74eb800-b6ce-11ea-9ef6-019d9847d7f7.png)

However, the only layout that actually draws anything besides the marker for the root node is the rectangular layout. In the circular and unrooted layout, no lines for the root node are drawn, and therefore the entry in the legend for the root node's unique color goes unused in the tree. This could confuse users:

![image](https://user-images.githubusercontent.com/4177727/85769603-ff3e1b80-b6ce-11ea-9728-2817ad79d35e.png)

A way around _this_ issue would be no longer coloring the root marker black, but instead allowing its color to change based on its node color (ensuring that the root node's unique color in the legend is actually reflected somewhere in the tree).

ElDeveloper:
> A way around this issue would be no longer coloring the root marker black, but instead allowing its color to change based on its node color (ensuring that the root node's unique color in the legend is actually reflected somewhere in the tree).

Interested in what others think but this sounds like a good option.

kwcantrell:
I agree, we should change the root marker to a special shape and allow its color to change based on its node color. This would allow the user to still visually see the root as well as resolve the legend issue.

Issue #138, opened 2020-03-03
Adding tooltips to the interface?


fedarko:
This is a small, low-priority thing I'd like to add in to Empress later on down the line -- recording it here for future reference.

We had some requests to add tooltips to the Qurro interface (screenshots [here](https://github.com/biocore/qurro/issues/225)), and this didn't actually represent much added work -- once things are set up with Bootstrap, etc. it's only ~3 lines of HTML per tooltip element.

This could complement some of the more complex functionality @kwcantrell has been working really well -- for example explaining what we mean by `Gradient`, `Trajectory`, etc. in the context of animations.

fedarko:
One thing worth mentioning: with #386, borders that match the background color are faster to add since nothing needs to be drawn in these cases. Would be nice to fit this into a tooltip for the barplot border options.

PR #212, opened 2020-06-25
Global tree properties

### LINKED ISSUE #172: dude where's my tree? (add recenter camera)
Playing around with the tutorial I realized that if a user zooms "too much", you can't bring back your tree, like here:
where

In case it helps this is the error in the console:
Screen Shot 2020-06-13 at 7 11 02 AM

Anyway, perhaps it will be worth adding a "recenter" button?

kwcantrell:
This PR resolves  #172 and #167. In addition, Empress now starts with the root node at center of screen and when user searches for a node, that node is place at the center of screen. Finally, the last point of #169 is address,namely, the green node goes away when the node menu disappears.

I wasn't sure how to best add the GUI elements so if you have suggestions on how to better implement them (mainly the center tree button) let me know.


empress/support_files/templates/side-panel.html
```diff
@@ -9,6 +9,25 @@
   </p>
 </div>
 
+<!-- Global Tree properties -->
+<button class="side-header collapsible">Tree Properties</button>
+<div class="side-content control hidden">
+  <p class="side-panel-notes">
+    If checked, this display node circles on all tree nodes that have a name
+    in the newick string that defined the tree.
```
fedarko, line 17:
```suggestion
    If checked, this draws circles at the positions of all tree nodes
    which had a specified name in the input Newick file.
```

empress/support_files/templates/side-panel.html
```diff
@@ -9,6 +9,25 @@
...
+  </p>
+  <p>
+    <label for="display-nodes-chk">Show Node Circles</label>
```
fedarko, line 20:
```suggestion
    <label for="display-nodes-chk">Show node circles?</label>
```

empress/support_files/templates/side-panel.html
```diff
@@ -9,6 +9,25 @@
...
+    <input id="display-nodes-chk" type="checkbox" checked="true">
+  </p>
+  <p class="side-panel-notes">
+    Make the root of the tree appear at the center of the screen.
```
fedarko, line 24:
```suggestion
    Centers the screen on the root node of the tree.
```

empress/support_files/templates/side-panel.html
```diff
@@ -9,6 +9,25 @@
...
+  </p>
+   <p>
+    <button id="center-tree-btn">Center Tree</button>
```
fedarko, line 27: The GUI is ok as is -- I think things are mostly self-explanatory -- but for my two cents, I'd suggest switching around the order of these two controls and their descriptions (putting the controls before their descriptions) and then applying the `indented` class to the description `<p>`s. Here's what this looks like on my end:

image

This 1) makes it clearer which descriptions map to which controls, and 2) makes these descriptions visually consistent with the "Only use tip-level metadata?" checkbox in the feature metadata section.

In general, I think this is the sort of situation where tooltips (#138) would be really useful (probably for later on, though... IMO there are more important things for us to address before taking the time to add those). Having tooltips set up would let us do things like just show a "Center Tree" button by itself, with a `(?)` or something next to it which would pop up this help text if people have further questions.

empress/support_files/js/drawer.js
```diff
@@ -306,10 +321,12 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
         c.uniformMatrix4fv(s.mvpMat, false, mvp);
 
         // draw tree nodes
```
fedarko, line 323:
```suggestion
        // draw tree node circles, if requested
```

empress/support_files/js/drawer.js
```diff
@@ -286,6 +289,18 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
         this.fillBufferData_(this.sProg_.nodeVertBuff, data);
     };
 
+    /**
+     * Display the tree nodes.
```
fedarko, line 293: As with the Empress object version of this function, this should specify that this will only take effect when `draw()` is called next.

empress/support_files/js/drawer.js
```diff
@@ -306,10 +321,12 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
-        c.uniform1i(s.isSingle, 1);
-        c.uniform1f(s.pointSize, 4.0);
-        this.bindBuffer(s.nodeVertBuff);
-        c.drawArrays(c.POINTS, 0, this.nodeSize);
+        if (this.showTreeNodes) {
+            c.uniform1i(s.isSingle, 1);
+            c.uniform1f(s.pointSize, 4.0);
+            this.bindBuffer(s.nodeVertBuff);
+            c.drawArrays(c.POINTS, 0, this.nodeSize);
+        }
```
fedarko, line 329: This isn't related to this review (feel free to ignore this question for later), but ... where in the code is the root node's marker circle drawn? I get that unchecking the checkbox hides the root node as well, but after going through the drawer and camera JS files (as well as `getNodeCoords()` in empress.js) I'm still not clear on how the root node is actually drawn. I think I'm probably missing something obvious, but figured I should ask you.

kwcantrell: I spent a while trying to figure this out but while creating tests for #142, I discovered that this is  actually a bug. Empress.prototype.getNodeCoords creates a node buffer (which is all initialize to zero) to store all the nodes in the tree. However, in getNodeCoords, we do not add nodes that start with EmpressNode which means parts of the buffer will still be 0. This causes the root node to be colored zero since its coordinates are (0,0) and black is (0, 0, 0).

This issue will be fixed once I submit my PR for #142

empress/support_files/js/select-node-menu.js
```diff
@@ -383,6 +383,7 @@ define(["underscore", "util"], function (_, util) {
         this.fmTable.classList.add("hidden");
         this.fmTable.innerHTML = "";
         this.drawer.loadSelectedNodeBuff([]);
+        this.empress.drawTree();
```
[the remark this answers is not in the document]
kwcantrell: Good point, we can either update the draw method to accept additional parameters that tell it what to draw or we can the parts of draw method responsible for drawing nodes to a new function.

empress/support_files/js/side-panel-handler.js
```diff
@@ -346,6 +350,25 @@ define(["underscore", "Colorer"], function (_, Colorer) {
...
+    SidePanel.prototype.addTreePropertiesTab = function () {
+        var scope = this;
+        this.treeNodesChk.onchange = function () {
+            scope.empress.setTreeNodeVisibility(scope.treeNodesChk.checked);
+            scope.empress.drawTree();
+        };
+        this.recenterBtn.onclick = function () {
+            scope.empress._drawer.centerCameraOn(0, 0);
+            scope.empress.drawTree();
+        };
```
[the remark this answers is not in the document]
kwcantrell: I agree, we should have drawTree come from a single point. I'll add a new method in empress and also update  the SidePanelHandler

empress/support_files/templates/side-panel.html
kwcantrell: Thanks, I'll adjust the order/description to be consistent with the others.

empress/support_files/js/drawer.js
```diff
@@ -380,12 +397,37 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
+        var center = gl.vec3.fromValues(
+            this.treeSpaceCenterX,
+            this.treeSpaceCenterY,
+            0
+        );
+
+        var centerMat = gl.mat4.create();
+        gl.mat4.fromTranslation(centerMat, center);
+
+        var nodePos = gl.vec3.fromValues(-1 * x, -1 * y, 0);
```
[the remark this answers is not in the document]
kwcantrell: Thanks, I didn't realize you can do this in js.

empress/support_files/js/select-node-menu.js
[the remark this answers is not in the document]
fedarko: That makes sense to me. Thanks for the detailed research -- saving the thick-lines optimization for later sounds good.

empress/support_files/js/side-panel-handler.js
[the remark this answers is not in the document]
fedarko: Sounds good, thanks!

empress/support_files/js/drawer.js
fedarko: That's insane -- that makes so much sense in retrospect, but it's also such an innocuous bug. Thanks for finding it!!! I'm gonna tag #211 here since the details of this may be relevant to that later on.

empress/support_files/js/empress.js
```diff
@@ -1106,5 +1111,53 @@ define([
         this.drawTree();
     };
 
+    /**
+     * Centers the viewing window at the average of the current layout.
+     */
+    Empress.prototype.centerLayoutAvgPoint = function() {
+        if (!(this._currentLayout in this.layoutAvgPoint)) {
```
ElDeveloper, line 1118: Nice, this helps defer the work until it is actually needed and it caches the results. Very useful!

PR #219, opened 2020-06-29
DOC: fix inconsistency in tree center button documentation

fedarko:
The behavior of this button was changed during comments in #212, but the corresponding help text wasn't updated. This fixes that.

(sorry -- I was going to comment this in #212 this morning, and then that PR got merged in :sweat_smile:)


empress/support_files/templates/side-panel.html
```diff
@@ -171,6 +171,6 @@
     <button id="center-tree-btn">Reset Camera</button>
   </p>
   <p class="side-panel-notes indented">
-    Centers the screen on the root node of the tree.
+    Centers the screen at the average coordinates of the tree's nodes.
```
ElDeveloper, line 174: Do you think this would make sense:

```suggestion
    Repositions the camera at the center of the tree.
```

Trying to think of something less technical but still accurate.

fedarko: Sure, sounds good to me.

fedarko: I think users might ask in the future for a "recenter camera" button that fits the screen to show the *entire* tree... but we can tackle that in the future.

kwcantrell:
@fedarko thanks for catching this!

PR #220, opened 2020-06-29
STY/BUG: Fix an issue with dropdown menus and rename IDs

### LINKED ISSUE #196: Distinguishing "name" vs. "ID" for nodes
This is a super super minor issue, but [Newick format](https://en.wikipedia.org/wiki/Newick_format) considers node names to just be "names", not "IDs". This convention is also adopted in [skbio.TreeNode](http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.html), in which `name` and `id` are distinct properties of nodes.

In Empress, we exclusively (as far as I can tell) use node `name`s, not `id`s. This also matches the way in which we validate the data (internal nodes are allowed to have non-unique names). (For reference, the skbio `id`s when I load the moving pictures tree in python seem to be mostly `None`; things like feature IDs on the tips are all stored in `name`.)

... So the point of this issue is that I think it would be good to modify the Empress UI (e.g. the search bar, node selection menu, etc.) to consistently say `name` instead of `ID` for nodes.

ElDeveloper:
Fixes an issue with drop-down menus pushing the label and checkbox in
the sidebar.

Long metadata columns would make the label and checkbox not be visible:

Screen Shot 2020-06-29 at 1 25 21 PM

I fixed this by setting the max width to a known value:

fixed

Also replaces "node id" references to be "node name" (note the screenshots above were capture before I put these fixes in).

Fixes #196

### DIFF
--- a/empress/support_files/css/empress.css
+++ b/empress/support_files/css/empress.css
@@ -58,6 +58,8 @@ select {
     -ms-appearance: none;
     appearance: none;
     padding-right: 30px;
+    /* ensures other elements are visible if there are long column names */
+    width: 160px;
 }
 
 select::-ms-expand {
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -198,14 +198,14 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
         var searchBtn = this.quickSearchBtn;
 
         var createClickEvent = function (e) {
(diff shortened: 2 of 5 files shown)


fedarko:
Thanks @ElDeveloper! I think this looks good to merge. (Good catch on the `tipID` thing in the feature metadata code, btw -- that's my bad.)

kwcantrell:
@ElDeveloper thanks this looks good!

PR #230, opened 2020-07-02
Empress tests

kwcantrell:
This PR addresses the empress tests in #142. Since test-empress.js is already a large file, I decided to upload this first to make reviewing a bit easy. This PR also fixes a bug that caused the root node to always get a black node. (I thought there was an issue for this but I couldn't find it).

Also getCoords, addTriangleCoords, addThickVerticalLineCoords, and thickenSameSampleLines do not have tests because there operations will be moved to the vertex shader in #190 and #214.

PR #231, opened 2020-07-02
163 remove tree data fix

kwcantrell:


kwcantrell:
We can still remove name from tree_data but after discussion with @ElDeveloper, this will be postponed until further discussion with @antgonza.

empress/support_files/js/empress.js
```diff
@@ -178,6 +178,11 @@ define([
      * Initializes WebGL and then draws the tree
      */
     Empress.prototype.initialize = function () {
+        // add default color/visible to all nodes
```
antgonza, line 181: This looks like duplicated code from where we initialize `this._numTips`, right? In other words, we might be able to add this to that for loop; look for `this._numTips = 0;`

tests/python/test_core.py
```diff
@@ -316,11 +316,7 @@ def test_filter_unobserved_features_from_phylogeny(self):
           'int_metadata': {},
           'feature_metadata_columns': [],
           'tree': [1, 1, 1, 1, 0, 1, 0, 0, 1, 0, 0, 1, 1, 0, 1, 0, 0, 0],
-          'tree_data': {1: {'color': [0.75, 0.75, 0.75],
-                            'name': 'a',
-                            'sampVal': 1,
-                            'single_samp': False,
-                            'visible': True,
+          'tree_data': {1: {'name': 'a',
```
ElDeveloper, line 319: @kwcantrell would it be hard to replace the dictionary of dictionaries for a list of lists? It seems like the key for tree data is an index eitherway. In addition, if we can change this:

`{'name': 'a', 'x2': 1, 'xc0': 2, 'xc1': 3, 'xr': 4, 'y2': 5, 'yc0': 6, 'yc1': 7, 'yr': 8}`

To this:

`['a', 1, 2, 3, 4, 5, 6, 7, 8]`

This should save python from writing all the repeated key data, and save JavaScript from parsing all of that. We would just need to load the data in a container once it's parsed in JavaScript. If this is straightforward to do, then let's do it in this PR otherwise we can defer to a later PR since this goes beyond the original scope.

kwcantrell: This is a good suggestion @ElDeveloper. I'll see if I can change treeData to a list of lists. However, I will need to modify a fair amount of the empress class so we should wait to merge this PR until #230 is merged.

ElDeveloper: Let's defer this until later then.

ElDeveloper:
@kwcantrell can you resolve the conflicts? Then it's ready to be merged.

PR #233, opened 2020-07-07
Store metadata and BIOM data in a compressed format in the HTML / JavaScript

### LINKED ISSUE #158: Storing JS BIOM data more efficiently
The JS BIOM table representation is structured as follows:

`{sample1: [feature1, feature2, feature3, ...], sample2: [feature3, feature4, ...], ...}`

...where `sample1`, `feature1`, etc. are the sample / feature IDs.

This approach inherently results in feature IDs being repeated multiple times, since they'll be listed for every sample they're in (a perfectly dense table with `N` samples will result in a JS representation where every feature ID is listed `N` times). If the input dataset isn't that huge and/or if the feature IDs are small strings this isn't a big deal, but when feature IDs are listed as sequences they represent (anywhere from a ~90 to ~500+ character string) this can inflate the `empress.html` (and the QZV) file sizes, because these long strings are listed multiple times.

A way to get around this problem would be finagling things so that each feature is assigned a unique numeric ID (starting at, say, 1, then going up to `number of features`), and the actual feature ID is stored only once somewhere in the data passed to Empress' JS code (maybe with the feature metadata, when we add that back in). This way, referencing a feature in the BIOM table won't require nearly as much space in the `empress.html` file.

Another way to get around this problem (this might be easier) would just be storing the transpose of the table --

`{feature1: [sample1, sample3, ...], feature2: [sample1, sample5, sample6. ...], ...}`

... since sample IDs are usually much shorter than feature IDs :) This is the approach Qurro currently takes.

Anyway, not urgent or anything, but this would probably help with the issue brought up in #74.

**Edit**: Per discussion with @wasade, the redbiom data model may be a useful example of how to implement this.

fedarko:
Closes #158.

The bulk of the changes here are in the new `compression_utils` python module (which gets compressed data ready to send to jinja2), and in the JavaScript BIOM table code (the compressed BIOM / sample metadata is formatted pretty differently from the previous way this data was stored, so I ended up rewriting most of the BIOM code -- the existing JS BIOM tests made this much less painful than I thought it might be :). I also added a decent amount of tests for the above modules, and made a few other changes throughout the codebase.

In terms of size compression, the early result from [here](https://github.com/biocore/empress/issues/158#issuecomment-649905429) still holds on the HMP dataset: the empress.html file goes down from 47MB to 26MB. The loading times seem... maybe slightly faster? but still mostly the same, at least for this dataset. Should probably benchmark this with an even larger dataset, though.


empress/compression_utils.py
```diff
@@ -0,0 +1,299 @@
+def compress_table(table):
```
ElDeveloper, line 1: Can you add a copyright header?

empress/compression_utils.py
```diff
@@ -0,0 +1,299 @@
...
+        f_ids_to_indices: dict
+            Inverse of f_ids: this maps feature IDs to their indices in f_ids,
+            analogous to s_ids_to_indices.
+        compressed_table: list
+            Two-dimensional list. The "outer list" is of length len(s_ids).
+            Each position i within this outer list holds an "inner list" of
+            arbitrary (but within the range [1, len(f_ids)]) length.
+            The i-th inner list contains the feature indices of the
+            features present (i.e. at any abundance > 0) within the
+            sample with index i. Each inner list is sorted in ascending order.
```
ElDeveloper, line 35: Is there any easy way to keep the abundance data around? his will likely be helpful for other issues down the road.

empress/support_files/js/biom-table.js
```diff
@@ -1,294 +1,447 @@
...
+        // Convert array of feature IDs to an array of indices
+        var fIndices = _.map(fIDs, function (fID) {
+            return scope._getFeatureIndexFromID(fID);
+        });
 
-        return result;
+        // Now, we can go through the table and find samples with matches
+        var containingSampleIDs = [];
+        _.each(this._tbl, function (presentFeatureIndices, sIdx) {
+            var sampleHasMatch = _.some(fIndices, function (fIdx) {
```
ElDeveloper, line 399: Can you move this anonymous function o outside of the _.each call?

fedarko: Done, thanks!

empress/compression_utils.py
```diff
@@ -0,0 +1,299 @@
...
+    num_shared_samples = len(metadata.index.intersection(sample_ids))
+    if num_shared_samples < len(sample_ids):
+        # Sanity check: metadata's samples should be a superset of the samples
+        # in s_ids_to_indices (empty samples were removed from the
+        # latter during table compression). If the metadata's samples are
+        # instead *missing* samples that are in s_ids_to_indices,
+        # something is seriously wrong.
+        raise ValueError(
+            "Metadata is missing sample IDs in s_ids_to_indices."
+        )
```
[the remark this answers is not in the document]
fedarko: Kind of -- at this point matching, etc. should already have been done, so we know that the data should be sane. This check is just verifying that the `s_ids_to_indices` mapping was constructed in `compress_table()` (or wherever we got it from) properly.

That said, this error should never be thrown in practice, but if the core.py code changes and breaks something in how this function is called then this error will indicate that something is up. (Esp. if that change happens, say, a few months - a year from now when we've all forgotten how this code works...) I can remove this check if you really want me to, but I would strongly prefer to keep it if possible -- it shouldn't be a major performance bottleneck, and the peace of mind from having each step of this use sanity checking is desirable IMO.

empress/compression_utils.py
```diff
@@ -0,0 +1,299 @@
...
+
+    if sorted(s_ids_to_indices.values()) != list(range(len(sample_ids))):
+        raise ValueError("Indices (values) of s_ids_to_indices are invalid.")
```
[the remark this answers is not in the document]
fedarko: Same point as with https://github.com/biocore/empress/pull/233#discussion_r451038304 -- this should never happen in practice because `compress_table()` is well-tested, but it's just an added layer of security. As with there, I don't *think* sorting a list of numbers will be a huge performance issue, so I would prefer to keep this if possible.

empress/compression_utils.py
fedarko: Done, thanks!

empress/compression_utils.py
fedarko: Not sure about "easy" ways. The simplest solution that comes to mind is storing abundance data in a separate structure with the same dimensions as `compressed_table` -- so if a sample's row in the table is

`[0, 5, 6, 100]`

indicating that that sample contains four features at those indices, then this sample's row in the abundance data structure will be the corresponding abundances for those features, e.g.

`[1, 2, 20, 3]`

or something like that. However, I think this will require a pretty significant amount of work to squeeze in, so my vote is in favor of delaying that to a future issue. (I'll cc @antgonza here in case he has any thoughts on this.)

empress/compression_utils.py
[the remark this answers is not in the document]
ElDeveloper: If there isn't a test case that would flag a situation like that, then we should definitely add that. But otherwise I think we would still be able to catch those errors, right?

My concern is not performance, more along the lines of redundancy and when is enough redundancy sufficient.

empress/compression_utils.py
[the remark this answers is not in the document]
ElDeveloper: In principle we aren't exposing compression_utils to 3rd parties so the functionality should be mostly within Empress. If this was to be exposed to 3rd parties then I can understand why you would want to add these checks. Are there hotspots in the code that you think aren't being tested enough that would lead to issues like this?

empress/compression_utils.py
ElDeveloper: That sounds like an easy enough solution, thanks! Let's not implement that but whenever needed this could be expanded as you describe.

empress/compression_utils.py
```diff
@@ -0,0 +1,299 @@
...
+    # Although the dict we just produced is compressed, we can go further
+    # without much extra effort. We implicitly sort the indices in ascending
+    # and then smoosh the dict into a list which can be accessed the exact
+    # same way (since [0] will now refer to the first element, [1] to the
+    # second, etc.)
+    metadata_vals = []
+    for i in range(len(sample_ids)):
+        metadata_vals.append(indices_to_metadata_vals[i])
+
+    return [str(c) for c in str_metadata.columns], metadata_vals
```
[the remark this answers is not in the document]
fedarko: This is a good point -- I just refactored the function to do this. It relies a lot more on pandas builtins so it's simpler now :)

empress/compression_utils.py
```diff
@@ -0,0 +1,299 @@
...
+        compressed_table.append(present_feature_indices)
+        # And update sample_ids and s_ids_to_indices accordingly.
+        sample_ids.append(sample_column.name)
+        s_ids_to_indices[sample_column.name] = sidx
+        sidx += 1
+
+        # We don't return anything -- this function's only purpose is to cause
+        # side effects within the scope of compress_table(). This is... kind
+        # of a cursed way of using df.apply() but it works, and it should be
+        # decently efficient since .apply() is faster than most iteration \._./
```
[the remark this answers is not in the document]
fedarko: I would prefer to keep this using `DataFrame.apply()`, since that is usually much faster than normal Python for loops when iterating over large DataFrames ([ref](https://engineering.upside.com/a-beginners-guide-to-optimizing-pandas-code-for-speed-c09ef2c6a4d6)). (We had a similar discussion a few weeks ago [about this code](https://github.com/biocore/empress/pull/183/files#r442490052), and apply proved to be much faster than manual iteration there.)

However, I agree that the function used for applying had too many side effects. I refactored this part of `compress_table()` so that the inner apply function only modifies `compressed_table` and nothing else -- this should make the code much easier to follow while still letting us get the speed benefits of using apply. LMK if this seems like a fair compromise to you.

fedarko:
@ElDeveloper I've addressed all of the comments (with the exception of the two discussions on the checks at the start of `compress_sample_metadata()`) and merged in the changes from upstream, so conflicts should be resolved. Please let me know if you'd like to discuss things further tomorrow so we can get this in for the release!

ElDeveloper:
Thanks @fedarko!

PR #237, opened 2020-07-08
Match feature table and ordination

ElDeveloper:
This would otherwise lead to situations where mismatching categories
would be displayed in the ordination and phylogeny. In principle users
should use the rarefied table that was used for generating the
ordination i.e. samples should have a 1:1 match.


fedarko:
@ElDeveloper The code changes look fine to me (with one conceptual question which I'll get to below). As I understand it, this PR replaces the `table.qza` in the repository with the rarefied table from the moving pictures tutorial?

I have some concerns about this. For one, I don't think the rarefied table should be used for the sample coloring functionality for Empress -- unless I'm missing something, there isn't a "need" to rarefy the data analogous to how rarefaction is needed for ordinations. Using the rarefied table for coloring the tree will likely result in a decently different coloring than a non-rarefied table, and IMO we should try to preserve all of the data as much as possible.

My contention is that it would be better to accept non-rarefied tables in this case (i.e. tables where the ordination might have a subset of the samples in the table, due to these samples being dropped out by rarefaction), and then alter the matching code to just subset the table's samples to those in the ordination (_without altering the other samples' abundances_). This should probably be accompanied by a warning about these samples being dropped due to not being in the ordination. (Of course, samples in the ordination but not the table should cause an error, as would ordinations where none of the samples are present in the table.)

However, if you disagree with this, please let me know -- we can chat. Regardless of what we decide on as the "best practice," I think we should document this somewhere so that users are well-informed. If there *isn't* a clear best practice, then we should document that.

(If we do decide that passing in rarefied tables is a good idea, then the part of the README with the "`table.qza` view | download" stuff should be updated accordingly -- at minimum the links should be updated to point to the `rarefied_table.qza` file, but ideally the filename should be changed to `rarefied_table.qza` also. Also IMO it would be best to include both the rarefied and non-rarefied table in the repository, so that the non-rarefied table can be used for the stand-alone visualization.)

rob-knight:
I can see arguments either way — rarefying will certainly affect the visualization, and in a stochastic way, but not rarefying will give too much visual weight to the samples with the largest # sequences, which may be undesirable depending on application. Can we make it a user-defined setting?

ElDeveloper:
That makes sense, @fedarko and I just chatted about it. We'll document this clearly, and note that it's a user-configurable choice.

empress/tools.py
```diff
@@ -146,16 +156,45 @@ def match_inputs(
...
+                raise DataMatchingError(
+                    "The feature table has more samples than the ordination. "
+                    "These are the problematic sample identifiers: %s. You can"
+                    " override this error by using the "
+                    "--p-filter-extra-samples flag" %
+                    (', '.join(sorted(extra)))
+                )
+            elif extra:
+                ff_table = ff_table[ord_ids]
+                ff_table = ff_table.loc[ff_table.sum(axis=1) > 0]
```
fedarko, line 184: This line looks like it's filtering empty features from the table? I think this check should be deferred until after matching is done -- I suspect there might be weird corner cases where, say, this would completely filter out features from the table, resulting in matching errors below (even if the table would otherwise share features with the tree or something).

Relevant to #244. Also relevant to the documentation of the consequences of using a "raw table" in the README.

If you'd like to defer addressing this until later today I think that'd be ok (should I try to smoosh in #244 before the release today?), but at minimum I'd request that you add a comment explaining what this line is doing (at least personally speaking, I always forget which pandas axis is which :)

empress/tools.py
```diff
@@ -146,16 +156,45 @@ def match_inputs(
...
+        else:
+            raise DataMatchingError(
+                "The ordination has more samples than the feature table"
```
fedarko, line 187:
```suggestion
                "The ordination has more samples than the feature table."
```

empress/tools.py
```diff
@@ -146,16 +156,45 @@ def match_inputs(
...
+            )
 
+        if ord_ids.issubset(table_ids):
+            extra = table_ids - ord_ids
+            if extra and not filter_extra_samples:
+                raise DataMatchingError(
+                    "The feature table has more samples than the ordination. "
+                    "These are the problematic sample identifiers: %s. You can"
+                    " override this error by using the "
+                    "--p-filter-extra-samples flag" %
```
fedarko, line 179:
```suggestion
                    "--p-filter-extra-samples flag." %
```

empress/tools.py
fedarko, line 184: Do you think it'd be worth adding a warning here with the number of samples that were dropped from the table, so this is consistent with how `--p-filter-missing-features` and `--p-ignore-missing-samples` work? Can understand if you'd prefer to not do that to avoid alarm fatigue.

README.md
```diff
@@ -58,6 +58,14 @@ This generates a visualization of a phylogenetic tree at
 
 ### Example 2: Using Empress to visualize a tree in tandem with an ordination
 
+**Note**: In tandem plots we recommend that users carefully consider which
+feature table they would like to visualize. These can either be a *filtered
+table* (rarefied or without low abundance features), or a *raw table*. If you
+select a *filtered table* the tree will include less data than in the *raw
+dataset*. If you select the *raw table*, you might find that some nodes in the
+tree won't be represented by any of the samples in the ordination (if the
+ordination was made using a *filtered table*).
```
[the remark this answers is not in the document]
ElDeveloper: Thanks this is really good!

empress/tools.py
```diff
@@ -146,16 +156,45 @@ def match_inputs(
...
+        # don't allow for disjoint datasets
+        if not (table_ids & ord_ids):
+            raise DataMatchingError(
+                "No samples in the feature table are present in the "
+                "ordination"
+            )
 
+        if ord_ids.issubset(table_ids):
+            extra = table_ids - ord_ids
+            if extra and not filter_extra_samples:
```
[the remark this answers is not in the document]
ElDeveloper: Good call!

empress/tools.py
ElDeveloper: That would be great if you can add a PR for #244.

ElDeveloper: In principle users should see the error first, so we gave them a warning already, let's defer.

README.md
```diff
@@ -58,6 +58,26 @@ This generates a visualization of a phylogenetic tree at
...
+
+There are some pros and cons for either of these choices. If you use a
+*filtered table*, then the Empress visualization will include less data than in
+the *raw dataset*: this will impact sample presence information, sample
+metadata coloring, and other parts of the visualization. If you select the *raw
+table*, you might find that some nodes in the tree won't be represented by any
+of the samples in the ordination (if the ordination was made using a *filtered
+table*). If you'd like to read more about this, there's some informal
+discussion in [pull request 237](https://github.com/biocore/empress/pull/237).
+
```
fedarko, line 80: something i just thought of (sorry for not bringing this up earlier), but since the table.qza file is changed in this PR the README should be updated to match that, right? i.e. the "view"/download links should be altered at least (preferably with a sentence or 2 of explaining which table from the MP tutorial this is)

empress/tools.py
fedarko: Understood. Thanks for the explanation!

fedarko: Ok, gotcha. I've been working on #244, will try to have it in shortly...

README.md
```diff
@@ -86,6 +89,7 @@ qiime empress plot \
     --m-sample-metadata-file docs/moving-pictures/sample_metadata.tsv \
     --m-feature-metadata-file docs/moving-pictures/taxonomy.qza \
     --o-visualization docs/moving-pictures/empress-tree-tandem.qzv
+    --p-filter-extra-samples
```
fedarko, line 92: GitHub isn't letting me suggest it on the correct line, but i think this is missing a backslash after the end of the above line

ElDeveloper:
Thanks @fedarko!

Issue #128, opened 2019-11-22
Create metadata tests (to address NaN issue)


kwcantrell:
to address #126

gibsramen:
To clarify - what should we be testing for here? If I'm understanding correctly, we would like to verify that if a sample/feature metadata file has NaN(s) that things work as normal?

ElDeveloper:
@gibsramen, yes that's what we would like to check.

ElDeveloper:
This is relevant because of the serialization to JSON.

fedarko:
I think probably the most painless way to test this (as far as I can tell) would be adding a test to `tests/python/test_core.py` which verifies that the `Empress._to_dict()` method's output looks ok. Could probably reuse a lot of the data / logic from `test_to_dict()`.

gibsramen:
Cool, yeah @fedarko that was my plan as well.

ElDeveloper:
One more thing to check would be calling `make_empress` in a test, and ensure this doesn't raise any errors when there are NaNs. This is because the Template calls `tojson`.

ElDeveloper:
Closed by @gibsramen in #238.

PR #228, opened 2020-07-01
fix #217 and add shifting to the tree

antgonza:

PR #238, opened 2020-07-08
Add test for NaNs in metadata

gibsramen:
Relevant to #128 

Added NaN to both sample & feature metadata and perform similar test to `to_dict`. Also calls `make_empress` per @ElDeveloper's suggestion.


gibsramen:
Didn't realize there was an update to `test_core.py`. Will address.

gibsramen:
So it looks like with #228 some of the handling of NaNs may have changed. When I wrote ac9d97f it was sufficient to set `dict_a_nan["sample_data"]["Sample2"]["Metadata4"]` to `np.nan`, whereas now the equivalent requires comparing to the *string* `nan` (i.e. `dict_a_nan["compressed_sample_metadata"][1][3] = str(np.nan)`).

Just want to confirm that this is desired behavior/handling.

tests/python/test_core.py
```diff
@@ -233,6 +233,37 @@ def test_to_dict_with_feature_metadata(self):
...
+
+        tree_data = obs['tree_data']
+        exp = dict_a_nan['tree_data']
+        self.assert_almost_equal_tree_data(tree_data, exp)
+
+        dict_a_nan.pop('tree_data')
+        obs.pop('tree_data')
+        self.assertEqual(obs, dict_a_nan)
+
+        viz.make_empress()
```
ElDeveloper, line 265: I suggest adding some minor checks that the template rendered correctly.

```suggestion
        res = viz.make_empress()
        
        self.assertTrue('empressRequire' in res)
        self.assertTrue('empress = new Empress' in res)
        self.assertTrue('emperor_require_logic' not in res)
```

tests/python/test_core.py
```diff
@@ -233,6 +233,37 @@ def test_to_dict_with_feature_metadata(self):
...
+        nan_feature_metadata.at["a", "fmdcol2"] = np.nan
+
+        viz = Empress(self.tree, self.table, nan_sample_metadata,
+                      nan_feature_metadata,
+                      filter_unobserved_features_from_phylogeny=False)
+        obs = viz._to_dict()
+        dict_a_nan = copy.deepcopy(DICT_A)
+        dict_a_nan["sample_data"]["Sample2"]["Metadata4"] = np.nan
+        dict_a_nan["tip_metadata"] = {
+            "a": {"fmdcol1": "asdf", "fmdcol2": np.nan}
```
ElDeveloper, line 250: I don't believe np.nan objects will serialize correctly once run through the JINJA template. I think these should be strings i.e. "nan".

ElDeveloper:
@gibsramen thanks for the PR, and welcome 🎉! You are correct we need to check for these to be strings, not np.nan objects. Otherwise JINJA's tojson is going to fail.

ElDeveloper:
Thanks @gibsramen, and congrats on your first contribution to Empress! 🌲 ✨

kwcantrell:
Thanks @gibsramen!

PR #239, opened 2020-07-08
notify user of non-unique

### LINKED ISSUE #161: Colors not updating (alert the user when there's no unique features)
Hi, 

First of all, this new tool looks awesome! But I am trying to visualise the tree based on different metadata colors and I only get coloured lines when the metadata variable has only one value, but no colouring when the metadata variable has more than one value. 

Screenshot 2020-05-24 at 00 36 43
Screenshot 2020-05-24 at 00 36 34

Would you please let me know how to fix this problem? 

Cheers, 
Pablo

kwcantrell:
Closes #161.
This PR adds a toast message if no groups were found with unique features when coloring the tree using sample metadata/animation. In addition, when coloring the tree using sample metadata, if a category is not represented on the tree then it will be removed from the legend.

### DIFF
--- a/empress/support_files/js/animator.js
+++ b/empress/support_files/js/animator.js
@@ -222,7 +222,7 @@ define(["Colorer", "util"], function (Colorer, util) {
         var obs = frame.obs;
 
         if (Object.keys(keyInfo).length === 0) {
-            util.toastMsg("No unique features found in current frame.");
+            util.toastMsg("No unique branches found for this frame");
         }
 
         // draw new legend
--- a/empress/support_files/js/animator.js
+++ b/empress/support_files/js/animator.js
@@ -207,9 +207,9 @@ define(["Colorer", "util"], function (Colorer, util) {
      * @private
      */
     Animator.prototype._collectFrame = function (frame) {
-            if (!(0 < frame < this.totalFrames)) throw 'Invalid Frame';
-            this.queuedFrames[frame] = this.retriveFrame(frame);
-            this.framesRdy[frame] = true;
+        if (frame < 0 || frame > this.totalFrames) throw "Invalid Frame";
+        this.queuedFrames[frame] = this.retriveFrame(frame);
+        this.framesRdy[frame] = true;
     };
 
     /**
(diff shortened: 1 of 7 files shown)


empress/support_files/js/side-panel-handler.js
```diff
@@ -206,6 +206,11 @@ define(["underscore", "Colorer"], function (_, Colorer) {
         var col = this.sColor.value;
         var hide = this.sHideChk.checked;
         var keyInfo = this.empress.colorBySampleCat(colBy, col);
+        if (keyInfo === null) {
+            util.toastMsg("No unique features were found.");
```
ElDeveloper, line 210:
```suggestion
            util.toastMsg("No unique branches found for this metadata category");
```

empress/support_files/js/empress.js
```diff
@@ -972,6 +979,13 @@ define([
         }
         var result = util.keepUniqueKeys(obs, notRepresented);
 
+        // remove all groups that do not contain unique features
+        categories = Object.keys(result);
+        for (i = 0; i < categories.length; i++) {
+            category = categories[i];
+            if (result[category].size === 0) delete result[category];
+        }
```
ElDeveloper, line 987: I think this and other similar loops can be replaced by underscore's pick. The following example keeps all the keys with a value that has a length greater than zero.

```
_.pick({'a': [], 'b': [1, 2, 3], 'c': [33]}, function(value, key){return value.length > 0})
```

empress/support_files/js/animator.js
```diff
@@ -220,6 +221,10 @@ define(["Colorer"], function (Colorer) {
         var keyInfo = frame.keyInfo;
         var obs = frame.obs;
 
+        if (Object.keys(keyInfo).length === 0) {
+            util.toastMsg("No unique features found in current frame.");
```
ElDeveloper, line 225:
```suggestion
            util.toastMsg("No unique branches found for this frame");
```

ElDeveloper:
Just one minor request regarding centering the message in Empress' viewport.

ElDeveloper:
Thanks @kwcantrell!

Issue #223, opened 2020-06-29
Using 0-indexing for the tree indices


fedarko:
Suggestion from @dhakim87 during code review today. Indices are currently 1-indexed to adhere with the original BP paper.

PR #240, opened 2020-07-08
Position the selection menu normally, even for nodes with duplicate names (when clicked on)

### LINKED ISSUE #169: UI tweaks for node selection

fedarko:
Closes #169. (More specifically, this addresses the second and currently-only-unchecked point in #169.)

When a node is clicked on, `CanvasEvents` already finds its key (an umambiguous identifier) in `empress._treeData`. Now, this key is used to center the camera on the node that was clicked on, as well as only highlight this node in green. Additionally, I believe the sample presence info should now be correct (since no iteration is done over all of the nodes with this name).

If a node is selected by searching and its name is ambiguous, then the old behavior is used (the camera is centered on the root, and all nodes with this name are highlighted). However, now there's some extra work done in the selection menu code that hides the sample presence info for this case.


empress/support_files/js/canvas-events.js
```diff
@@ -318,43 +335,71 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
...
+            openMenu(node, [nodeKey]);
         } else {
-            this.quickSearchBar.classList.add("invalid-search");
+            var keyList = this.empress._nameToKeys[nodeName];
+            if (keyList !== undefined) {
+                // At least one node with this name exists
+                if (keyList.length > 1) {
+                    // Multiple nodes have this name, so just place the menu at
+                    // the root of the tree
+                    node = this.empress._treeData[this.empress._tree.size - 1];
```
kwcantrell, line 392: Currently, empress uses a nodes postorder position in the tree starting at 1. So, I believe the root node should be this.empress._tree.size. 

On a separate note, this was brought up during the code review that using 1-based index can lead to confusion since 0-base index is typical. I used 1-based index because the paper the bp-tree was based on used 1-based index. Do you think it would be worthwhile to convert empress to use 0-based index. This would be outside the scope of this PR.

empress/support_files/js/select-node-menu.js
```diff
@@ -261,20 +264,30 @@ define(["underscore", "util"], function (_, util) {
             throw "showInternalNode(): nodeKeys is not set!";
         }
 
+        var name = this.empress._treeData[this.nodeKeys[0]].name;
+
         var isDup = false;
-        if (this.nodeKeys.length > 1) {
+        var isUnambiguous = true;
+        var keysOfNodesWithThisName = this.empress._nameToKeys[name];
```
kwcantrell, line 271: I though this.nodeKeys was this?

empress/support_files/js/canvas-events.js
fedarko: Thanks for catching this! Just altered this code to use the root's position for this stuff. Although now that I'm looking at it, what do you think about just positioning the menu/camera at the first (say, the first in the BP structure, or whatever) matching node? I think this might make more sense to the user, since the menu's "arrow" is pointing at a node that matched their search (and since they'll see the warning in the menu).

Addressing your other point: yeah, I think using 0-based indices would be a good idea in the future (although once we have testing in I guess it isn't a big deal). I opened #223 recently for this; would be nice for us to to get to that sometime.

empress/support_files/js/select-node-menu.js
fedarko: Sorry I was unclear -- long story short, the way this works now is that if a user clicks on a node, then nodeKeys will just have a single entry of that clicked-on-node's key (even if the node's name is ambiguous). If the user searched for a node with a duplicate name (e.g. `0.000` in the moving pictures data), then we have no way of knowing which node they meant, so nodeKeys will contain all of the `0.000`-named nodes.

So, `nodeKeys` and `keysOfNodesWithThisName` should always be the same *unless* this menu is for a node with a duplicate name that the user clicked on.

I reworked some of the code and added more thorough documentation, so hopefully this makes things less confusing.

fedarko:
@kwcantrell comments have been addressed, as described in my responses. Let me know what you think!

kwcantrell:
@fedarko thanks this looks good!

PR #243, opened 2020-07-08
Reveal selected samples from a selected node in a tandem plot

ElDeveloper:
Users can now reveal selected samples by clicking on a node in the tree. Selections are shown when a node is clicked, and hidden when the node's menu is removed from the screen.

node-select

This was paired-programmed with @kwcantrell 🥇


fedarko:
This is really exciting!

Haven't gone through the code yet but when I click on the root node I get the following browser error:

bruh

... and, after this, I can't select any nodes (the menus don't even come up; I just get the same error again, for both tips and internal nodes):

bruh2

Looks like Emperor is breaking somewhere? Not sure.

ElDeveloper:
@fedarko thanks for catching this, this is related to https://github.com/biocore/empress/pull/237 once that's merged this shouldn't be a problem.

empress/support_files/js/empress.js
```diff
@@ -1174,5 +1174,29 @@ define([
         this.drawTree();
     };
 
+    /**
+     * Set a callback when a the node menu is shown on screen
+     *
+     * The callback will receive a list of samples as the only argument. This
+     * is inteded to be used with emperor.
```
fedarko, line 1181:
```suggestion
     * is intended to be used with Emperor.
```

empress/support_files/js/empress.js
```diff
@@ -1174,5 +1174,29 @@ define([
...
+     *
+     * @param {Function} callback Callback to execute.
+     */
+    Empress.prototype.setOnNodeMenuVisibleCallback = function (callback) {
+        this._events.selectedNodeMenu.visibleCallback = callback;
+    };
+
+    /**
+     * Set a callback when a the node menu is removed from the screen
```
fedarko, line 1190:
```suggestion
     * Set a callback when the node menu is removed from the screen
```

empress/support_files/js/empress.js
```diff
@@ -1174,5 +1174,29 @@ define([
...
+     *
+     * The callback will receive a list of samples as the only argument. This
+     * is inteded to be used with emperor.
```
fedarko, line 1193:
```suggestion
     * is intended to be used with Emperor.
```

empress/support_files/js/empress.js
```diff
@@ -1174,5 +1174,29 @@ define([
...
+     * @param {Function} callback Callback to execute.
+     */
+    Empress.prototype.setOnNodeMenuVisibleCallback = function (callback) {
+        this._events.selectedNodeMenu.visibleCallback = callback;
+    };
+
+    /**
+     * Set a callback when a the node menu is removed from the screen
+     *
+     * The callback will receive a list of samples as the only argument. This
```
fedarko, line 1192:
```suggestion
     * The callback will receive an array of samples as the only argument. This
```

(since js technically doesn't have lists)

empress/support_files/js/empress.js
```diff
@@ -1174,5 +1174,29 @@ define([
         this.drawTree();
     };
 
+    /**
+     * Set a callback when a the node menu is shown on screen
+     *
+     * The callback will receive a list of samples as the only argument. This
```
fedarko, line 1180:
```suggestion
     * The callback will receive an array of samples as the only argument. This
```

empress/core.py
```diff
@@ -24,6 +24,8 @@
 TEMPLATES = os.path.join(SUPPORT_FILES, 'templates')
 SELECTION_CALLBACK_PATH = os.path.join(SUPPORT_FILES, 'js',
                                        'selection-callback.js')
+TIP_CLICK_CALLBACK_PATH = os.path.join(SUPPORT_FILES, 'js',
+                                       'tip-click-callback.js')
```
fedarko, line 28: This callback is called `tip-click-callback`, but it looks like this information is also sent over for internal nodes. Should this just be renamed to `node-click-callback` or something?

empress/support_files/js/select-node-menu.js
```diff
@@ -334,6 +350,9 @@ define(["underscore", "util"], function (_, util) {
             // retrive the sample data for the tips
             var samples = emp._biom.getSamplesByObservations(tips);
 
+            // used for the emperor callback
+            this._samplesInSelection = this._samplesInSelection.concat(samples);
```
fedarko, line 354: It's a little paranoid :), but I'd suggest resetting `this._samplesInSelection` to `[]` before this for loop. I know it's cleared in `clearSelectedNode()`, but if that operation is removed in the future (or if `clearSelectedNode()` is somehow not called before creating another menu), then this would silently break and include unrelated samples in the content sent to Emperor (also, IDK how easy it'll be to test this module, so the more careful the better for the time being IMO).

empress/support_files/js/tip-click-callback.js
```diff
@@ -0,0 +1,13 @@
...
+    ec.decViews.scatter.setEmissive(0x8c8c8f, samples);
+    ec.sceneViews[0].needsUpdate = true;
+});
+
+empress.setOnNodeMenuHiddenCallback(function (samples) {
+    samples = ec.decModels.models.scatter.getPlottableByIDs(samples);
+
+    ec.decViews.scatter.setEmissive(0x000000, samples);
+    ec.sceneViews[0].needsUpdate = true;
+});
```
fedarko, line 13: This code looks OK, but would suggest renaming the module to node-click-callback or something (see the comment in core.py).

tests/python/test_core.py
```diff
@@ -319,6 +319,30 @@ def test_no_intersection_between_tree_and_table(self):
...
+        self.assertNotEqual(id(self.sample_metadata), id(viz.samples))
+
+        self.assertIsNone(viz.features)
+
+        assert_ordination_results_equal(viz.ordination, self.pcoa)
+
+        # emperor is instantiated as needed but not yet setup
+        self.assertTrue(isinstance(viz._emperor, Emperor))
+
+        # ensure the rendering works correctly
```
fedarko, line 341:
```suggestion
        # ensure the callbacks were rendered
```

This avoids using the word "correctly" here, because this doesn't really check that the callbacks were inserted in the correct place or do the correct thing -- just that they were inserted at all. But this is minor :)

tests/python/test_core.py
```diff
@@ -319,6 +319,30 @@ def test_no_intersection_between_tree_and_table(self):
...
+        assert_frame_equal(self.sample_metadata, viz.samples)
+        self.assertNotEqual(id(self.sample_metadata), id(viz.samples))
+
+        self.assertIsNone(viz.features)
+
+        assert_ordination_results_equal(viz.ordination, self.pcoa)
+
+        # emperor is instantiated as needed but not yet setup
+        self.assertTrue(isinstance(viz._emperor, Emperor))
+
```
fedarko, line 340: It isn't really clear to me that most of this code is necessary -- it looks like it was copied in from `test_init_with_ordination()` (which in turn looks to has some stuff copied from `test_init()`). I'd suggest creating a utility function that takes in `viz` as a parameter and runs some or all of these checks, to avoid code reuse (and make adding future tests easier).

(I recognize that refactoring this module is by no means within the scope of this PR, so if you want to delay this until later and open an issue instead that's fine.)

empress/support_files/js/select-node-menu.js
ElDeveloper: Added!

tests/python/test_core.py
ElDeveloper: I think I'll defer this one until later each test looks slightly different.

ElDeveloper:
@fedarko should be ready for another pass, thanks!

empress/support_files/js/select-node-menu.js
fedarko: Sorry, I don't see this anywhere?

empress/core.py
fedarko: Sorry, it looks like this conversation was marked as resolved but I don't think this change was applied...

empress/support_files/js/tip-click-callback.js
fedarko: (sorry, this was marked as resolved but i don't think has been addressed... if i'm misunderstanding something just let me know)

empress/core.py
```diff
@@ -113,8 +115,8 @@ def __init__(self, tree, table, sample_metadata,
             self.base_url = 'support_files'
 
         self._validate_and_match_data(
-            ignore_missing_samples,
             filter_extra_samples,
+            ignore_missing_samples,
```
fedarko, line 119: this order disagrees with the parameter names in `validate_and_match`. I got around this by switching the param order in `tools.match_inputs()` as done in [this other branch's commit](https://github.com/fedarko/empress/commit/a7a6cb4aa69c1dbde05a61e3ae405f7cdf00af3e): https://github.com/biocore/empress/blob/aa40f2bea9bcaa812a7b11d1a1ca8f2f858478c6/empress/tools.py#L58-L67

PR #245, opened 2020-07-09
fix #241

### LINKED ISSUE #241: Tree fails to load when using rarefied table from moving pictures tutorial
If you replace `docs/moving-pictures/table.qza` for [this table from the tutorial](https://docs.qiime2.org/2020.6/data/tutorials/moving-pictures/core-metrics-results/rarefied_table.qza) everything works fine except the visualization fails to load with this error:

```
require-2.1.22.min.js:19 TypeError: Cannot read property 'x2' of undefined
    at Empress.getX (empress.js:217)
    at Empress.getCoords (empress.js:446)
    at Empress.drawTree (empress.js:210)
    at Empress.centerLayoutAvgPoint (empress.js:1174)
    at Empress.initialize (empress.js:203)
    at empress.html:312
    at Object.execCb (require-2.1.22.min.js:30)
    at ea.check (require-2.1.22.min.js:18)
    at ea.<anonymous> (require-2.1.22.min.js:23)
    at require-2.1.22.min.js:8
```

Just in case, [here's the broken artifact](https://view.qiime2.org/?src=https%3A%2F%2Fwww.dropbox.com%2Fs%2Fci5ncj8kl1kjktv%2Fbroken-empress-tree.qzv%3Fdl%3D0).

antgonza:
This fixes #241, the issue was when the last value had padded zeros. 

Note that empress alone works fine but when combining with Emperor we get:
Screen Shot 2020-07-08 at 7 19 42 PMk 
However, not fixing here as I think that falls out of the scope of this PR.


ElDeveloper:
@antgonza Would you mind explaining what the issue was? I am not able to understand from the code changes. Also, I pulled down your branch and was able to load a tandem plot with a rarefied and with an unrarefied table (for the moving pictures dataset). I don't know what might be causing the error you posted.

antgonza:
Sure. The issue/fix is in the new tests, basically when we have padded zeros in the last number, in the test -> `0, 0, 0, 0, 1, 1, 1`; the original code didn't deal with them and simply returned a 7. Now, it returns the padded zeros and the 7 -> `0, 0, 0, 0, 7`.

antgonza:
The Error is from Emperor, my guess: I'm using an old version. Thank you for checking!

ElDeveloper:
Thanks @antgonza!

PR #246, opened 2020-07-09
Position Empress and Emperor exactly next to each other and add a clear dividing line

fedarko:
This doesn't close any of the currently-open issues (if anything it *adds* slightly more CSS), but it makes tandem plots look a lot prettier. (It also gets rid of that scrollbar that kept showing up in tandem plots.)

Before: (note the gap in the middle of the screen, and the scrollbar on the right edge)
image

After:
image


empress/support_files/vendor/reset.css
```diff
@@ -0,0 +1,20 @@
+/* This is a heavily pared down version of reset.css, from
```
ElDeveloper, line 1: Any reason not to include this class in empress.css? Doesn't seem like much of a "module".

fedarko: Fair -- the reason I initially had this in a separate file was because I downloaded the (longer) original reset.css and then removed the stuff that we didn't need. But yeah, no reason to stick it in its own file -- I just moved it into empress.css and updated things accordingly.

fedarko:
Comment has been addressed, thanks @ElDeveloper! :princess: :penguin:

ElDeveloper:
Thanks @fedarko !

PR #247, opened 2020-07-10
Remove empty samples / features in the table after matching but before shearing

### LINKED ISSUE #244: Move empty sample / feature removal from the table to before tree shearing
This came up while chatting with @ElDeveloper about #237. Currently (unless I'm missing something), empty samples / features are only removed from the table in `compress_table()`, which is one of the last things done in Empress' python code. It would be preferable to move this work to be earlier in the code so that this is done before tree shearing -- this way, features in the tree and table (but not present in any samples in the table) would be removed from the tree if `--p-filter-unobserved-features-from-phylogeny` is passed.

fedarko:
Ensures that empty features in the table are removed from the tree if shearing is requested (could impact datasets where the user rarefied a table and this resulted in certain features having abundances of 0). Also gives a fancy error if these "empty" samples or features are present in the ordination (should probably never happen in practice, but if it does happen then the user'll know at least).

Closes #244.


empress/compression_utils.py
```diff
@@ -7,27 +7,121 @@
...
+        # Note: this has the side effect of, if the dtypes of the sample
+        # metadata are not "homogeneous", converting the dtypes all to object.
+        # See https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.transpose.html. # noqa: E501
+        # Since we'll eventually convert the sample and feature metadata's
+        # values to strings anyway, this shouldn't make a difference, although
+        # it would be ideal to do the string conversions before this step so
+        # there's no ambiguity here.
+        filtered_sample_metadata = filtered_table.align(
+            filtered_sample_metadata.T, join="inner", axis="columns"
+        )[1].T
```
ElDeveloper, line 88:
```suggestion
        filtered_sample_metadata = filtered_sample_metadata.loc[
            filtered_table.columns
        ]
```

tests/python/test_compression_utils.py
```diff
@@ -35,6 +49,19 @@ def setUp(self):
...
+        # perspective. We don't consider a sample with 0 for all of its
+        # metadata, or a metadata field with 0 for all samples, to be empty.)
+        self.sm_ef = pd.DataFrame(
+            {
+                "Metadata1": [0, 0, 0],
+                "Metadata2": [0, 0, 0],
+                "Metadata3": [1, 2, 3],
+                "Metadata4": ["abc", "def", "ghi"]
+            },
+            index=list(self.table_ef.columns)[:]
```
ElDeveloper, line 63:
```suggestion
            index=self.table_ef.columns.copy()
```

tests/python/test_compression_utils.py
```diff
@@ -123,24 +150,186 @@ def setUp(self):
...
+            proportion_explained=self.proportion_explained
+        )
+
+    # stdout mocking based on https://stackoverflow.com/a/46307456/10730311
+    # and https://docs.python.org/3/library/unittest.mock.html
+    @unittest.mock.patch("sys.stdout", new_callable=io.StringIO)
+    def test_remove_empty_1_empty_sample_and_feature(self, mock_stdout):
+        ft, fsm = remove_empty_samples_and_features(self.table, self.sm)
+        assert_frame_equal(ft, self.table_ef)
+        assert_frame_equal(fsm, self.sm_ef.astype("object"))
```
ElDeveloper, line 199:
```suggestion
        assert_frame_equal(fsm, self.sm_ef)
```

ElDeveloper:
Thanks @ElDeveloper

Issue #248, opened 2020-07-10
Remove feature metadata for nodes sheared from the tree


fedarko:
Sheared? Shorn? IDK.

Anyway, I think it's possible to specify feature metadata for nodes in the tree that will be removed if shearing is requested due to not being present in the table. However, the feature metadata will remain, and get passed up to the visualization.

I think this might cause problems with feature metadata coloring, and at the very least it'll take up extra space that doesn't need to be used. This could be fixed by adding a bit of code immediately after shearing the tree that filters the feature metadata accordingly (probably raising an error or warning if no feature metadata remains, i.e. all of it was for parts of the tree that are not described by the table).

This is probably a fairly rare corner-case since in practice (at least for 16S studies?) feature metadata will probably only describe stuff in the user's table.

fedarko:
This, along with #250, is blocking release of v0.2.0 (combo of this and #244 now working means that feature ID `8406...` in the MP dataset gets removed from the tree but not from the feature metadata, breaking feature metadata coloring in the visualization)

PR #251, opened 2020-07-11
BUG: Fix an issue with missing features

### LINKED ISSUE #250: Check that a tip exists in the BIOM table before triggering the visibleCallback to show present samples in Emperor
One of the funky issues @ElDeveloper and I ran into today that is delaying a release.

ElDeveloper:
Missing features in a tandem plot shouldn't be shown

Fixes #250

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,4 +55,5 @@ docs:
 		--m-sample-metadata-file docs/moving-pictures/sample_metadata.tsv \
 		--m-feature-metadata-file docs/moving-pictures/taxonomy.qza \
 		--o-visualization docs/moving-pictures/empress-tree-tandem.qzv \
+		--p-no-filter-unobserved-features-from-phylogeny \
 		--p-filter-extra-samples
--- a/Makefile
+++ b/Makefile
@@ -55,5 +55,4 @@ docs:
 		--m-sample-metadata-file docs/moving-pictures/sample_metadata.tsv \
 		--m-feature-metadata-file docs/moving-pictures/taxonomy.qza \
 		--o-visualization docs/moving-pictures/empress-tree-tandem.qzv \
-		--p-no-filter-unobserved-features-from-phylogeny \
 		--p-filter-extra-samples
--- a/empress/support_files/js/biom-table.js
+++ b/empress/support_files/js/biom-table.js
@@ -200,12 +200,27 @@ define(["underscore", "util"], function (_, util) {
     };
 
     /**
-     * Return a set of observation ids
+     * Return the feature ids in the BIOM table and inputed list
      *
-     * @return {Array} A set of feature identifiers.
+     * @param {Array[String]} other Array of strings to compute the
+     *                              intersection against.
+     * @return {Array[String]} feature ids in the BIOM table and the inputed
+     * list.
      */
-    BIOMTable.prototype.getObsIDs = function () {
-        return this._fIDs;
+    BIOMTable.prototype.getObsIDsIntersection = function(other) {
+        return _.intersection(this._fIDs, other);
+    };
+
+    /**
+     * Return the feature ids in the inputed list but not in the BIOM table.
+     *
+     * @param {Array[String]} other Array of strings to compute the
+     *                              set-difference against.
+     * @return {Array[String]} feature ids in the inputed list but not in the
+     * BIOM table.
+     */
+    BIOMTable.prototype.getObsIDsDifference = function(other) {
+        return _.difference(other, this._fIDs);
     };
 
     /**
--- a/empress/support_files/js/biom-table.js
+++ b/empress/support_files/js/biom-table.js
@@ -200,6 +200,15 @@ define(["underscore", "util"], function (_, util) {
     };
 
     /**
+     * Return a set of observation ids
+     *
+     * @return {Array} A set of feature identifiers.
+     */
+    BIOMTable.prototype.getObsIDs = function () {
+        return this._fIDs;
+    };
+
+    /**
      * Returns a list of observations (features) present in the input samples.
      *
      * @param {Array} samples Array of sample IDs
--- a/empress/support_files/js/biom-table.js
+++ b/empress/support_files/js/biom-table.js
@@ -200,24 +200,23 @@ define(["underscore", "util"], function (_, util) {
     };
 
     /**
-     * Return the feature ids in the BIOM table and inputed list
+     * Return the feature IDs shared by the BIOM table and input array
      *
      * @param {Array[String]} other Array of strings to compute the
      *                              intersection against.
-     * @return {Array[String]} feature ids in the BIOM table and the inputed
-     * list.
+     * @return {Array[String]} feature IDs shared by the BIOM table and other.
      */
     BIOMTable.prototype.getObsIDsIntersection = function (other) {
         return _.intersection(this._fIDs, other);
     };
 
     /**
-     * Return the feature ids in the inputed list but not in the BIOM table.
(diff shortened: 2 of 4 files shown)


empress/support_files/js/select-node-menu.js
```diff
@@ -402,6 +402,27 @@ define(["underscore", "util"], function (_, util) {
         }
     };
 
+    /**
+     * Warn the user when a tip or tips are not present in the feature table
+     * and ordination
+     *
+     * @return {Array} the features represented in the dataset.
+     */
+    SelectedNodeMenu.prototype._checkTips = function (tips) {
```
fedarko, line 411: Minor, but i'd suggest renaming this function so that it's clearer that it does stuff besides just validation. Maybe `_checkAndFilterTips()` or something? Up to you, though.

empress/support_files/js/select-node-menu.js
```diff
@@ -402,6 +402,27 @@ define(["underscore", "util"], function (_, util) {
...
+        // find the tips that can be used in the UI
+        var tipsForTable = _.intersection(this.empress._biom.getObsIDs(), tips);
```
fedarko, line 413: Actually, hm. I think a more sustainable solution to this is making a new function in the BIOM table JS that takes in a list of tip IDs and returns the intersection and difference of those with `this._fIDs`. There are two advantages of this:

1. No need to worry about the array copying thing I mentioned above, since the user doesn't interact with any of the BIOM private things directly
2. We can test that the intersection / difference functionality works without having to test the `select-node-menu` class. (I know that this code is really just using underscore functions so of course it'll work right now, but it's still more complex functionality that I'd prefer to have covered by some tests if possible. Also, if underscore's API would break due to a version update or something, having this tested would be good.)

empress/support_files/js/biom-table.js
```diff
@@ -199,6 +199,15 @@ define(["underscore", "util"], function (_, util) {
         return _.indexOf(arr, num, true) >= 0;
     };
 
+    /**
+     * Return a set of observation ids
+     *
+     * @return {Array} A set of feature identifiers.
+     */
+    BIOMTable.prototype.getObsIDs = function () {
+        return this._fIDs;
```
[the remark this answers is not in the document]
ElDeveloper: Very worth while, good catch!

empress/support_files/js/select-node-menu.js
ElDeveloper: Donezo!

ElDeveloper:
@fedarko, should be addressed now!

Makefile
```diff
@@ -55,4 +55,5 @@ docs:
 		--m-sample-metadata-file docs/moving-pictures/sample_metadata.tsv \
 		--m-feature-metadata-file docs/moving-pictures/taxonomy.qza \
 		--o-visualization docs/moving-pictures/empress-tree-tandem.qzv \
+		--p-no-filter-unobserved-features-from-phylogeny \
```
fedarko, line 58: Is this intentionally included, or was it just as a partial fix for #248?

tests/test-biom-table.js
```diff
@@ -277,6 +277,41 @@ require(['jquery','BiomTable'], function($, BiomTable) {
...
+            var obs;
+            var oids = ['o1', 'o2', 'o3', 'o4', 'o5', 'o6', 'o7', 'o8', 'o9',
+                        'o10'];
+            obs = this.biomTable.getObsIDsIntersection(oids);
+            assert.deepEqual(obs, oids);
+
+            obs = this.biomTable.getObsIDsIntersection(['o1']);
+            assert.deepEqual(obs, ['o1']);
+
+            obs = this.biomTable.getObsIDsIntersection(['oh no']);
```
fedarko, line 290: :)

empress/support_files/js/select-node-menu.js
```diff
@@ -366,8 +366,10 @@ define(["underscore", "util"], function (_, util) {
                 }
             }
 
-            // retrive the sample data for the tips
-            var samples = emp._biom.getSamplesByObservations(tips);
+            // retrive the sample data for the tips in the table
```
fedarko, line 369:
```suggestion
            // retrieve the sample data for the tips in the table
```

empress/support_files/js/select-node-menu.js
```diff
@@ -402,6 +404,31 @@ define(["underscore", "util"], function (_, util) {
         }
     };
 
+    /**
+     * Warn the user when a tip or tips are not present in the feature table
+     * and ordination
```
fedarko, line 409:
```suggestion
     * Given an array of tip names, warns the user about those that are not
     * present in the BIOM table (using a toast message) and returns just the
     * tip names in the array that are present as features in the table.
```

empress/support_files/js/select-node-menu.js
```diff
@@ -402,6 +404,31 @@ define(["underscore", "util"], function (_, util) {
...
+     *
+     * @return {Array} the features represented in the dataset.
```
fedarko, line 411:
```suggestion
     * @return {Array} the tip names that are also represented in the table.
```

empress/support_files/js/biom-table.js
```diff
@@ -199,6 +199,30 @@ define(["underscore", "util"], function (_, util) {
...
+        return _.intersection(this._fIDs, other);
+    };
+
+    /**
+     * Return the feature ids in the inputed list but not in the BIOM table.
+     *
+     * @param {Array[String]} other Array of strings to compute the
+     *                              set-difference against.
+     * @return {Array[String]} feature ids in the inputed list but not in the
+     * BIOM table.
```
fedarko, line 220:
```suggestion
     * @return {Array[String]} feature IDs in the input array but not in the
     *                         BIOM table.
```

empress/support_files/js/biom-table.js
```diff
@@ -199,6 +199,30 @@ define(["underscore", "util"], function (_, util) {
         return _.indexOf(arr, num, true) >= 0;
     };
 
+    /**
+     * Return the feature ids in the BIOM table and inputed list
```
fedarko, line 203:
```suggestion
     * Return the feature IDs shared by the BIOM table and input array
```

empress/support_files/js/biom-table.js
```diff
@@ -199,6 +199,30 @@ define(["underscore", "util"], function (_, util) {
...
+     *
+     * @param {Array[String]} other Array of strings to compute the
+     *                              intersection against.
+     * @return {Array[String]} feature ids in the BIOM table and the inputed
+     * list.
```
fedarko, line 208:
```suggestion
     * @return {Array[String]} feature IDs shared by the BIOM table and other.
```

empress/support_files/js/biom-table.js
```diff
@@ -199,6 +199,30 @@ define(["underscore", "util"], function (_, util) {
...
+     */
+    BIOMTable.prototype.getObsIDsIntersection = function (other) {
+        return _.intersection(this._fIDs, other);
+    };
+
+    /**
+     * Return the feature ids in the inputed list but not in the BIOM table.
```
fedarko, line 215:
```suggestion
     * Return the feature IDs in the input array but not in the BIOM table.
```

Makefile
ElDeveloper: Unintentional - good catch!

ElDeveloper, line 58:
```suggestion
```

ElDeveloper:
All addressed, thanks @fedarko!

fedarko:
Sweet, thanks @ElDeveloper!

PR #252, opened 2020-07-13
[WIP] fix-236

antgonza:


.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+
+name: CI-empress-test
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
kwcantrell, line 11: I am not too familiar with .yml files. Does this imply this will only work on PR from master branches? For example would I have to first merge my branch into my githubs master branch and then create a PR from my master or does the trigger whenever a pull request tries to merge into biocore/master?

antgonza: Note that this is a WIP so still trying to figure things out ... but AFAIK that's right! Only PRs and new pushes against master ... which is normally fine

PR #257, opened 2020-07-14
DEV: Add make-dev-page.py

### LINKED ISSUE #242: Streamline development process
Figure out something like emperor's build.py that generates a "test" visualization that doesn't need QIIME2 or much reprocessing.

ElDeveloper:
This creates a development page of Empress (optionally a tandem plot)
using the moving pictures tutorial data. To see changes from the
JavaScript and CSS files you can reload the generated page from the web
browser.

In the animated GIF below, I show how to see changes to the source by reloading the page (initially the log message is not visible in the console):

dev-cycle

Fixes #242

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,14 +33,19 @@ If you just want to run the Python or JavaScript tests, you can run
 ## Front-end development
 
 For convenience, a utility script `./tests/python/make-dev-page.py` is bundled
-in the test suite. If no arguments are set, the script will load the moving
-pictures dataset and create a page `development-page.html`. Using this page
-developers can modify CSS and JS files and simply reload the page on the
-browser to see their changes take effect.
+in the test suite. This script works best if you install the package in
+"editable mode" i.e. by running `pip install -e .` from the base directory.
+
+After installing the package, the script can be run without any arguments. This
+will load the moving pictures dataset and create a page
+`development-page.html`. Using this page developers can modify CSS and JS files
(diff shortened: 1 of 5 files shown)


kwcantrell:
@ElDeveloper thanks for adding this! I think this will help improve the development experience! Would it be possible to have the data set be a parameter? I have a couple datasets I like to test with.

ElDeveloper:
@kwcantrell, great suggestion that's very handy. I just added the CLI configuration to make this possible. If no arguments are set then the moving pictures dataset is loaded. I also added a couple of calls to .travis.yml so we check that this works as the code evolves.

CONTRIBUTING.md
```diff
@@ -30,6 +30,18 @@ make test
...
+For convenience, a utility script `./tests/python/make-dev-page.py` is bundled
+in the test suite. If no arguments are set, the script will load the moving
+pictures dataset and create a page `development-page.html`. Using this page
+developers can modify CSS and JS files and simply reload the page on the
+browser to see their changes take effect.
+
+Note, modifications to the Python code, or the template code require that you
+re-run the script. For a full list of options and arguments, you can run
+`./tests/python/make-dev-page.py --help`.
+
```
kwcantrell, line 44: Can you modify this file to tell users to use pip install -e .

This script wasn't working for me because I used pip install .

kwcantrell:
Thanks @ElDeveloper

ElDeveloper:
Thanks @kwcantrell!

PR #260, opened 2020-07-15
Removes a lot of unused CSS and improves CSS for integration with Emperor

### LINKED ISSUE #256: Prevent panning from selecting UI text in tandem plots
When you pan the tree such that the cursor moves across the divider in a tandem plot:

panning

...or even in a non-tandem plot, if the cursor moves onto the side panel:

panning2

... then a bunch of stuff in the UI gets highlighted by the cursor.

This isn't really a big problem, but fixing it would be nice. I'm guessing there's probably some obscure CSS thing that could be used to handle this -- maybe the plot being panned should disable text selection, or something similar.

ElDeveloper:
The input's CSS declarations for Empress weren't namespaced so the style affected Emperor.

This also centers the "loading animation" on the tree not on the screen - makes things look nicer in a tandem plot.

I've also fixed #256 by making the text non-selectable.


fedarko:
Thanks @ElDeveloper, this looks really good! I'm still going through the code changes, but one comment from trying this branch out: I think there are a few more changes needed before we can close #256 -- the "visible" label in the top-left Emperor corner still gets selected during panning, as does the "open menu" (`#show-ctrl`) button when the Empress side panel is closed.

image

The `#show-ctrl` thing should be fixable by just splitting
```css
#side-panel {
    position: absolute;
    top: 0px;
    right: 0px;
    width: 450px;
    max-height: 100%;
    overflow-y: auto;

    /* prevent user selection when panning around
     * https://stackoverflow.com/a/6900392
     */
    -webkit-user-select: none; /* Safari */
    -moz-user-select: none; /* Firefox */
    -ms-user-select: none; /* IE10+/Edge */
    user-select: none; /* Standard */
}
```

into, say,

```css
#side-panel {
    position: absolute;
    top: 0px;
    right: 0px;
    width: 450px;
    max-height: 100%;
    overflow-y: auto;
}

#side-panel, #show-ctrl {
    /* prevent user selection when panning around
     * https://stackoverflow.com/a/6900392
     */
    -webkit-user-select: none; /* Safari */
    -moz-user-select: none; /* Firefox */
    -ms-user-select: none; /* IE10+/Edge */
    user-select: none; /* Standard */
}
```
... but I'll leave the decision there up to you.

I'm not sure what you think about making the Emperor "visible" thing non-selectable completely (IDK if being able to select this text is useful in Emperor). Maybe giving the `<label>` an ID and then having Empress' CSS make that ID non-selectable would be a compromise solution...? We could just open an issue on Emperor to make this non-selectable and then close #256, I think that'd be fine since it's not an Empress issue per se.

ElDeveloper:
@fedarko thanks for catching that! Just added those changes and submitted a [PR to Emperor](https://github.com/biocore/emperor/pull/776). Seems to be working well:

non-selectable

fedarko:
@ElDeveloper This looks fine to me, let me know if you're done with changes and I can merge this tomorrow :)

PR #264, opened 2020-07-16
BLD: Require Emperor 1.0.1 or newer

### LINKED ISSUE #263: Check installed Emperor version to verify that it has the expected functionality
See #262 for some details. I don't think any users have run into this yet, but I think it's a matter of time. I guess if, a version of Emperor without certain callback functionality is installed, the python code should be able to detect this? And ideally raise an error explaining the situation to the user. (Otherwise, I think the user will just click around in tandem plots without anything happening.)

There are probably a few different ways of doing this:

- Find and interpret the version info of the currently-installed Emperor package. Raise an error if the version is too "early" (defining a way to compare version numbers might be tricky and imprecise, since these aren't necessarily just numbers)
- If there isn't anything to replace for any of [these lines of code](https://github.com/biocore/empress/blob/0ca08b7461058ff8202ec3217102571a0a4e3a72/empress/core.py#L391-L394), raise an error
- Define a more formal interface where Emperor can say what "callbacks" / functionality types it supports, and Empress can inspect this interface and determine whether to fail or not accordingly

ElDeveloper:
Should help us avoid silent issues

Fixes #263

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ with open('README.md') as f:
     long_description = f.read()
 
 base = ["numpy", "scipy", "pandas", "click",
-        "jinja2", "scikit-bio", "biom-format", "iow", "emperor"]
+        "jinja2", "scikit-bio", "biom-format", "iow", "emperor>=1.0.1"]
 test = ["pep8", "flake8", "nose"]
 all_deps = base + test
 


fedarko:
Thanks @ElDeveloper!

Issue #182, opened 2020-06-18
Remove "hide non sample features" checkbox


fedarko:
1. Before doing any coloring/etc.:

![image](https://user-images.githubusercontent.com/4177727/84979245-d96bb200-b0e3-11ea-9211-de003e7020d2.png)

2. After checking the checkbox:

![image](https://user-images.githubusercontent.com/4177727/84979207-bccf7a00-b0e3-11ea-9c54-a862db1a70d0.png)

3. After pressing "update" to color nodes by body-site:

![image](https://user-images.githubusercontent.com/4177727/84979218-c35df180-b0e3-11ea-954a-c200e51247ab.png)

Interestingly, at this point, checking / unchecking the box doesn't seem to do anything.

Screenshots taken on the moving pictures tree, visualized using the current master branch as of writing.

kwcantrell:
That is the intended behavior (well maybe not the second one, the check box is only supposed to be available after the tree is colored). The idea was to hide all branches that were not colored so that you can better see what is colored. The current behavior is to hide all uncolored tips and internal nodes whose children aren't colored. I guess the name is a bit miss leading. Perhaps were should either rename the checkbox or change the behavior so and just hide the tips that are not observed in a sample (I think the second option might be better).

kwcantrell:
will be resolved with #101

ElDeveloper:
Closed as part of #277

PR #267, opened 2020-07-17
Apply linting and stylechecking to JS test files

fedarko:
This PR doesn't add any functionality to Empress. All it does is include the test JS files in `make stylecheck` and `make jsstyle`. This'll be nice going forward, because it'll help keep these tests linted and looking consistent with the rest of the JS codebase.

(Figured it'd be easiest to add this in as its own PR, so that these changes can be reviewed in isolation.)

Most of the changes here are pretty minor stylistic things (e.g. single -> double quotes, spacing), but there are a handful of more substantial things that should probably be looked over. Some of these things I can think of:

1. The tree data in `test-empress.js` had duplicate `inSample` keys (each node had both a `false` and `true` version of this key set when initialized). I think that only the `true`s were being kept (also, if you remove all the `false`s nothing changes, but if you remove all the `true`s instead then one of the tests breaks), but another set of eyes on this code might be nice.

2. There are a few places in the tests where a negative value was used as the right term in a subtraction, e.g. [this line](https://github.com/biocore/empress/compare/master...fedarko:style-js-tests?expand=1#diff-79e9258925e6e54eb935a04856f3c4bbR33) which used `angle.sin - (-1/2)`. I ran prettier first, which (IIRC) changed this to `angle.sin - -1 / 2`, and then jshint complained about that, so I wound up just changing this to `angle.sin + 1 / 2`. But I can see how the first version of this might be preferable.


tests/test-circular-layout-computation.js
```diff
@@ -72,40 +74,40 @@ require(['jquery', 'BPTree', 'Empress'], function($, BPTree, Empress) {
...
+            equal(coords[(node - 1) * 10 + 5], 0); // end x position
+            equal(coords[(node - 1) * 10 + 6], -1); // end y position
 
             // For the arc for node 3 start at (2,0) and ends at (-2, 0)
             // check if arc for node 3 is correct
             ok(Math.abs(coords[30] - 2) < 1.0e-15); // start x arc position
             ok(Math.abs(coords[31 - 0]) < 1.0e-15); //start y arc position
-            ok(Math.abs(coords[175] - -2) < 1.0e-15) // end x arc position
-            ok(Math.abs(coords[176] - 0 < 1.0e-15)) // end y arc position
+            ok(Math.abs(coords[175] + 2) < 1.0e-15); // end x arc position
```
kwcantrell, line 109:
```suggestion
            ok(Math.abs(coords[175] - (-2)) < 1.0e-15); // end x arc position
```
This is not a big dead so if the linter complains do worry about it.

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-                        "visible": true
+                        color: [1.0, 1.0, 1.0],
+                        xr: 13,
+                        yr: 14,
+                        xc2: 27,
+                        yc2: 28,
+                        x2: 41,
+                        y2: 42,
+                        name: "root",
+                        inSample: true,
```
kwcantrell, line 50:
```suggestion
                        inSample: false,
```

This should actually be false. It looks like there was actually a mistake in the test.

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-                        "visible": true
+                        color: [1.0, 1.0, 1.0],
+                        xr: 11,
+                        yr: 12,
+                        xc2: 25,
+                        yc2: 26,
+                        x2: 39,
+                        y2: 40,
+                        name: "EmpressNode6",
+                        inSample: true,
```
kwcantrell, line 62:
```suggestion
                        inSample: false,
```

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-                        "visible": true
+                        color: [1.0, 1.0, 1.0],
+                        xr: 9,
+                        yr: 10,
+                        xc2: 23,
+                        yc2: 24,
+                        x2: 37,
+                        y2: 38,
+                        name: "internal",
+                        inSample: true,
```
kwcantrell, line 74:
```suggestion
                        inSample: false,
```

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-                        "visible": true
+                        color: [1.0, 1.0, 1.0],
+                        xr: 7,
+                        yr: 8,
+                        xc2: 21,
+                        yc2: 22,
+                        x2: 35,
+                        y2: 36,
+                        name: "internal",
+                        inSample: true,
```
kwcantrell, line 86:
```suggestion
                        inSample: false,
```

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-                        "visible": true
+                        color: [1.0, 1.0, 1.0],
+                        xr: 3,
+                        yr: 4,
+                        xc2: 17,
+                        yc2: 18,
+                        x2: 31,
+                        y2: 32,
+                        name: "2",
+                        inSample: true,
```
kwcantrell, line 98:
```suggestion
                        inSample: false,
```

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-                        "visible": true
+                        color: [1.0, 1.0, 1.0],
+                        xr: 5,
+                        yr: 6,
+                        xc2: 19,
+                        yc2: 20,
+                        x2: 33,
+                        y2: 34,
+                        name: "3",
+                        inSample: true,
```
kwcantrell, line 110:
```suggestion
                        inSample: false,
```

tests/test-empress.js
```diff
@@ -1,125 +1,127 @@
...
-
+                        color: [1.0, 1.0, 1.0],
+                        xr: 1,
+                        yr: 2,
+                        xc2: 15,
+                        yc2: 16,
+                        x2: 29,
+                        y2: 30,
+                        name: "1",
+                        inSample: true,
```
kwcantrell, line 122:
```suggestion
                        inSample: false,
```

tests/test-empress.js
```diff
@@ -240,116 +266,190 @@ require(["jquery", "BPTree", "Empress", "BiomTable", "util", "chroma"], function
...
-                3, 4, 0.75, 0.75, 0.75,
-                5, 6, 0.75, 0.75, 0.75,
-                7, 8, 0.75, 0.75, 0.75,
-                9, 10, 0.75, 0.75, 0.75,
-                13, 14, 0.75, 0.75, 0.75,
+                1,
+                2,
+                0.75,
+                0.75,
+                0.75,
```
kwcantrell, line 279: Would it be possible to form the array so that each line has five elements (i.e. the first line would be 1, 2, 0.75, 0.75, 0.75). The reason being that every five elements in this array represent one WebGl vertex. However, this is not a big issue so if the linter complains don't worry about it.

tests/test-empress.js
```diff
@@ -240,116 +266,190 @@ require(["jquery", "BPTree", "Empress", "BiomTable", "util", "chroma"], function
...
             deepEqual(empressRecCoords, rectCoords);
 
             var circCoords = new Float32Array([
-                15, 16, 0.75, 0.75, 0.75,
-                17, 18, 0.75, 0.75, 0.75,
-                19, 20, 0.75, 0.75, 0.75,
-                21, 22, 0.75, 0.75, 0.75,
-                23, 24, 0.75, 0.75, 0.75,
-                27, 28, 0.75, 0.75, 0.75,
+                15,
```
kwcantrell, line 311: Again would it be possible to make each line have 5 elements?.

tests/test-empress.js
```diff
@@ -240,116 +266,190 @@ require(["jquery", "BPTree", "Empress", "BiomTable", "util", "chroma"], function
...
-                31, 32, 0.75, 0.75, 0.75,
-                33, 34, 0.75, 0.75, 0.75,
-                35, 36, 0.75, 0.75, 0.75,
-                37, 38, 0.75, 0.75, 0.75,
-                41, 42, 0.75, 0.75, 0.75,
+                29,
+                30,
+                0.75,
+                0.75,
+                0.75,
```
kwcantrell, line 351: same as other previous 2

tests/test-empress.js
```diff
@@ -240,116 +266,190 @@ require(["jquery", "BPTree", "Empress", "BiomTable", "util", "chroma"], function
...
-         });
+        });
 
-         test("Test setNonSampleBranchVisibility", function() {
+        test("Test setNonSampleBranchVisibility", function () {
             // set node 6's inSample=false,  this means node 6's visible
             // should be the opposite of the input to
             // setNonSampleBranchVisibility
-            this.empress._treeData[6].inSample=false;
+            this.empress._treeData[6].inSample = false;
```
kwcantrell, line 387:
```suggestion
            this.empress._treeData[6].inSample = true;
```

tests/test-empress.js
```diff
@@ -240,116 +266,190 @@ require(["jquery", "BPTree", "Empress", "BiomTable", "util", "chroma"], function
...
 
             // pass in true
             // Note: this means node 6's visible should be false
             this.empress.setNonSampleBranchVisibility(true);
-            for(var i = 1; i <7; i++) {
-                var node = this.empress._treeData[i];
+            for (i = 1; i < 7; i++) {
+                node = this.empress._treeData[i];
                 if (i !== 6) {
                     equal(node.visible, true);
```
kwcantrell, line 405:
```suggestion
                    equal(node.visible, false);
```

tests/test-empress.js
```diff
@@ -240,116 +266,190 @@ require(["jquery", "BPTree", "Empress", "BiomTable", "util", "chroma"], function
...
                 } else {
                     equal(node.visible, false);
```
kwcantrell, line 407:
```suggestion
                    equal(node.visible, true);
```

tests/test-vector-ops.js
```diff
@@ -1,93 +1,83 @@
...
             ok(Math.abs(angle.cos - Math.sqrt(2) / 2 < 1.0e-15));
             ok(Math.abs(angle.sin - Math.sqrt(2) / 2 < 1.0e-15));
 
             // arbitrary point
-            angle = VectorOps.getAngle([-5*Math.sqrt(3) / 2, -5 / 2]);
-            ok(Math.abs(angle.cos - (-1*Math.sqrt(3))/2) < 1.0e-15);
-            ok(Math.abs(angle.sin - (-1/2)) < 1.0e-15);
+            angle = VectorOps.getAngle([(-5 * Math.sqrt(3)) / 2, -5 / 2]);
+            ok(Math.abs(angle.cos - (-1 * Math.sqrt(3)) / 2) < 1.0e-15);
+            ok(Math.abs(angle.sin + 1 / 2) < 1.0e-15);
```
kwcantrell, line 33:
```suggestion
            ok(Math.abs(angle.sin - (-(1 / 2)) < 1.0e-15);
```
Again not a huge issue so if the linter complains don't worry about it (I think the linter will complain seeing this change was the original line)

tests/test-vector-ops.js
```diff
@@ -1,93 +1,83 @@
...
-            }
-            var rPoint = VectorOps.rotate([1,1],angle);
-            ok(Math.abs(rPoint[0]- 0.3660254037844386) < 1.0e-15);
-            ok(Math.abs(rPoint[1] - -1.3660254037844386) < 1.0e-15);
+                cos: -1 / 2,
+                sin: (-1 * Math.sqrt(3)) / 2,
+            };
+            var rPoint = VectorOps.rotate([1, 1], angle);
+            ok(Math.abs(rPoint[0] - 0.3660254037844386) < 1.0e-15);
+            ok(Math.abs(rPoint[1] + 1.3660254037844386) < 1.0e-15);
```
kwcantrell, line 57:
```suggestion
            ok(Math.abs(rPoint[1] - (-1.3660254037844386)) < 1.0e-15);
```

Again not a big issue

tests/test-empress.js
fedarko: Definitely! This is a really good point, thanks for bringing it up. I'll add in [`// prettier-ignore`](https://prettier.io/docs/en/ignore.html#javascript) comments to override certain prettier directives like this.

tests/test-empress.js
fedarko: From what i can tell this change (and the other changes proposed in this test) would have a pretty significant impact on what the test is checking for. Would you mind adding suggestions that would update the comments in this test accordingly?

(I realize that we're planning to remove this functionality soon anyway, so not super urgent...)

fedarko:
@kwcantrell Thanks for the review! Per our discussion at this morning's meeting, I've removed the `setNonSampleBranchVisibility()` test in test-empress.js, as well as all references in this test file to the `.inSample` and `.visible` attributes of nodes in the tree data (since #182 will remove the functionality that uses these attributes).

The Travis build should pass now. Let me know if you see anything else :)

kwcantrell:
thanks @fedarko!

PR #270, opened 2020-07-18
Reveal a node on the tree when the user clicks on a biplot arrow

### LINKED ISSUE #203: Enable clicking on ordination arrows
To highlight the internal nodes or tips in the tree.

ElDeveloper:
If a biplot arrow is clicked, the node is revealed in the tree:

biplot

biplot

As a side effect, the samples in the emperor plot that have that feature are also highlighted.

Fixes #203


rob-knight:
This is so cool!

kwcantrell:
This is really cool @ElDeveloper! I am still going over the PR but I feel like this is different enough compared to the normal emperor/empress plot that I think it would be nice if you could also include an example visualization in the docs folder.

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,20 @@ define([
...
+    /**
+     * Show the node menu for a node name
+     *
+     * @param {String} nodeName The name of the node to show.
+     */
+    Empress.prototype.showNodeMenuForName = function (nodeName) {
+        if (!this._tree.containsNode(nodeName)) {
+            // TODO: should we show an error or something?
+            return;
+        }
```
kwcantrell, line 1222: I believe we are verifying this in the python code which would imply that all arrows would belong in the tree. If this is the case, then it may be helpful to simply throw an error in case this function is used outside of the scope of biplot interactions. However, if we are not verifying this in the python code then we should also display a toast message so that the user can be notified.

ElDeveloper: That's a good point, I don't believe this is being verified in the code at the moment. Would you prefer that we validated that or that we show the error message? I think validate would be better, thoughts @fedarko?

ElDeveloper:
@kwcantrell thanks for reviewing I just added biplot.qza and along the way noticed that there's an edge case to the biplot computation for this small dataset, so I will be addressing that in scikit-bio (see this issue https://github.com/biocore/scikit-bio/issues/1714). For now I included a small fix in the CLIs.

ElDeveloper:
@kwcantrell I addressed the feature metadata matching, should be good to review/merge.

empress/_plot.py
```diff
@@ -38,6 +38,9 @@ def plot(output_dir: str, tree: NewickFormat, feature_table: pd.DataFrame,
         # select the top N most important features based on the vector's
         # magnitude (coped from q2-emperor)
         feats = pcoa.features.copy()
+        # in cases where the the axes are all zero there might be all-NA
```
kwcantrell, line 41:
```suggestion
        # in cases where the axes are all zero there might be all-NA
```

kwcantrell:
Thanks @ElDeveloper!

ElDeveloper:
Thanks @kwcantrell

PR #274, opened 2020-07-22
Add tests for leaf/internal node sample presence

gibsramen:
Shifted some of the sample presence computation from select-node-menu.js to empress.js and added some tests.

### DIFF
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -1210,5 +1210,97 @@ define([
         this._events.selectedNodeMenu.hiddenCallback = callback;
     };
 
+    /**
+     * Calculate the number of samples in which a leaf node appears for each
+     * level of a given metadata field.
+     *
+     * @param {String} nodeName Name of the (leaf) node for which to calculate
+     *                          sample presence.
+     * @param {String} fields Metadata fields for which to calculate leaf
+     *                       sample presence.
+     * @return {Object} ctData returns a mapping of field categories to number of
+     *                  samples in which the given leaf node is present and
+     *                  sample field value equals category.
+     */
+    Empress.prototype.computeLeafSamplePresence = function (nodeName, fields) {
+        var ctData = {};
+
+        for (var f = 0; f < fields.length; f++) {
+            var field = fields[f];
+            var obs = this._biom.getObsCountsBy(field, nodeName);
+            var categories = _.keys(obs);
+            ctData[field] = {};
+            for (var c = 0; c < categories.length; c++) {
+                var cat = categories[c];
+                ctData[field][cat] = obs[cat];
+            }
+        }
+
+        return ctData;
+    };
+
+    /**
+     * Retrieve the tips in the subtree of a given node key.
+     *
+     * @param {Number} nodeKey Key value of node.
+     * @return {Array} tips Tips of the subtree.
+     */
+    Empress.prototype.findTips = function (nodeKey) {
+        // find first and last preorder positions of the subtree spanned
+        // by the current internal node
+        var t = this._tree;
+        var n = t.postorderselect(nodeKey);
+        var start = t.preorder(t.fchild(n));
+        var end = t.preorder(t.lchild(n));
+        while (!t.isleaf(t.preorderselect(end))) {
+            end = t.preorder(t.lchild(t.preorderselect(end)));
+        }
+
+        // find all tips within the subtree
+        var tips = [];
+        for (j = start; j <= end; j++) {
+            var node = t.preorderselect(j);
+            if (t.isleaf(node)) {
+                tips.push(t.name(node));
+            }
+        }
+
+        return tips;
+    };
+
+    /**
+     * Calculate the number of samples in which tips of an internal node
+     * appear for each level of metadata fields.
+     *
+     * @param {String} nodeName Name of the (leaf) node for which to calculate
+     *                          sample presence.
+     * @param {String} fields Metadata fields for which to calculate leaf
+     *                       sample presence.
+     * @param {Object} fieldsMap Object that maps fields to sample presence.
+     * @return {Object} fieldsMap Returns a mapping of field categories to number of
+     *                  samples in which tips of the internal node are present and
+     *                  sample field value equals category.
+     */
+    Empress.prototype.computeIntSamplePresence = function (
+        samples,
+        fields,
+        fieldsMap
+    ) {
+        // iterate over the samples and extract the field values
+        for (j = 0; j < fields.length; j++) {
+            field = fields[j];
+
+            // update fields mapping object
+            var result = this._biom.getSampleValuesCount(samples, field);
+            fieldValues = Object.keys(result);
+            for (k = 0; k < fieldValues.length; k++) {
+                fieldValue = fieldValues[k];
+                fieldsMap[field][fieldValue] += result[fieldValue];
+            }
+        }
+
+        return fieldsMap;
+    };
+
     return Empress;
 });
--- a/empress/support_files/js/select-node-menu.js
+++ b/empress/support_files/js/select-node-menu.js
@@ -241,7 +241,7 @@ define(["underscore", "util"], function (_, util) {
         );
 
         // 2. Add sample presence information for this tip
-        var ctData = {};
+        var ctData = this.empress.computeLeafSamplePresence(name, this.fields);
 
         // 2.1 The samples represented by this tip are sent to Emperor
         this._samplesInSelection = this.empress._biom.getSamplesByObservations(
@@ -345,28 +345,10 @@ define(["underscore", "util"], function (_, util) {
         if (isUnambiguous) {
             // this.nodeKeys has a length of 1
             var nodeKey = this.nodeKeys[0];
-
-            // find first and last preorder positions of the subtree spanned
-            // by the current internal node
-            var emp = this.empress;
-            var t = emp._tree;
-            var n = t.postorderselect(nodeKey);
-            var start = t.preorder(t.fchild(n));
-            var end = t.preorder(t.lchild(n));
-            while (!t.isleaf(t.preorderselect(end))) {
-                end = t.preorder(t.lchild(t.preorderselect(end)));
-            }
-
-            // find all tips within the subtree
-            var tips = [];
-            for (j = start; j <= end; j++) {
-                var node = t.preorderselect(j);
-                if (t.isleaf(node)) {
-                    tips.push(t.name(node));
-                }
-            }
+            var tips = this.empress.findTips(nodeKey);
 
             // retrieve the sample data for the tips in the table
+            var emp = this.empress;
             var samples = emp._biom.getSamplesByObservations(
                 this._checkAndFilterTips(tips)
             );
@@ -374,18 +356,12 @@ define(["underscore", "util"], function (_, util) {
             // used for the emperor callback
             this._samplesInSelection = this._samplesInSelection.concat(samples);
 
-            // iterate over the samples and extract the field values
-            for (j = 0; j < this.fields.length; j++) {
-                field = this.fields[j];
+            fieldsMap = emp.computeIntSamplePresence(
+                samples,
+                this.fields,
+                fieldsMap
+            );
 
-                // update fields mapping object
-                var result = emp._biom.getSampleValuesCount(samples, field);
-                fieldValues = Object.keys(result);
-                for (k = 0; k < fieldValues.length; k++) {
-                    fieldValue = fieldValues[k];
-                    fieldsMap[field][fieldValue] += result[fieldValue];
-                }
-            }
             SelectedNodeMenu.makeSampleMetadataTable(fieldsMap, this.smTable);
             this.smSection.classList.remove("hidden");
             this.smTable.classList.remove("hidden");
--- a/tests/test-empress.js
+++ b/tests/test-empress.js
@@ -677,5 +677,53 @@ require([
                     1.0e-15
             );
         });
+
+        test("Test computeLeafSamplePresence", function() {
+            var e = this.empress;
+            var fields = e._biom._smCols;
+            var ctData = e.computeLeafSamplePresence("2", fields);
+
+            var lf2presence = {
+                "f1": {"a": 4, "b": 1},
+                "grad": {"1": 1, "2": 2, "3": 2, "4": 0},
+                "traj": {"t1": 2, "t2": 1, "t3": 2, "t4": 0}
+            };
+            deepEqual(ctData, lf2presence);
+        });
+
+        test("Test computeIntSamplePresence", function() {
+            var e = this.empress;
+            e._tree.names_ = ["root", "internal", "1", "internal",
+                              "2", "3", "6"];
+
+            var fieldsMap = {};
+            var fields = e._biom._smCols;
+            for (i = 0; i < fields.length; i++) {
+                field = fields[i];
+                var possibleValues = e._biom.getUniqueSampleValues(field);
+                for (j = 0; j < possibleValues.length; j++) {
+                    var possibleValue = possibleValues[j];
+                    if (!(field in fieldsMap)) fieldsMap[field] = {};
+                    fieldsMap[field][possibleValue] = 0;
+                }
+            }
+
+            var nodeKey = 4;
+            var tips = e.findTips(nodeKey);
+            deepEqual(tips, ["2", "3"]);
+
+            var intersection = e._biom.getObsIDsIntersection(tips);
+            var diff = e._biom.getObsIDsDifference(tips);
+            var samples = e._biom.getSamplesByObservations(intersection);
+
+            var ctData = e.computeIntSamplePresence(samples, fields, fieldsMap);
+
+            var int4presence = {
+                "f1": {"a": 5, "b": 1},
+                "grad": {"1": 2, "2": 2, "3": 2, "4": 0},
+                "traj": {"t1": 2, "t2": 2, "t3": 2, "t4": 0}
+            };
+            deepEqual(ctData, int4presence);
+        });
     });
 });


gibsramen:
Totally forgot to stylecheck - will do so and update.

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
         this._events.selectedNodeMenu.hiddenCallback = callback;
     };
 
+    /**
+     * Calculate the number of samples in which a leaf node appears for each
+     * level of a given metadata field.
```
fedarko, line 1215:
```suggestion
     * Calculate the number of samples in which a leaf node appears for the
     * unique values of a metadata field across a list of metadata fields.
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     *
+     * @param {String} nodeName Name of the (leaf) node for which to calculate
+     *                          sample presence.
+     * @param {String} fields Metadata fields for which to calculate leaf
```
fedarko, line 1219:
```suggestion
     * @param {Array} fields Metadata fields for which to calculate leaf
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     *                       sample presence.
+     * @return {Object} ctData returns a mapping of field categories to number of
+     *                  samples in which the given leaf node is present and
+     *                  sample field value equals category.
```
fedarko, line 1223:
```suggestion
     * @return {Object} ctData Maps metadata field names to another Object,
     *                  which in turn maps unique metadata values to the number
     *                  of samples with this metadata value in this field that
     *                  contain the given leaf node.
```

fedarko: i'm not sure how best to word these -- feel free to alter these suggestions as you see fit :)

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+        return ctData;
+    };
+
+    /**
+     * Retrieve the tips in the subtree of a given node key.
+     *
+     * @param {Number} nodeKey Key value of node.
+     * @return {Array} tips Tips of the subtree.
+     */
+    Empress.prototype.findTips = function (nodeKey) {
```
ElDeveloper, line 1248: Could we move this method into the BPTree class? Something like BPTree.descendantTips(name)?

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     *                  samples in which tips of the internal node are present and
+     *                  sample field value equals category.
+     */
+    Empress.prototype.computeIntSamplePresence = function (
+        samples,
+        fields,
+        fieldsMap
+    ) {
+        // iterate over the samples and extract the field values
+        for (j = 0; j < fields.length; j++) {
```
ElDeveloper, line 1290:
```suggestion
        for (var j = 0; j < fields.length; j++) {
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+            field = fields[j];
+
+            // update fields mapping object
+            var result = this._biom.getSampleValuesCount(samples, field);
+            fieldValues = Object.keys(result);
+            for (k = 0; k < fieldValues.length; k++) {
```
ElDeveloper, line 1296:
```suggestion
            for (var k = 0; k < fieldValues.length; k++) {
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     *
+     * @param {String} nodeName Name of the (leaf) node for which to calculate
+     *                          sample presence.
+     * @param {String} fields Metadata fields for which to calculate leaf
+     *                       sample presence.
+     * @return {Object} ctData returns a mapping of field categories to number of
+     *                  samples in which the given leaf node is present and
+     *                  sample field value equals category.
+     */
+    Empress.prototype.computeLeafSamplePresence = function (nodeName, fields) {
```
ElDeveloper, line 1225: For consistency this should be `computeTipSamplePresence` right?

empress/support_files/js/select-node-menu.js
```diff
@@ -241,7 +241,7 @@ define(["underscore", "util"], function (_, util) {
         );
 
         // 2. Add sample presence information for this tip
-        var ctData = {};
```
fedarko, line 244: it looks like the code `computeLeafSamplePresence()` was based on is still located within this file. Could you please remove it?

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+
+        for (var f = 0; f < fields.length; f++) {
+            var field = fields[f];
+            var obs = this._biom.getObsCountsBy(field, nodeName);
+            var categories = _.keys(obs);
+            ctData[field] = {};
+            for (var c = 0; c < categories.length; c++) {
+                var cat = categories[c];
+                ctData[field][cat] = obs[cat];
+            }
```
fedarko, line 1236: I didn't notice this until now (this is probably my bad), but I think this code could be simplified a lot.

```suggestion
            var field = fields[f];
            ctData[field] = this._biom.getObsCountsBy(field, nodeName);
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     * @param {String} fields Metadata fields for which to calculate leaf
+     *                       sample presence.
+     * @return {Object} ctData returns a mapping of field categories to number of
+     *                  samples in which the given leaf node is present and
+     *                  sample field value equals category.
+     */
+    Empress.prototype.computeLeafSamplePresence = function (nodeName, fields) {
+        var ctData = {};
+
+        for (var f = 0; f < fields.length; f++) {
```
ElDeveloper, line 1228: Not sure if any of this would be handy, but I figured I would point you to the utility functions for Objects that are already implemented in underscore.js

https://underscorejs.org/#objects

tests/test-empress.js
```diff
@@ -677,5 +677,53 @@ require([
...
+        });
+
+        test("Test computeIntSamplePresence", function() {
+            var e = this.empress;
+            e._tree.names_ = ["root", "internal", "1", "internal",
+                              "2", "3", "6"];
+
+            var fieldsMap = {};
+            var fields = e._biom._smCols;
+            for (i = 0; i < fields.length; i++) {
```
ElDeveloper, line 701:
```suggestion
            for (var i = 0; i < fields.length; i++) {
```

tests/test-empress.js
```diff
@@ -677,5 +677,53 @@ require([
...
+                field = fields[i];
+                var possibleValues = e._biom.getUniqueSampleValues(field);
+                for (j = 0; j < possibleValues.length; j++) {
```
ElDeveloper, line 704:
```suggestion
                for (var j = 0; j < possibleValues.length; j++) {
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+                tips.push(t.name(node));
+            }
+        }
+
+        return tips;
+    };
+
+    /**
+     * Calculate the number of samples in which tips of an internal node
+     * appear for each level of metadata fields.
```
fedarko, line 1273:
```suggestion
     * Calculate the number of samples in which at least one tip of an internal
     * node appears for the unique values of a metadata field across a list of
     * metadata fields.
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     *
+     * @param {String} nodeName Name of the (leaf) node for which to calculate
+     *                          sample presence.
```
fedarko, line 1276:
```suggestion
     * @param {String} nodeName Name of the (internal) node to calculate
     *                          sample presence for.
```

fedarko: In general I think a few of the "for which"s throughout the docs here can be replaced with just "for"s. But it doesn't make a huge difference; I'll leave the choice up to you.

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     * @param {String} fields Metadata fields for which to calculate leaf
+     *                       sample presence.
```
fedarko, line 1278:
```suggestion
     * @param {String} fields Metadata fields for which to calculate internal
     *                        node sample presence.
```

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     * @param {Object} fieldsMap Object that maps fields to sample presence.
+     * @return {Object} fieldsMap Returns a mapping of field categories to number of
+     *                  samples in which tips of the internal node are present and
+     *                  sample field value equals category.
```
fedarko, line 1282: Would be nice to restructure this analogously to the suggestion I added for the `computeLeafSamplePresence()` return value.

empress/support_files/js/empress.js
```diff
@@ -1210,5 +1210,97 @@ define([
...
+     */
+    Empress.prototype.computeIntSamplePresence = function (
+        samples,
+        fields,
+        fieldsMap
+    ) {
+        // iterate over the samples and extract the field values
```
fedarko, line 1289: I think the docs for this function should be updated. At minimum, they should mention that `samples` is provided to this function, not computed here. (Also, this should probably mention that `fieldsMap` should be initialized so that all unique values' values in the inner Object are set to 0, as is done in `showInternalNode()` in the selected node menu file.)

So, as I understand it, this function doesn't actually have much to do with internal nodes -- the list of samples associated with some internal nodes is already provided to here. I would recommend doing one of two things:

1. Move more of the sample presence info computation code in `showInternalNode()` to this function -- e.g. initializing `fieldsMap`, computing `samples`, etc.

2. Keep the scope of this function as is, but rename it to something like `getSampleCountInformationForMetadata()` that makes it clearer that this doesn't really consider the structure of the tree.

I realize that either of these options (my vote is for option 1 personally, but up to you) are pretty nontrivial, so feel free to LMK if you have any questions.

tests/test-empress.js
```diff
@@ -677,5 +677,53 @@ require([
...
+                "grad": {"1": 1, "2": 2, "3": 2, "4": 0},
+                "traj": {"t1": 2, "t2": 1, "t3": 2, "t4": 0}
+            };
+            deepEqual(ctData, lf2presence);
+        });
+
+        test("Test computeIntSamplePresence", function() {
+            var e = this.empress;
+            e._tree.names_ = ["root", "internal", "1", "internal",
+                              "2", "3", "6"];
```
fedarko, line 697: Is this attribute used anywhere? Sorry if I'm missing something.

tests/test-empress.js
```diff
@@ -677,5 +677,53 @@ require([
...
+                }
+            }
+
+            var nodeKey = 4;
+            var tips = e.findTips(nodeKey);
+            deepEqual(tips, ["2", "3"]);
+
+            var intersection = e._biom.getObsIDsIntersection(tips);
+            var diff = e._biom.getObsIDsDifference(tips);
+            var samples = e._biom.getSamplesByObservations(intersection);
```
fedarko, line 717: yeah, the fact that all of this work has to be done in this test relates to the point above re: moving more functionality into `computeIntSamplePresence()`. Ideally it would be nice to be able to pass just the node key and `fields` into that function and get the mapping without the caller having to do any extra work.

tests/test-empress.js
gibsramen: Yeah, this line is (as far as I can tell) necessary to compute the tips of the internal node. Excluding it results in `TypeError: Cannot read property '4' of null` in returning the names of the nodes in BPTree.

fedarko: Ahhh, got it. Thanks for the explanation!

It isn't super important, but I think it'd make sense to have the names be defined in the BPTree test setup, rather than within this test: it looks like the creation of `var tree` at the top of `test-empress.js` can be modified as follows --

```js
                var tree = new BPTree(
                    new Uint8Array([1, 1, 1, 0, 1, 1, 0, 1, 0, 0, 0, 1, 0, 0]),
                    ["root", "internal", "1", "internal", "2", "3", "6"], // replaces the null that was previously here
                    null,
                    null
                );
```

Just tried this out, and it looks like this prevents that TypeError and doesn't interfere with any of the other tests here.

tests/test-empress.js
gibsramen: Yeah that makes sense. Do you have a suggestion for how to handle the behavior of `_checkAndFilterTips` where it warns the user of tips not present in the BIOM table? It's not immediately clear to me how to translate this behavior from `select-node-menu.js` to `empress.js`? This function uses callbacks from `SelectedNodeMenu` that I'm not sure how to address.

fedarko: Yeesh, I didn't think about that. Hm.

This is tough, because we want to move as much stuff as possible into Empress.js, but we also want to avoid redoing work (i.e. computing `samples` multiple times).

I guess keeping things so that `computeIntSamplePresence()` requires a list of samples to be provided would make sense (although as stated above I'd recommend the function then be renamed for clarity's sake).

Another way would be having `computeIntSamplePresence()` compute the `intersection` (as in the second line of `_checkAndFilterTips()`) itself, and just have it return `samples` along with `fieldsMap`. This way the samples could still be added to `_samplesInSelection` without problems. (However, this'd mean that `_checkAndFilterTips()` should probably be restructured to just compute the `diff` and open the toast message, without doing intersection-related stuff... although this'd also mess up the leaf presence code, and require some extra work there.)

Yeah, eesh, sorry. I've been staring at this code for like 30 minutes now and can't think of an ideal solution. Whatever sounds best to you is ok with me -- we probably can't make this perfect, but we can at least come close-ish :)

gibsramen: Ok, cool. What I have right now is similar to the latter. `computeIntSamplePresence` now returns an object like so which feels inelegant:

```javascript
var intSamplePresence = {
    "fieldsMap": fieldsMap,
    "diff": diff,
    "samples": samples
}
```

So the code in `select-node-menu.js` can handle the toasting and Emperor callback.

fedarko: That sounds perfectly fine to me, not inelegant at all. This approach actually matches up with some other places in Empress' codebase where one function returns multiple things, e.g. [this line](https://github.com/biocore/empress/blob/3a6bf88211621e7b4fd11c99568bac8fe25a715b/empress/support_files/js/vector-ops.js#L127) in one of the thick-line-drawing utility functions.

Thanks for working on this! :shipit:

empress/support_files/js/empress.js
```diff
@@ -1231,7 +1231,89 @@ define([
...
+    /**
+     * Calculate the number of samples in which at least one tip of an internal
+     * node appears for the unique values of a metadata field across a list of
+     * metadata fields.
+     *
+     * @param {String} nodeKey Key of the (internal) node to calculate
+     *                         sample presence for.
+     * @param {Array} fields Metadata fields for which to calculate internal
+     *                       node sample presence.
+     * @return {Object} samplePresence Returns a mapping of three entries:
```
fedarko, line 1266:
```suggestion
     * @return {Object} samplePresence A mapping with three entries:
```

empress/support_files/js/empress.js
```diff
@@ -1231,7 +1231,89 @@ define([
...
+     *                                 (1) fieldsMap Maps metadata field names
+     *                                 to Object mapping unique metadata values
+     *                                 to the number of samples with this metadata
+     *                                 value in this field containing at least one
+     *                                 tip in the subtree of the given nodeKey.
+     *                                 (2) diff Array of tip names not present
+     *                                 as features in the table
+     *                                 (3) samples Array of samples represented by
+     *                                 tips present in the table.
```
fedarko, line 1275: Summarizing things in these sorts of complicated situations is always a mouthful, but these docs do a great job of communicating things. Really well done!

empress/support_files/js/empress.js
```diff
@@ -1231,7 +1231,89 @@ define([
...
+            field = fields[k];
+
+            // update fields mapping object
+            var result = this._biom.getSampleValuesCount(samples, field);
+            fieldValues = Object.keys(result);
+            for (var m = 0; m < fieldValues.length; m++) {
+                fieldValue = fieldValues[m];
+                fieldsMap[field][fieldValue] += result[fieldValue];
+            }
+        }
```
fedarko, line 1306: I feel like we should be able to combine this for loop with the one above it (the one using `var i = 0`)... but I recognize this is code that's already been in here a while, so it's not your job to fix everything up right now. Might be worth making a new issue so we can keep track of this for future work.

empress/support_files/js/select-node-menu.js
```diff
@@ -320,73 +311,33 @@ define(["underscore", "util"], function (_, util) {
...
-            var possibleValues = this.empress._biom.getUniqueSampleValues(
-                field
-            );
-            for (j = 0; j < possibleValues.length; j++) {
-                var possibleValue = possibleValues[j];
-                if (!(field in fieldsMap)) fieldsMap[field] = {};
-                fieldsMap[field][possibleValue] = 0;
-            }
-        }
+        var field, fieldValues, fieldValue, i, j, k;
```
fedarko, line 314: Doesn't look like these variables are used here any more. Could you please remove this line? (sorry if I'm missing something and one of these is actually used... in that case feel free to leave parts of this line remaining as desired :)

empress/support_files/js/select-node-menu.js
```diff
@@ -320,73 +311,33 @@ define(["underscore", "util"], function (_, util) {
...
+            this._samplesInSelection = this._samplesInSelection.concat(
+                samplePresence.samples
+            );
+
+            this._checkTips(samplePresence.diff);
+
+            SelectedNodeMenu.makeSampleMetadataTable(
+                samplePresence.fieldsMap,
+                this.smTable
+            );
```
fedarko, line 340: The trimmed-down code is so pretty :D

empress/support_files/js/select-node-menu.js
```diff
@@ -406,16 +357,9 @@ define(["underscore", "util"], function (_, util) {
 
     /**
      * Given an array of tip names, warns the user about those that are not
-     * present in the BIOM table (using a toast message) and returns just the
-     * tip names in the array that are present as features in the table.
-     *
-     * @return {Array} the tip names that are also represented in the table.
+     * present in the BIOM table (using a toast message)
```
fedarko, line 360: GitHub isn't letting me make a suggestion here, but this docstring should be trimmed down slightly to just something like "Given an array of tip names that are not present in the BIOM table, warns the user about them using a toast message" (along those lines). Since this function no longer actually computes `diff`, we should make it clear that the caller of the function should already have ensured that these tips are the `diff`. If ... that doesn't sound too silly.

tests/test-empress.js
```diff
@@ -677,5 +685,31 @@ require([
...
+            var fields = e._biom._smCols;
+            var values = e.computeIntSamplePresence(4, fields);
+
+            var int4presence = {
+                f1: { a: 5, b: 1 },
+                grad: { "1": 2, "2": 2, "3": 2, "4": 0 },
+                traj: { t1: 2, t2: 2, t3: 2, t4: 0 },
+            };
+            deepEqual(values.fieldsMap, int4presence);
+        });
```
fedarko, line 713: Would you mind adding another test that uses the node with key `7` (the root of the test tree) as the internal node to compute presence info for? This is a common enough use case that IMO we should explicitly test it.

(Figuring out the expected presence information should be pretty straightforward, since all samples should be "captured" by the root -- so every entry in the sample metadata [aka the `var sm` defined towards the top of this test file] should be represented in this information. So there'll be 5 `a`s and 2 `b`s for `f1`, etc.)

empress/support_files/js/empress.js
gibsramen: Makes sense. I'll just keep it like this for now and then we can open an issue about it.

ElDeveloper:
Thanks so much both! 🎉

Issue #222, opened 2020-06-29
Browser / ES version considerations


fedarko:
From code review:

- [ ] Possible to use `Map` instead of `Object`. [Pretty much all non-IE major browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) support this, and there seem to be some performance benefits besides the obvious extra clarity of seeing that something is explicitly a mapping

- [ ] Be consistent with `var`, `let`, and `const`. I just did some quick estimates by running `grep -ri "let " empress/support_files/js/* | wc -l`, etc. and found that there are **412 `var`s**, **9 `let`s**, and **4 `const`s** in the Empress JS codebase. We should prooobably just make everything `var`.

- [ ] Use foreach instead of for loops? Where possible, and where it wouldn't be an efficiency decrease. (`Map`s can apparently be directly iterated over, so maybe those would pair well with this.)

We may want to go about this by figuring out what browser / JS version we want to target, and then... adjusting our build setup accordingly? Maybe by adjusting jshint, or adding Babel somewhere to verify that we can use these things and still target all the environments we want to support. I'm not really sure what the best practice would be (I suspect it'd involve a lot of messing around with adding webpack or rollup to the project).

Thanks @dhakim87, @wasade, and @qiyunzhu for bringing up these points respectively during code review today!

fedarko:
From discussion with @ElDeveloper and @kwcantrell, it sounds like we're mostly moving towards just supporting Firefox / Chrome for now. So using JS features that e.g. break IE is ok.

PR #275, opened 2020-07-23
Validate input line widths; scale widths based on number of leaves in the tree

### LINKED ISSUE #135: Possible to set a negative line width
I assume this isn't intended (if it's actually a feature feel free to close this :) -- assuming it isn't, should be possible to add in `min="1"` or something to the corresponding `<input>` element.

Edit: should also add JS that validates that the new branch thickness is within a range of acceptable values, as a failsafe. Relates to #144.

Screenshot of a -10 width below (I admit it does look kind of cool, though) --

issue

fedarko:
Closes #135. This sort of addresses #144, but I'm not suuuper satisfied with my solution to that here so I think it might be best to leave that open for now.

### Work on #135 (possible to set negative line widths in the UI) 
Adds `util.parseAndValidateLineWidth()`, which takes as input a HTML element and checks to see if it represents a "valid" line width or not. If not, it changes the input's value to 0; in any case, it returns a float representation of the input's (new?) value. This function is used to process all 3 line width inputs in the application (sample metadata, feature metadata, animation).

### Work on #144 (scaling line widths to look more consistent across differently-sized trees)
I tried a bunch of different things -- at first I had a function that computed the area of the visualization and then scaled line widths based on that (see comments in #144 for more info on this plan; a version of the repo that does this is [here](https://github.com/fedarko/empress/tree/db647fad05928b0b6de376022c6daa0ec46365e1)). However, the thing with this is that the layouts (currently on the python side of things) all scale the coordinates to fit within ~4020 x 4020 ... so all of the trees I tried had roughly the same "area," as far as the JS code could tell, meaning that this didn't really do much (line widths were still kind of inconsistent between mid- and small- sized trees). I also tried @kwcantrell's suggestion of mimicking how [`zoomAmount` is set in empress.js](https://github.com/biocore/empress/blob/4987a806ca3ce2d0a5538905f019434f1026ed0b/empress/support_files/js/empress.js#L1171) and using that as a scaling factor, but this didn't seem to make that much of a difference when I tried it.

I think it should be possible to either take the scaling factors for each layout in python and pass that information to the JS (or take the area of the unscaled layouts and pass that to the JS), but for a variety of reasons* I settled on doing things in a simpler way -- by just scaling line widths using the number of leaves in the tree. The formula I settled on (as of right now) is

`(2 * lw) / Math.pow(Math.log10(this._tree.numleaves()), 2)`

Where `lw` is the input line width.

This seems to perform decently well for multiple sizes of trees. It's not perfectly consistent, but things seem to at least be in similar ballparks. Here are two datasets with the same set line width of 2, one with 5 leaves (a test dataset I set up, LMK if you want me to put it on dropbox or something) and one with 770 leaves (just the moving pictures tree zoomed in on a clade) --

image

image

Anyway -- as mentioned there are definitely better ways to handle this, but for now this is hopefully sufficient.

\* including but not limited to: I wasn't sure how to handle things when there were different x and y scaling factors (i.e. in the rectangular layout); I didn't think of computing the area of the unscaled layout until I started writing this PR message; eventually layouts will be computed in the JS anyway; and at this point I've spent enough time wrestling with this issue that I think everyone, me included, would prefer for me to get back to working on barplots :)

### Other fixes

In addition to the above stuff, this PR changes the default line widths to 0 and renames `Line Width` to `Extra Line Width` in the UI. This lets us scale line width in a more easily understandable way. I also added [`step="any"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step) to the line width inputs, so now they shouldn't complain when you enter in float widths (e.g. 0.5). There are also a few other misc. grammar / code fixes throughout this PR (e.g. fixed a bug in the toast function and increased the default duration; abstracted out some reused code in util.js; etc.), which will hopefully make things slightly nicer.

Thanks!


empress/support_files/js/util.js
```diff
@@ -100,6 +98,24 @@ define(["underscore"], function (_) {
...
+     *
+     * This function was created from code that was originally in
+     * splitNumericValues() in Emperor's codebase. That code, in turn, was
+     * based on http://stackoverflow.com/a/9716488.
+     *
+     * @param {String} value Value to check for validity
+     * @return {Boolean} true if value represents a finite number, else false
+     */
+    function isValidNumber(value) {
+        return !isNaN(parseFloat(value)) && isFinite(value);
```
kwcantrell, line 116:
```suggestion
        return !isNaN(parseFloat(value)) && isFinite(parseFloat(value));
```
This check will fail if value is "1/3".

empress/support_files/js/util.js
```diff
@@ -126,25 +141,58 @@ define(["underscore"], function (_) {
...
+     * @param {Number} duration - milliseconds to keep toast visible (optional)
+     *                            If this is undefined, then this will default
+     *                            to 3000 ms (i.e. 3 seconds).
      */
-    function toastMsg(msg) {
-        var duration = 2000,
-            toast = document.getElementById("toast");
+    function toastMsg(msg, duration) {
+        var effectiveDuration = _.isUndefined(duration) ? 3000 : duration;
+        var toast = document.getElementById("toast");
```
kwcantrell, line 153:
```suggestion
    function toastMsg(msg, duration=3000) {
        var toast = document.getElementById("toast");
```

Not a big deal, I just think it looks a little cleaner.

empress/support_files/js/util.js
```diff
@@ -126,25 +141,58 @@ define(["underscore"], function (_) {
...
+     *                               these things are checked for here -- we
+     *                               only look at the value of this element.
+     * @return {Number} Sanitized number that can be used as input to
+     *                  Empress.thickenSameSampleLines().
+     */
+    function parseAndValidateLineWidth(inputEle) {
+        if (isValidNumber(inputEle.value)) {
+            var pfVal = parseFloat(inputEle.value);
+            if (pfVal >= 0) {
+                return parseFloat(inputEle.value);
```
kwcantrell, line 183:
```suggestion
                return pfVal;
```

empress/support_files/js/empress.js
```diff
@@ -566,22 +565,47 @@ define([
...
+            // accordingly. This way, when doing things like updating the
+            // layout that'll require re-drawing the tree based on the most
+            // recent settings, we'll have access to the correct line width.
+            this._currentLineWidth = lw;
+            if (lw === 0) {
+                // But, yeah, if lw is 0 we can just return early.
+                return;
+            }
+        }
+        this._currentLineWidth = lw;
```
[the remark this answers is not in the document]
fedarko: Excellent point, thanks for catching this!

empress/support_files/js/util.js
fedarko: Weirdly enough, the recommended way to do this check seems to be without calling `parseFloat()` on the value passed to `isFinite()` -- see https://stackoverflow.com/a/9716488/10730311.

This function does return false if the input is `"1/3"`, but this is a good thing IMO: JavaScript does not parse fraction strings as expected --

```js
> parseFloat("1/3");
1
```

Changing the passed value to `isFinite()` to use `parseFloat()` causes `isValidNumber("1/3")` to be true, which is not desirable because we'd then interpret it as `1` misleadingly.

That all said, I thought it'd be good to add some unit tests in for `isValidNumber()` that directly check this functionality -- was working on that when I ran into those QUnit problems I mentioned on slack... once I get around those I'll update this.

Does this all sound reasonable?

empress/support_files/js/util.js
fedarko: Agreed that this does seem nicer. It looks like default parameters don't work on Internet Explorer, though (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#Browser_compatibility)). @ElDeveloper we're not officially supporting IE, right?

Relevant to #222...

empress/support_files/js/util.js
kwcantrell: makes sense

fedarko:
Ok, changes should be in to the line width PR now (now including those tests for isValidNumber() on the "1/3" case). Thanks all!

empress/support_files/js/util.js
kwcantrell: No, I don't think we are supporting IE. I know multiple functions (such as the modifications Antonio made to bp-tree) use default args. I also have some recollection that some other code we used on is not support in IE (although I don't remember what that was).

tests/test-biom-table.js
```diff
@@ -562,7 +550,7 @@ require(["jquery", "BiomTable"], function ($, BiomTable) {
             deepEqual(
                 obsReturned,
                 {
-                    3: new Set([["o1", "o3", "o5", "o6", "o7", "o9"]]),
+                    3: new Set(["o1", "o3", "o5", "o6", "o7", "o9"]),
```
kwcantrell, line 553: I just wanted to verify that the update version of QUnit can compare Sets. I noticed you updated converted the tests in getObs from Sets to Arrays but didn't do that here.

fedarko: Ah, yeah, good question. From what I can tell the new version of QUnit can compare Sets properly -- if you alter the `getObsBy()` tests to still use an Object of Sets (but have the Sets using the correct feature IDs), the test still works.

The reason I changed those tests to use arrays was because that was before I realized what the problem was, and wanted to see if using arrays would fix things. (If you'd like things to be more consistent here, I can either update the rest of these tests to use sorted arrays or update the `getObsBy()` tests to use sets; either would work.)

The reason _this_ particular line has the inner brackets removed is because the initial way of writing it (`new Set([["o1", "o3", "o5", "o6", "o7", "o9"]])`) creates a Set with a single element, which is that inner array. This differs from the actual expected output, which is a Set containing each of those strings as top-level elements if that makes sense. (The old version of QUnit didn't consider this to be a problem since it couldn't compare sets, but the updated version of QUnit failed on this test. Which ... was a good thing :)

kwcantrell: Would you be able to revert getObsBy back to using Sets then? That way it is more clear since it will match the output of getObsBy

fedarko: Just converted it back to using sets. Thanks for the feedback :)

kwcantrell:
thanks @fedarko!

PR #277, opened 2020-07-24
Node collapse

kwcantrell:
I am creating this PR to jump start the review process since this PR is fairly large. I am still working on finishing up some tests however, the important parts are tested.

 
Notes
----------------
Each layout has a different method to collapse clades as seen below:
| Unrooted | Rectangular | Circular|
| ------------ | ---------------- | ---------- |
| collapse-unroot |  collapse-rectangular | collapse-circ

In addition, Rectangular and Circular layout will also have a "symmetric" version as seen below:
| Normal | Symmetric |
| ---------- | -------------- |
| collapse-rectangular | rec-sym |
| collapse-circ |  cir-sym |

Collapsed clade can also be clicked on. Doing so, at the time being, is equivalent to selecting the root node. In the future, clicking on a clade will display information about the clade (such as its group membership, # tips, # internal node,...)

It should also be noted that an issue can occur with the Unrooted/Circular layout if the clade is more than a 180°:
circ-bug
unroot-lay

This is not a bug per se but a consequence of the collapsing algorithm (I believe iTol shares the same issue). This happens because the "left" and "right" most children of the clade are behind the root. This, in turn, causes the clade shape to do one of two things: point in the wrong direction (this happened in the above Unrooted layout) or overlap other clades/root (this happened in the circular layout).

Minor changes:
- cangle has been added treeData for each tip
- lengths has been added to bp-tree
- an in-order tree traversal has been added to bp-tree (and will be cached after it is called)
- inSampe has been removed from treeData
- hide checkbox has been replaced with collapse checkbox
- setNonSampleBranchVisibility() has been removed


empress/support_files/js/empress.js
```diff
@@ -1230,5 +1264,621 @@ define([
...
+                addPoint(getCoords(child));
+            };
+
+            // triangle from 1, 4, 2
+            addTriangle(cladeInfo.left);
+
+            // triangle from 1, 4, 3
+            addTriangle(cladeInfo.right);
+        } else if (this._currentLayout === "Rectangular") {
+            // Rectangular layou is a triangle. Symmetric version is used if
```
ElDeveloper, line 1417:
```suggestion
            // Rectangular layout is a triangle. Symmetric version is used if
```

empress/support_files/js/empress.js
```diff
@@ -1230,5 +1264,621 @@ define([
...
+                x = sX * cos - sY * sin;
+                y = sX * sin + sY * cos;
+                addPoint([x, y]);
+            }
+        }
+
+        this._collapsedCladeBuffer.push(...cladeBuffer);
+    };
+
+    /**
```
ElDeveloper, line 1520: MInor but this docstring is probably missing the `@private` tag.

empress/support_files/js/empress.js
```diff
@@ -1230,5 +1264,621 @@ define([
...
+        }
+        // stores the clade nodes
+        var cladeNodes = [];
+
+        // Nodes in the clade are found by performing a postorder traversal
+        // starting at the left most child of the clade and ending on cladeRoot
+
+        // find left most child
+        // Note: initializing lchild as cladeRoot incase cladeRoot is a tip
+        lchild = cladeRoot;
```
ElDeveloper, line 1672:
```suggestion
        var lchild = cladeRoot;
```

empress/support_files/js/empress.js
```diff
@@ -1230,5 +1264,621 @@ define([
...
+        var cRoot = getCoords(cladeRoot);
+        var left = getCoords(clade.left);
+        var right = getCoords(clade.right);
+        var deep = getCoords(clade.deepest);
+        if (this._currentLayout === "Unrooted") {
+            // In Unrooted layout, to check if point is within in the collapsed
+            // clade, we first calculate the area of the collapsed clade.
+            // (The shape of the collapsed clade is a quad whose vertices are
+            // (1) root, (2) "left" most child, (3) "right" most child, and
+            // (4) "deepest" child). Next, we form four triangls whose vertices
```
ElDeveloper, line 1721:
```suggestion
            // (4) "deepest" child). Next, we form four triangles whose vertices
```

empress/support_files/js/empress.js
```diff
@@ -1230,5 +1264,621 @@ define([
...
+    /**
+     * Checks if (x, y) is within the bounds of a collapsed clade
+     *
+     * @param {Array} point (x, y) coordinates of a point
+     *
+     * @return {Number} if point is in a collapsed clade then root of the
+     *                  the collapse clade will be returned otherwise -1 is
+     *                  returned.
+     */
+    Empress.prototype.isInAClade = function (point) {
```
ElDeveloper, line 1858: The name of this method is a bit misleading, would likely be more appropriate as `getRootNodeForPointInClade` - or something like that up to you. `isInClade` makes it sound like it returns a boolean value if the point is in a clade, but really it is returning the root for a clade if the point falls in that space.

tests/test-empress.js
```diff
@@ -677,5 +685,115 @@ require([
                     1.0e-15
             );
         });
+
+        test("Test collapseClades", function () {
+            // red should be the only collapsable clade
```
ElDeveloper, line 690:
```suggestion
            // red should be the only collapsible clade
```

empress/support_files/templates/side-panel.html
```diff
@@ -174,6 +188,28 @@
...
+    <input type="radio" id="normal" name="collapse-radio" value="normal"
+           checked="true" class="empress-input">
+  </p>
+  <p class="side-panel-notes indented">
+    The shape of the collapsed clade will show both the depth and width of the
+    clade.
+  </p>
+  <p>
+    <label for="symmetric">Symmetric</label>
+    <input type="radio" id="symmetric" name="collapse-radio" value="symmetric"
```
ElDeveloper, line 205: It would be nice if this input could be disabled when the unrooted layout is set.

empress/support_files/templates/side-panel.html
```diff
@@ -132,10 +143,13 @@
     </label>
   </p>
   <p>
-    <label for="animate-hide-non-feature">Hide Non-Sample Features</label>
-    <input id="animate-hide-non-feature" type="checkbox" value="false"
+    <label for="animate-collapse-chk">Collapse Clades</label>
+    <input id="animate-collapse-chk" type="checkbox" value="false"
```
ElDeveloper, line 147: Should this input be disabled once the animation starts?

empress/support_files/js/empress.js
```diff
@@ -1230,5 +1264,621 @@ define([
...
+            color = this._treeData[node].color;
+            var isTip = this._tree.isleaf(this._tree.postorderselect(node));
+
+            if (
+                visible &&
+                !isTip &&
+                group[node] !== -1 &&
+                color !== this.DEFAULT_COLOR
+            ) {
+                this._collapseClade(node);
```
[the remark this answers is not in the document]
kwcantrell: Good catch! It would definitely collapse two clades that have the same color but different metadata values. Likely, should be an easy fix. Instead of using the color to group nodes, I can just use metadata.

kwcantrell:
@ElDeveloper thank you for the comments. I believe I have addressed all of your comments. Note that clades are now collapsed based on the _group array (which uses metadata value to set the groups) instead of color. I also added some more tests. 

Also, I don't have an issue with moving the collapse code to a different module. Although I think we should push the to another PR since it would require a bit of refactoring.

empress/support_files/js/empress.js
```diff
@@ -1502,6 +1561,619 @@ define([
...
+        // Once a internal node is reached that belongs to a group (i.e. not -1)
+        // than that node will be marked as the root of the clade and then
+        // collaped.
+        // Collapsing a clade will set the .visible property of members to
+        // false and will then be skipped in the for loop.
+        // Note: if the root of a clade has DEFUALT_COLOR then it will not be
+        // collapsed (since all of its children will also have DEFAULT_COLOR)
+        var inorder = this._tree.inOrderNodes();
+        for (var node in inorder) {
+            node = inorder[node];
```
ElDeveloper, line 1638: This pattern can be changed for:

_.each(inorder, function(node) {
    
})

tests/python/test_core.py
```diff
@@ -254,6 +254,7 @@ def test_to_dict(self):
         viz = Empress(self.tree, self.table, self.sample_metadata,
                       filter_unobserved_features_from_phylogeny=False)
         obs = viz._to_dict()
+        print(obs)
```
ElDeveloper, line 257:
```suggestion
```

tests/test-animation-panel-handler.js
```diff
@@ -158,34 +158,60 @@ require([
     test("__resumeOptions", function () {
         // set up so that animator state machine is on first frame
         this.panel.animator.totalFrames = 3;
-        this.panel.animator.curFrame = 1;
+
+        // Note: this use to be 1 but the animator class has been updated
```
ElDeveloper, line 162:
```suggestion
        // Note: this used to be 1 but the animator class has been updated
```

ElDeveloper:
Thanks @kwcantrell

PR #280, opened 2020-07-28
DOC: Improve contributing docs

### LINKED ISSUE #262: Update contributing docs re: current workflow
- The top section should mention `pip install -e .[all]` as the ideal way to install Empress in a dev environment (the README's instructions no longer use `pip install .` for Empress, and although the front-end dev section now mentions this it's too far down IMO)

- Should request that the dev version of Emperor is installed (this isn't explicitly stated anywhere, and if this isn't done [i.e. the default version of Emperor in QIIME 2 is used] then some of the python tests will break due to callback stuff not having a place to be inserted)

- Should mention that, if messing around with qunit-puppeteer isn't possible / desirable, it's possible to run tests for the JS code using the `tests/index.html` file

Shoutouts to @gibsramen for making me aware of these issues :rocket:

ElDeveloper:
Fixes #262

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,8 +3,10 @@
 ## Setting up an Empress development environment
 
 You should be able to install Empress as described in the README. However,
-instead of running `pip install .`, please run `pip install .[all]`. This will
-install some various dependencies needed to test Empress' python code.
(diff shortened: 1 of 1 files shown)


kwcantrell:
thanks @ElDeveloper

fedarko:
Ditto, thanks @ElDeveloper for fixing these!

PR #281, opened 2020-07-29
Rename a few things in the JS code

### LINKED ISSUE #186: Rename the coloring utilities in empress.js to focus on "groups", not "samples"
See [this thread](https://github.com/biocore/empress/pull/183/files#r442383438) for details. This would make this code more readable.

fedarko:
Closes #186. This doesn't remove `.inSample`, since that is already being removed in #277 anyway.


ElDeveloper:
Looks great, thanks so much @fedarko!

Issue #278, opened 2020-07-25
Add additional configuration settings in tree properties


ElDeveloper:
- [ ] Checkbox to set whether or not to ignore unrepresented tips for color propagation (default is to ignore them).
- [ ] Color select for shared branches i.e. when a branch is represented by multiple categories it should be displayed in this color (default is no color, which means the color of this branches is unchanged).

PR #282, opened 2020-07-29
ENH: Focus on node when clicking on an arrow

ElDeveloper:
Before, the menu would show but the camera wouldn't refocus.

### DIFF
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -1495,7 +1495,7 @@ define([
         }
 
         this._events.selectedNodeMenu.clearSelectedNode();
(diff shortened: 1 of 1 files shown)


kwcantrell:
thanks @ElDeveloper!

Do you think it might be better to set whether or not the camera should refocus on the node in this case (as well as when a user searches for a node) as a global parameter? I can see cases where the user may not want to refocus the camera when clicking on an arrow (or searching for a node). For example, the user wants to see a nodes location w.r.t the whole tree.

ElDeveloper:
I'll submit a separate PR to add those settings and some the color propagation setting in #278

kwcantrell:
thanks @ElDeveloper!

PR #283, opened 2020-07-29
Add settings to ignore absent tips, and not change camera focus on selected tips

ElDeveloper:
In the gif below, two new settings to control how absent tips are treated in sample coloring, and what to do when selecting a node. 

setting


kwcantrell:
@ElDeveloper this looks good! If you can just add a simple test to make sure _projectObservations() works as expected when  ignoreAbsentTips is set. After that, I think this should be good to merge.

kwcantrell:
thanks @ElDeveloper!

PR #288, opened 2020-07-31
Fixed error

kwcantrell:


empress/support_files/js/select-node-menu.js
```diff
@@ -244,9 +244,14 @@ define(["underscore", "util"], function (_, util) {
         var ctData = this.empress.computeTipSamplePresence(name, this.fields);
 
         // 2.1 The samples represented by this tip are sent to Emperor
-        this._samplesInSelection = this.empress._biom.getObsIDsIntersection([
-            name,
-        ]);
+        // check if name if a feature in the biom table if the length is 0
```
fedarko, line 247:
```suggestion
        // check if this tip is present in the BIOM table. The array returned
        // by BIOMTable.getObsIDsDifference() contains the feature IDs present
        // in the input array but not in the BIOM table -- so if the length of
        // this array is zero, this feature is present in the table.
```

I spent longer than i'd like to admit staring at this block of code, so i tried to rewrite the comment in a way I found more understandable. feel free to change things around as you like

kwcantrell:
thanks @fedarko for the comments. I applied both your suggestions so the PR should be good to go.

fedarko:
Thanks @kwcantrell! This looks good to me.

PR #293, opened 2020-08-04
Draw feature / sample metadata tip-level barplots in the rectangular layout

### LINKED ISSUE #201: Barplots in the circular / rectangular layout
This was mentioned in #97 (which has since been closed, since the focus of that was on the circular layout).

Now that the circular layout is implemented and tested, supporting visualizing tip-level feature metadata as barplots would be a really cool feature to add. This could be useful for a few different types of feature metadata, ranging from Songbird/ALDEx2/... differentials (or other "importance scores") to taxonomy annotation confidence values, etc.

fedarko:
Closes #201. (I guess once we merge this in we should open up a new issue for extending this to the circular layout / showing legends / other enhancements we want to add on.)

Arbitrary numbers of barplot layers can be added. Each layer can encode either feature or sample metadata: feature metadata barplots are more straightforward (e.g. "importance scores for tips", "assigned taxonomy of tips"), while sample metadata barplots currently only support stacked barplots of sample presence information (e.g. "which body sites was this tip present in?")

Currently, this only supports the rectangular layout, but generalizing this to draw rings in the circular layout should be very possible. When a non-supported layout is selected, barplots will be hidden; they'll be reshown upon the layout changing back to a supported one.

barplots

Potentially-problematic things that we should discuss in a review:

1. Although the height is set based on the y-scaling factor from the rectangular layout (and thus scales well with arbitrarily small or large trees), the widths are in arbitrary units (the same as line thicknesses).
2. Does the length-scaling formula seem mathematically sound? I think it seems reasonable, but this isn't my strong suit. (The quantitative color-scaling stuff I took from Emperor, so that should be solid.)
3. Drawing sample metadata barplots is kind of slow for the HMP dataset (takes ~10 seconds on my system). Would be good to speed this up.
4. This adds in Spectrum.js, which is also used in emperor. Although our use of Require.js should mean that the JS files shouldn't conflict (thanks Yoshiki!), should we worry about the CSS conflicting? Our Spectrum version is slightly newer than Emperor's, but we could use the exact same JS/CSS as Emperor if desired.

I'll try to add in legend support / fix some of the above issues as I can before release today, but wanted to get in what I have now. Thanks!


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F293%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F293%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F293%2Fempress-tree.qzv)

empress/support_files/js/barplot-layer.js
```diff
@@ -0,0 +1,501 @@
...
+     *                                    notify the BarplotPanel if this layer
+     *                                    is removed.
+     * @param {HTMLElement} layerContainer DOM element to which the new layer's
+     *                                     HTML should be added.
+     * @param {Number} num The "number" of this layer. Layer 1 should be
+     *                     the closest layer to the tips of the tree, and
+     *                     so on.
+     *
+     * @return {BarplotLayer}
+     * construct BarplotLayer
```
fedarko, line 27:
```suggestion
     * @constructs BarplotLayer
```

kwcantrell:
This is very cool @fedarko!

empress/support_files/js/util.js
```diff
@@ -1,4 +1,4 @@
-define(["underscore"], function (_) {
+define(["underscore", "chroma"], function (_, chroma) {
```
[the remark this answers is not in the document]
fedarko: Thanks for catching this! You're completely right -- I initially defined `hex2RGB()` as a function in `util.js`, and then I figured it'd make more sense to lump it in with the `Colorer` stuff so I moved it out. Looks like I forgot to delete the chroma import :) -- will do that now.

fedarko:
OK, this should be ready to merge now; added more stuff to the README, including details on adding extra barplot layers and using sample metadata coloring, and removed some of the stuff on length scaling (I imagine when we get to #300 we can add more details for that in to the README). Thanks!

kwcantrell:
thanks @fedarko

README.md
```diff
@@ -34,23 +34,24 @@ In this tutorial we will be using Empress within the QIIME 2 environment, and de
 
 ## Downloading Input Artifacts and Metadata  
 
-Before we start, we’ll need to download the necessary input artifacts which are produced following the [Moving Pictures Tutorial](https://docs.qiime2.org/2020.6/tutorials/moving-pictures/), these are:  
+Before we start, we’ll need to download the necessary input artifacts -- all of which are produced during the [Moving Pictures tutorial](https://docs.qiime2.org/2020.6/tutorials/moving-pictures/). These are:
```
mestaki, line 37: Note that the biplot is not actually produced during Moving Pictures tutorial. It is based on MP but the tutorial doesn't make it. We grab this from a different source. Just semantics though

fedarko: Good catch, thanks for noticing this! This is good to point out -- I'll submit a PR that fixes this.

PR #301, opened 2020-08-05
Fix 258

kwcantrell:
This PR converts treeData from a dictionary to an array. One thing to note is that the first element of treeData is 0. This is equivalent to null and exists because currently empress uses 1-based index for postorder positions. This will be removed in #223.


kwcantrell:
After looking over #293, we will get some conflicts in the tests. It probably would be easier if we merged #293 first.

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F301%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F301%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F301%2Fempress-tree.qzv)

fedarko:
Haven't gone over the code in detail but it looks like node selection (either internal or tip) is broken -- getting an error that

```
select-node-menu.js:421 Uncaught TypeError: emp.getNodeInf is not a function
    at SelectedNodeMenu.setSelectedNodes (select-node-menu.js:421)
    at openMenu (canvas-events.js:396)
    at CanvasEvents.placeNodeSelectionMenu (canvas-events.js:404)
    at HTMLCanvasElement.mouseClick (canvas-events.js:198)
```

antgonza:
@fedarko, how did you catch that error?

fedarko:
@antgonza I opened up the mchelper empress-tree qzv and tried clicking on a tip, and noticed that it didn't work!

So ... Penguin McHelper is already paying off :)

kwcantrell:
thanks @fedarko. I typed getNodeInf instead of getNodeInfo...forgot a letter :)

empress/support_files/js/empress.js
```diff
@@ -94,21 +95,43 @@ define([
...
+         *       not assumed to be unique. Use nameToKeys to convert a name to
+         *       list of indicies associated with it. indicies start at 1
          * @private
          */
         this._treeData = treeData;
 
+        /**
+         * @type {Object}
+         */
+        this._td_to_ind = td_to_ind;
```
fedarko, line 116: I don't really mind that much, but would it be possible to use [camel case](https://en.wikipedia.org/wiki/Camel_case) for this variable? (e.g. `this._tdToInd` instead of `this._td_to_ind`). Main motivation is consistency with the rest of the JS code.

...that said obviously this isn't important and I can spot a handful of other places throughout the JS code which don't use camel case, so it might make more sense to defer this to an issue

empress/support_files/js/empress.js
```diff
@@ -94,21 +95,43 @@ define([
...
+        // add additional info to treeData
+        // Note: this information is not added in core.py because the values
+        //       are initialized the same.
+        for (var key in this._td_to_ind) {
+            this._td_to_ind[key] += 4;
+        }
+        this._td_to_ind.color = 0;
+        this._td_to_ind.isColored = 1;
+        this._td_to_ind.visible = 2;
+        this._td_to_ind.inSample = 3;
```
fedarko, line 127: Could we delete the `inSample` property entirely? I don't think we're using it for anything anymore.

empress/support_files/js/empress.js
```diff
@@ -94,21 +95,43 @@ define([
...
+                this._td_to_ind.color,
+                0,
+                this.DEFAULT_COLOR
+            );
+            this._treeData[i].splice(this._td_to_ind.isColored, 0, false);
+            this._treeData[i].splice(this._td_to_ind.visible, 0, true);
+            this._treeData[i].splice(this._td_to_ind.inSample, 0, false);
             if (this._tree.isleaf(this._tree.postorderselect(i))) {
                 this._numTips++;
             }
```
fedarko, line 143: I know you didn't add this code in during this PR :), but it looks like `this._numTips` isn't used anywhere. Would you mind removing the two references to it in this file?

empress/support_files/js/empress.js
```diff
@@ -262,6 +285,17 @@ define([
         this.centerLayoutAvgPoint();
     };
 
+    Empress.prototype.getNodeInfo = function (node, attr) {
```
fedarko, line 318: Would you mind adding documentation to this and `setNodeInfo()`? Doesn't have to be very detailed, but it would be nice to mention (among other things) what `node` is, that `attr` is the name of an attribute (not its index), and that this can return `undefined` if `attr` is invalid. (Since I imagine we'll be using these functions heavily in the future!)

empress/support_files/js/empress.js
```diff
@@ -527,60 +561,12 @@ define([
 
     Empress.prototype.getX = function (nodeObj) {
         var xname = "x" + this._layoutToCoordSuffix[this._currentLayout];
-        return nodeObj[xname];
+        return nodeObj[this._td_to_ind[xname]];
```
fedarko, line 594: Could this be replaced with `getNodeInfo(nodeObj, xname)`? (Would make changing things in the future easier.)

empress/support_files/js/empress.js
```diff
@@ -527,60 +561,12 @@ define([
...
-        } else {
-            // the root is not drawn in the unrooted layout
-            numLines = this._tree.size - 1;
-        }
-
-        // Every line takes up two coordinates, and every coordinate takes up
-        // VERTEX_SIZE (as of writing this is set to 5, for x, y, r, g, b)
-        // spaces in coords.
-        return 2 * numLines * this._drawer.VERTEX_SIZE;
+        return nodeObj[this._td_to_ind[yname]];
```
fedarko, line 599: see comment above re: using `getNodeInfo()` here

empress/support_files/js/empress.js
```diff
@@ -620,13 +606,17 @@ define([
         var tree = this._tree;
 
         // The coordinates (and colors) of the tree's nodes.
-        var coords = new Float32Array(this.computeNecessaryCoordsSize());
+        var coords = [];
```
fedarko, line 639: This makes a lot more sense XD

tests/python/test_core.py
```diff
@@ -29,11 +29,11 @@
 class TestCore(unittest.TestCase):
 
     def assert_almost_equal_tree_data(self, tree_data, exp):
-        for node in tree_data:
-            for attr in tree_data[node]:
+        # node skip treeData[0] because currently nodes are stored in postorder
```
fedarko, line 32:
```suggestion
        # skip treeData[0] because currently nodes are stored in postorder
```

empress/support_files/js/empress.js
kwcantrell: In python, it is conventional to separate variables using underscores (I believe its called snake case?) instead of camel case.  Most of the python code uses snake case so, if we want to make things consistent with the JS code then it would be better to open an issue. @ElDeveloper do you have any opinions on this?

empress/core.py
```diff
@@ -264,38 +264,75 @@ def _to_dict(self):
         # TODO: figure out implications of screen size
         layout_to_coordsuffix, default_layout = self.tree.coords(4020, 4020)
 
-        tree_data = {}
+        # store node data in a postorder fashion.
+        # Note: currently, bp-tree uses 1-based index so the first element is 1
+        #       this will be fixed in #223
+        tree_data = [0] * (self.tree.count() + 1)
```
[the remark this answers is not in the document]
kwcantrell: Modified to use append. It do make the code simpler. Although I'm not to sure about a performance boost.

empress/core.py
```diff
@@ -264,38 +264,75 @@ def _to_dict(self):
...
+            "name": 0,
+            "x2": 1,
+            "y2": 2,
+            "xr": 3,
+            "yr": 4,
+            "xc1": 5,
+            "yc1": 6,
+            "xc0": 7,
+            "yc0": 8,
+            "angle": 9,
```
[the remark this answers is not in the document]
kwcantrell: After using append, the loop is no longer dependent on angle being the last index in the all-nodes-specific field.

empress/support_files/js/empress.js
kwcantrell: We should be able to remove it. I believe at some point we did remove inSample. Not too sure how it crept back in.

empress/support_files/js/empress.js
kwcantrell: Removed

empress/support_files/js/empress.js
kwcantrell: Thanks, I completely forgot.

empress/support_files/js/empress.js
fedarko: I'd assumed how we were doing things was that we were using snake case for python and camel case for JS, but we could totally do things consistently one way or another -- I prefer using separate case for separate languages but I don't really mind either way :) Agreed -- @ElDeveloper, please let us know what you recommend.

empress/support_files/js/empress.js
kwcantrell: Chnaged

empress/support_files/js/empress.js
kwcantrell: changed

empress/support_files/js/empress.js
```diff
@@ -292,6 +315,17 @@ define([
...
     };
 
+    Empress.prototype.getNodeInfo = function (node, attr) {
+        return node[this._td_to_ind[attr]];
+    };
+
+    Empress.prototype.setNodeInfo = function (node, attr, value) {
+        if (this.getNodeInfo(node, attr) === undefined) {
+            throw "Invalid attribute.";
+        }
```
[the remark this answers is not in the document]
kwcantrell: Agree, I'll add a note in the docstring that lets user know it will not perform any validation on its input parameters.

empress/support_files/templates/empress-template.html
```diff
@@ -133,7 +133,8 @@ <h3 class="hidden" id="menu-sm-header">Sample Presence Information</h3>
         );
         var empress = new Empress(
             tree,
-            {{ tree_data | tojson }},
+            {{ tree_data }},
+            {{ td_to_ind }},
```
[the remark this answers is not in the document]
kwcantrell: added

empress/support_files/js/empress.js
ElDeveloper: Part of the confusion here is likely due to the fact that it's a Python-born variable that percolates to JS so the translation needs to happen at one point or another, right? Hence I would suggest that the JSON object that's serialized includes the snake case but that the constructor for `Empress` uses camelcase.

empress/support_files/js/empress.js
```diff
@@ -292,6 +310,36 @@ define([
         this.centerLayoutAvgPoint();
     };
 
+    /**
+     * Retrive an attribute from a node.
+     *
+     * @param{Array} node An array that holds all the attributes for a given
+     *                     node. Note: this is an entry in this._treeData.
+     * @param{String} attr The attribute to retive from node.
```
fedarko, line 318:
```suggestion
     * @param{String} attr The attribute to retrieve from the node.
```

empress/support_files/js/empress.js
```diff
@@ -292,6 +310,36 @@ define([
...
+     *
+     * @return The attribute, If attr is not a valid attribute of node, then
```
fedarko, line 320:
```suggestion
     * @return The attribute; if attr is not a valid attribute of node, then
```

empress/support_files/js/empress.js
```diff
@@ -292,6 +310,36 @@ define([
...
+
+    /**
+     * Sets an attribute of a node.
+     *
+     * Note: this method does not perfom any kind of validation. It is assumed
+     *       that node, attr and value are all valid.
+     *
+     * @param{Array} node An array that holds all the attributes for a given
+     *                     node. Note: this is an entry in this._treeData.
+     * @param{String} attr The attribute to set in node.
```
fedarko, line 335:
```suggestion
     * @param{String} attr The attribute to set for the node.
```

empress/support_files/templates/empress-template.html
```diff
@@ -138,7 +138,8 @@ <h3 class="hidden" id="menu-sm-header">Sample Presence Information</h3>
         );
         var empress = new Empress(
             tree,
-            {{ tree_data | tojson }},
+            {{ tree_data }},
```
ElDeveloper, line 141: Can you also pass tree_data through `| tojson`? This is important to guarantee compatibility between types in Python and JavaScript.

empress/core.py
```diff
@@ -266,38 +266,66 @@ def _to_dict(self):
...
+        }
         names_to_keys = {}
         # Note: tree_data starts with index 1 because the bp tree uses 1 based
         # indexing
         for i, node in enumerate(self.tree.postorder(include_self=True), 1):
-            tree_data[i] = {
-                'name': node.name,
-            }
+            tree_data.append([])
+            tree_data[i].append(node.name)
```
fedarko, line 304: Oh - shoot, sorry, I think I wasn't clear in my first message. I meant only using append for adding the internal lists for the tree_data, but still using td_to_ind to add the node attributes to each list. I can totally see how I was unclear, though... would it be possible to revert the uses of append below this to what you had them as before? if not, no worries, I can take care of it later. Sorry for messing things up.

empress/support_files/js/empress.js
```diff
@@ -97,27 +98,44 @@ define([
...
+        // Note: this information is not added in core.py because the values
+        //       are initialized the same.
+        for (var key in this._tdToInd) {
+            this._tdToInd[key] += 3;
+        }
+        this._tdToInd.color = 0;
+        this._tdToInd.isColored = 1;
+        this._tdToInd.visible = 2;
+
         // count number of tips and set default color/visible
```
fedarko, line 128:
```suggestion
        // set default color/visible status for each node
```

fedarko:
thanks @kwcantrell! :100:

PR #309, opened 2020-08-06
Fix a bug with barplot length scaling; abstract and test length-scaling code

fedarko:
The bug in question occurs when any of the values in the feature metadata field used for length scaling are integer values represented as floats (e.g. `"0.0"`). The mapping between unique numeric feature metadata values and their corresponding lengths is contained in a mapping where, previously, the keys were passed through  `parseFloat()` first. The problem with this is that `parseFloat("0.0")` is `0`, and keys in Objects are treated as Strings -- so the entry for `0` was inaccessible to the nodes with a feature metadata value of `0.0`, bizarrely. (The code [assumes that a value being missing as a key from that mapping means that the value in question was non-numeric](https://github.com/biocore/empress/blob/3d22b50d116fb4317060f2d65f45e91a0cb59029/empress/support_files/js/empress.js#L1360-L1366)... but in this case it was because the Object was created incorrectly.)

Anyway, now the code stores the original string values as the keys in the Object, not the `parseFloat()`-ed values. (This is the same way the color-scaling code works -- the reason that that code doesn't suffer from this bug is likely because I was able to adapt it directly from Emperor, while the length-scaling code required some extra tweaking :) The problematic code in question has also been abstracted to a new utility function, which is now decently tested.

(This bug came up when testing visualizing feature importance scores as barplots -- the majority of importance scores in the `moving-pictures-classifier/feature_importance.qza` file in [this tutorial](https://docs.qiime2.org/2020.6/tutorials/sample-classifier/) are 0s, so I noticed that a lot of these were just missing from the barplot.)

### DIFF
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -1315,7 +1315,7 @@ define([
                 // TODO: verify that this handles negative values properly
                 // and/or support drawing negative values in the opposite
                 // direction as positive ones
-                fm2length[fn] =
+                fm2length[n] =
                     ((fn - valMin) / valRange) * lengthRange +
                     layer.scaleLengthByFMMin;
             });
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -1212,7 +1212,6 @@ define([
         var maxX = prevLayerMaxX;
         var fm2color, colorFMIdx;
         var fm2length, lengthFMIdx;
-        var msg;
         // Map feature metadata values to colors, if requested (i.e. if
         // layer.colorByFM is true). If not requested, we'll just use the
         // layer's default color.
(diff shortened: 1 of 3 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F309%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F309%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F309%2Fempress-tree.qzv)

kwcantrell:
Thanks @fedarko, this looks good to me. I think it should be good to merge.

PR #313, opened 2020-08-08
Speed up and test sample metadata barplot computations

### LINKED ISSUE #298: Optimize sample metadata barplot computation
Since it takes ~10-15 sec to compute the barplots for the HMP dataset tree. The bottleneck here is likely the fact that every tip requires accessing the BIOM table to figure out how many samples of what types contain this tip, etc.

fedarko:
Closes #298.

On the HMP test dataset we've been working with, this went from taking ~10 seconds to draw a sample metadata barplot layer to less than 1 second. The code to compute sample metadata group "frequencies" was modified so that, rather than working on a feature-by-feature basis, all features' frequencies are now computed in a single pass through the BIOM table. Based on discussion with @ElDeveloper about ways to speed this up :)

The code to compute the "frequency map" is now stored in the `BiomTable` class and unit-tested, also.


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F313%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F313%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F313%2Fempress-tree.qzv)

empress/support_files/js/biom-table.js
```diff
@@ -472,13 +472,102 @@ define(["underscore", "util"], function (_, util) {
...
+        // a 2D array (which is how fIdx2counts has been stored)
+        var fID2freqs = {};
+        var totalSampleCount;
+        _.each(this._fIDs, function (fID, fIdx) {
+            totalSampleCount = fIdx2sampleCt[fIdx];
+            fID2freqs[fID] = {};
+            _.each(fIdx2counts[fIdx], function (count, smValIdx) {
+                if (count > 0) {
+                    fID2freqs[fID][uniqueSMVals[smValIdx]] =
+                        count / totalSampleCount;
```
kwcantrell, line 565: We should either let the user know that the sample stack bar plots are displaying proportions instead of counts or we should just display the counts (or all the user to choose).

empress/support_files/js/biom-table.js
```diff
@@ -472,13 +472,102 @@ define(["underscore", "util"], function (_, util) {
...
+        _.each(this._tbl, function (presentFeatureIndices, sIdx) {
+            // Figure out what metadata value this sample has at the column.
+            cVal = scope._sm[sIdx][colIdx];
+            cValIdx = smVal2Idx[cVal];
+            // Increment s.m. value counts for each feature present in this
+            // sample
+            _.each(presentFeatureIndices, function (fIdx) {
+                fIdx2counts[fIdx][cValIdx]++;
+                fIdx2sampleCt[fIdx]++;
+            });
```
kwcantrell, line 550:
```suggestion
        _.each(this._tbl, function (samples, sIdx) {
            // Figure out what metadata value this sample has at the column.
            cVal = scope._sm[sIdx][colIdx];
            cValIdx = smVal2Idx[cVal];
            // Increment s.m. value counts for each feature present in this
            // sample
            _.each(samples, function (fIdx) {
                fIdx2counts[fIdx][cValIdx]++;
                fIdx2sampleCt[fIdx]++;
            });
```

Since _tbl is a 2D array of samples by features, this may make more sense to developers who are not to familiar with biom-table. Minor suggestion feel free to ignore. (:

empress/support_files/js/biom-table.js
```diff
@@ -472,13 +472,102 @@ define(["underscore", "util"], function (_, util) {
...
+        });
+
+        // Convert counts to frequencies
+        // Also, return an Object where the keys are feature IDs pointing to
+        // other Objects where the keys are sample metadata values, rather than
+        // a 2D array (which is how fIdx2counts has been stored)
+        var fID2freqs = {};
```
kwcantrell, line 557: This is just a note: It might be less memory expensive/more efficient to iterate through it this was 2D array especially for large trees. Storing (even temporarily) an object that uses fId's as keys can be quite memory intensive + js has a bit more (although not a ton) of overhead compared to arrays when iterating. I would keep this the same for now and we can come back to this if memory becomes a problem when working with large trees.

empress/support_files/js/empress.js
```diff
@@ -1139,50 +1139,51 @@ define([
...
                 var name = this.getNodeInfo(node, "name");
+                var freqs = feature2freqs[name];
                 // Don't draw bars for tips that aren't in the BIOM table
                 // (Note that this is only for the sample metadata barplots --
                 // these tips could still ostensibly have associated
                 // feature metadata)
-                if (this._biom.getObsIDsDifference([name]).length > 0) {
+                if (_.isUndefined(freqs)) {
                     continue;
                 }
```
kwcantrell, line 1157: We should be able to just iterate through feature2freqs instead of the whole tree. It would save a bit of computation cycles since we will not have to call isleaf on all nodes of the tree (which can be a lot on large trees) and verify that the feature is in biom-table. It might require a bit of refactoring so I wouldn't worry about it for this PR.

empress/support_files/js/biom-table.js
fedarko: That's a good point -- and supporting displaying "variable-length" stacked barplots (where e.g. 1 sample has a given fixed length, so a tip present in 2 samples vs. a tip present in 20 samples will display differently) would be a cool feature to add. I've added an issue for this at #322, and updated the README to be clearer about proportions being displayed.

empress/support_files/js/biom-table.js
fedarko: If it's ok I'd like to keep this as is -- there are a couple of other places in the BIOM table JS where `presentFeatureIndices` is used when iterating over the table in this way. I added some context to the comment before this loop, so hopefully that makes things clearer as a compromise :) You're totally correct, though -- we should make this code as non-intimidating as possible, since these functions are all getting used pretty frequently ...

empress/support_files/js/empress.js
fedarko: This is an **awesome** catch, and I think it would add in a nice speedup -- I'll refactor things and see how it goes :) Thanks so much!

empress/support_files/js/biom-table.js
fedarko: Understood and agreed. Added a TODO to the code detailing this optimization.

One weird-ish thing is that none of the Empress JS code besides the BIOM table knows anything about feature indices -- it only stores information about feature names (aka IDs), as far as I know. Moving the index stuff to empress.js / treeData might be a hassle, but I think it could save a decent amount of space -- I guess in-the-BIOM-table tip names are technically stored in two places (treeData and the BIOM table).

empress/support_files/js/empress.js
fedarko: Change has been made, thanks!

fedarko:
@kwcantrell things should be ready to merge now, I think. Thanks!

fedarko:
... ok, nvm, now they should be good (one space angered prettier :sweat_smile:)

kwcantrell:
Thanks @fedarko

Issue #195, opened 2020-06-22
Interaction bugs between animation and sample / feature metadata coloring


fedarko:
Steps to reproduce:

1. Open the animation tab and start some sort of animation
2. Go to the sample or feature metadata coloring tab (without stopping the animation), and (if not already open) press the checkbox to open up the coloring options.
3. Change an option (if needed) and press "Update" in the sample or feature metadata coloring tab
4. The plot will now be updated to use the requested metadata coloring option
5. ... However, a few seconds later, the animation will continue (since we never stopped it), undoing the metadata coloring methods while still leaving them "checked."

You kind of have to try to break Empress to get this to happen, but it's still confusing -- so we should address this if we can IMO.

fedarko:
This will also be a problem with synchronized animations (#319). We can probably get around this by adding a `disableColoringUI()` function or something that disables all of the "breaking" Empress options while animations are happening, and an `enableColoringUI()` function that does the reverse (and is called when an animation is finished / restarted / etc.).

fedarko:
This was closed by #505. Thanks @kwcantrell! :tada: :tada: :tada:

PR #319, opened 2020-08-11
ENH: Add synchronized animations

ElDeveloper:
In an Empire plot, users can start an animation in Emperor, and the tree will
update synchronously with the trajectories. Only points in the gradient are
visualized in the tree.


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F319%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F319%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F319%2Fempress-tree.qzv)

docs/moving-pictures/sample_metadata.tsv
```diff
@@ -1,36 +1,36 @@
-sample-id	barcode-sequence	body-site	year	month	day	subject	reported-antibiotic-usage	days-since-experiment-start
```
kwcantrell, line 1: Was there a particular reason of separating samples into s1-* and s2-*? or was the goal to just create more trajectories?

empress/support_files/js/animator.js
```diff
@@ -241,6 +245,14 @@ define(["Colorer", "util"], function (Colorer, util) {
         this.empress.drawTree();
     };
 
+    Animator.prototype.showAnimationFrameAtIndex = function (index) {
```
kwcantrell, line 248: Can you add doc string

docs/moving-pictures/sample_metadata.tsv
ElDeveloper: Reason is to create meaningful trajectories in Emperor since those can only track one site at a time.

ElDeveloper:
Depends on: https://github.com/biocore/emperor/pull/780

empress/support_files/js/animator.js
```diff
@@ -241,6 +243,27 @@ define(["Colorer", "util"], function (Colorer, util) {
...
+     * Draw the tree corresponding a given frame.
+     *
+     * This method is used in empire plots, and won't draw a legend since the
+     * color values are visible in Emperor.
+     *
+     * @param{Number} frame The frame to retrieve. frame must be in the range
+     *                      [0, this.totalFrames] else an error will be thrown.
+     */
+    Animator.prototype.showAnimationFrameAtIndex = function (frame) {
+        if (frame < 0 || frame > this.totalFrames) throw "Invalid Frame";
```
fedarko, line 256: Should the upper limit here be `this.totalFrames - 1` (making the acceptable range `[0, this.totalFrames)`)? Since the frames are 0-indexed and since `gradientSteps` is an array with length equal to `this.totalFrames`.

We might want to defer clarifying these details to another issue, though.

empress/support_files/js/node-click-callback.js
```diff
@@ -22,3 +22,30 @@ ec.sceneViews[0].on("click", function (name, object) {
...
 });
+
+// synchronize animations in Empress
+ec.controllers.animations.addEventListener("animation-started", function (
+    payload
+) {
+    animator.setAnimationParameters(
+        payload.message.trajectory,
+        payload.message.gradient,
+        "discrete-coloring-qiime",
```
fedarko, line 33: Is it possible to pass the current color map name from Emperor instead?

empress/support_files/js/node-click-callback.js
```diff
@@ -22,3 +22,30 @@ ec.sceneViews[0].on("click", function (name, object) {
 
     empress.showNodeMenuForName(name);
 });
+
+// synchronize animations in Empress
+ec.controllers.animations.addEventListener("animation-started", function (
```
fedarko, line 27: Could you add a brief comment about what `payload` is in these three event listeners? I can infer that it's probably an `Object`, but more information would be nice (esp for people looking at this code who might not be familiar with Emperor's internals).

empress/support_files/js/animator.js
ElDeveloper: Good catch, just updated this!

empress/support_files/js/node-click-callback.js
ElDeveloper: Unfortunately colormaps are not selectable for animations it defaults to QIIME's. What I think would be best is to instead be able to pass the colors that are currently being used in Emperor's animations tab. However I don't think the Colorer class in Empress would support this.

If you have a few mins, maybe we can chat and pair-program this.

empress/support_files/js/node-click-callback.js
ElDeveloper: Added some docs and a link to even more docs!

empress/support_files/js/node-click-callback.js
fedarko: Sorry for being late on this. Yeah, it sounds like passing the colors would work out -- the Colorer class doesn't support this right now, IIRC, but the class is pretty simple under the hood so modifying this would be possible. Maybe we can sort through this sometime tomorrow if you're available?

ElDeveloper: I finally got a chance to look into this, not too complicated - the Colorer class is pretty flexible. Let me know what you think of the changes/interface.

fedarko (changes requested):
It's great that the colors are shared now! I spent some time playing around with this today -- it's super cool. There are a few issues that came up; detailing below. (I installed the master branch of Emperor alongside this PR version of Empress, since I think McHelper's Emperor version is a bit behind?)

### Problem 1

When certain fields are selected as the `trajectory` / `gradient` in Emperor, I see an `Uncaught TypeError: Cannot read property 'value' of undefined` error in the console:
barcseq png

With `days-since-experiment-start` as the gradient field, the following trajectory fields cause this error for the MP dataset:

1. `SampleID` (this one makes sense, since I don't think Empress considers sample ID to be a metadata field. We could probably add a special case or something that warns the user when they pick this for an animation.)
2. `barcode-sequence`
3. `month` (no idea why `year` is ok but this isn't)
4. `days-since-experiment-start`

### Problem 2

I sometimes get an error in the console that `THREE.DirectGeometry: Faceless geometries are not supported.` I can only seem to reproduce this with `days-since-experiment-start` as the gradient when `trajectory` is the trajectory, bizarrely:

traj

It's strange -- as you can see by the number next to the error message, a lot of these errors are thrown in succession. Each time I restart the animation (using this exact gradient / trajectory / speed) I get exactly 33 of these errors. Changing the speed seems to change the number of errors I get (???). No idea what's going on here.

### Problem 3

It's possible to start an Empress animation while an Emperor animation is also running. This results in the tree sort of switching back and forth between the two?

uhoh

Sort of similarly to #195, we should probably disable the Empress animation panel during synchronized animations. I'm ok with deferring this to a future issue if you'd prefer to get this PR in sooner rather than later.

### Question

Not all that important, but Empress animations (when finished) show a toast message saying `Animation complete`. Do you think synchronized animations should also show this on finishing? I don't really mind either way, but wanted to run it by you.

ElDeveloper:
Thanks @fedarko!

> ### Problem 1
> When certain fields are selected as the `trajectory` / `gradient` in Emperor, I see an `Uncaught TypeError: Cannot read property 'value' of undefined` error in the console:

This is an Emperor feature/bug if you see the traceback this originates from the `animations-controller.js`. This happens when only single-sample trajectories are selected so there's nothing to animate. This should be better handled on the Emperor side.

> ### Problem 2
> I sometimes get an error in the console that `THREE.DirectGeometry: Faceless geometries are not supported.` I can only seem to reproduce this with `days-since-experiment-start` as the gradient when `trajectory` is the trajectory, bizarrely:

This is an issue with THREE.js the most recent build that emperor uses raises this error, with no "visible" consequences. I haven't been able to figure out what's going on here. I think we just need to refactor the way Emperor handles `noop` cycles for asynchronous trajectories.
 
> ### Problem 3
> It's possible to start an Empress animation while an Emperor animation is also running. This results in the tree sort of switching back and forth between the two?

I've now created a method to disable and reset UI elements in the animation panel. Note that I had to move `__startOptions` to be a public method `startOptions` if needed I can do the same thing for other methods, but I think for now we only need this one.

> ### Question
> Not all that important, but Empress animations (when finished) show a toast message saying `Animation complete`. Do you think synchronized animations should also show this on finishing? I don't really mind either way, but wanted to run it by you.

I just added this as well. Luckily there's an `animation-ended` event so I added this call there.

empress/support_files/js/animation-panel-handler.js
```diff
@@ -101,6 +99,29 @@ define(["Colorer", "util"], function (Colorer, util) {
...
+
+        this.collapseChk.disabled = !enabled;
+        this.lWidth.disabled = !enabled;
+
+        this.pauseBtn.disabled = !enabled;
+        this.startBtn.disabled = !enabled;
+        this.stopBtn.disabled = !enabled;
+        this.resumeBtn.disabled = !enabled;
+        this.prevFrameBtn.disabled = !enabled;
+        this.nextFrameBtn.disabled = !enabled;
```
fedarko, line 121: I doubt that doing a bunch of negations will be a problem, but may as well try to make fast :)

```suggestion
        var isDisabled = !enabled;
        this.colorSelect.disabled = isDisabled;
        this.gradient.disabled = isDisabled;
        this.trajectory.disabled = isDisabled;

        this.collapseChk.disabled = isDisabled;
        this.lWidth.disabled = isDisabled;

        this.pauseBtn.disabled = isDisabled;
        this.startBtn.disabled = isDisabled;
        this.stopBtn.disabled = isDisabled;
        this.resumeBtn.disabled = isDisabled;
        this.prevFrameBtn.disabled = isDisabled;
        this.nextFrameBtn.disabled = isDisabled;
```

fedarko (approved):
Thanks, @ElDeveloper! Looks good to me. For reference I opened two issues about problems 1 and 2 in Emperor's repo, just to keep track of them. (https://github.com/biocore/emperor/issues/784, https://github.com/biocore/emperor/issues/785)

Two remaining points, but I'm ok to merge this code in with or without them:

1. Minor suggestion with the enabling / disabling code (see below, or above I forget which way github orders this stuff)

2. It looks like it's possible to start an animation in Empress and then start an animation in Emperor (and both collide). Sorry for not anticipating this in the earlier review. Would it be possible to extend the disabling stuff so that Empress animations temporarily disable Emperor animations (so it works the other way also)? I can imagine that being more work than it's worth (and it is such a silly corner case that I doubt anyone not actively trying to break things will run into it), so I'm ok deferring that to an issue.

ElDeveloper:
@fedarko, both notes should be addressed. Let me know what you think.

tests/test-animation-panel-handler.js
```diff
@@ -118,7 +118,35 @@ require([
         },
     });
 
-    test("startOptions", function () {
+    test("setOnAnimationStarted", function (assert) {
+        assert.expect(2);
```
fedarko, line 122: (just making this clear for folks unfamiliar with QUnit, aka me)
```suggestion
        // Exactly two assertions should be run within this test
        assert.expect(2);
```

fedarko (changes requested):
Thanks @ElDeveloper. I found a few issues:

1. It looks like synchronized animations don't work any more:

animbroke

2. Emperor animation UI stuff is are now disabled during Empress animations, but Empress animation UI stuff is still enabled when Emperor animations are running.

_Emperor animation is correctly disabled when animating in Empress:_

animgood

_Empress animation is still enabled when animating in Emperor, though:_
enablbroke

3. Every Empress animation seems to now start on its second, rather than its first, frame. Guessing there is an off-by-one error somewhere, but not sure what exactly caused this. I verified that this is not a problem on the current master version of Empress, so it seems like this is getting introduced in this PR.

framebroke

I am not sure where some of these are coming from. Would be free to try to pair-debug this on call later today / this week.

ElDeveloper:
> 1. It looks like synchronized animations don't work any more:

Not sure how this happened, but I just resolved conflicts and can't see the problem.

> 1. Emperor animation UI stuff is are now disabled during Empress animations, but Empress animation UI stuff is still enabled when Emperor animations are running.
> 

I can't reproduce this problem (may have to do with the resolved conflicts).


> 1. Every Empress animation seems to now start on its second, rather than its first, frame. Guessing there is an off-by-one error somewhere, but not sure what exactly caused this. I verified that this is not a problem on the current master version of Empress, so it seems like this is getting introduced in this PR.
> 

I found the reason for this, should be fixed.

fedarko (approved):
Thanks @ElDeveloper! Everything looks good now. (For reference, it looks like problems 1 and 2 were due to me using the `empire-biplot` version on McHelper -- pulling down the changes locally worked.)

I have one small observation; during Empress-started animations the `Collapse Clades` and `Extra Line Width` elements are left enabled, while this PR disables them during Emperor-started animations as well. It might be nice to make this consistent (either disabling them during Empress-started animations, or leaving them enabled during Emperor-started animations); however, I'm ok delaying this to later if you'd like. This code should be good to go now :+1:

fedarko:
Thanks @ElDeveloper, looks good to me! Will merge in when the build passes.

ElDeveloper:
Thanks @fedarko

PR #323, opened 2020-08-11
Fixes #317

### LINKED ISSUE #317: Use the biom.Table throughout the codebase
This will be more memory efficient than Pandas.

wasade:
Fixes #317 and uses `biom.Table` rather than `pd.DataFrame` for the feature table

### DIFF
--- a/empress/compression_utils.py
+++ b/empress/compression_utils.py
@@ -5,7 +5,6 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
-import numpy as np
 
 
 def remove_empty_samples_and_features(table, sample_metadata, ordination=None):
@@ -130,22 +129,14 @@ def compress_table(table):
     """
     feature_ids = table.ids(axis='observation')
     sample_ids = table.ids()
-    feature_ids_indexed = np.arange(len(feature_ids))
-    sample_ids_indexed = np.arange(len(sample_ids))
 
-    f_ids_to_indices = {fid: int(idx) for fid, idx in zip(feature_ids,
-                                                          feature_ids_indexed)}
-    s_ids_to_indices = {sid: int(idx) for sid, idx in zip(sample_ids,
-                                                          sample_ids_indexed)}
+    f_ids_to_indices = {fid: idx for idx, fid in enumerate(feature_ids)}
+    s_ids_to_indices = {sid: idx for idx, sid in enumerate(sample_ids)}
 
     compressed_table = []
-    for vec in table.iter_data(dense=False):
+    for vec in table.iter_data(axis='sample', dense=False):
         # vec is a row from a scipy.sparse.csr_matrix
-        compressed_table.append(list(feature_ids_indexed[vec.indices]))
-
-    # the values from compressed_table are np.int64 types, which the json
-    # encoder by default does not know how to handle.
-    compressed_table = [[int(v) for v in row] for row in compressed_table]
+        compressed_table.append([int(i) for i in vec.indices])
 
     return (
         list(sample_ids), list(feature_ids), s_ids_to_indices,
--- a/empress/compression_utils.py
+++ b/empress/compression_utils.py
@@ -5,6 +5,7 @@
 #
 # The full license is in the file LICENSE, distributed with this software.
 # ----------------------------------------------------------------------------
+import numpy as np
 
 
 def remove_empty_samples_and_features(table, sample_metadata, ordination=None):
@@ -16,9 +17,8 @@ def remove_empty_samples_and_features(table, sample_metadata, ordination=None):
 
     Parameters
     ----------
-    table: pd.DataFrame
-        Representation of a feature table. The index should describe feature
-        IDs; the columns should describe sample IDs.
+    table: biom.Table
+        Representation of a feature table.
     sample_metadata: pd.DataFrame
         Sample metadata. The index should describe sample IDs, and the columns
         should describe sample metadata fields (e.g. "body site").
@@ -31,7 +31,7 @@ def remove_empty_samples_and_features(table, sample_metadata, ordination=None):
 
     Returns
     -------
-    filtered_table: pd.DataFrame
+    filtered_table: biom.Table
         Copy of the input feature table with empty samples and features
         removed.
     filtered_sample_metadata: pd.DataFrame
@@ -48,24 +48,18 @@ def remove_empty_samples_and_features(table, sample_metadata, ordination=None):
     ----------
         - Adapted from qurro._df_utils.remove_empty_samples_and_features().
     """
-    orig_tbl_samples = set(table.columns)
-    orig_tbl_features = set(table.index)
+    orig_tbl_samples = set(table.ids())
+    orig_tbl_features = set(table.ids(axis='observation'))
 
-    # (In Qurro, I used (table != 0) for this, but (table > 0) should also
-    # work. Should be able to assume a table won't have negative abundances.)
-    # This approach based on https://stackoverflow.com/a/21165116/10730311.
-    gt_0 = (table > 0)
-    filtered_table = table.loc[
-        gt_0.any(axis="columns"), gt_0.any(axis="index")
-    ]
-    if filtered_table.empty:
+    filtered_table = table.remove_empty(inplace=False)
+    if filtered_table.is_empty():
         raise ValueError("All samples / features in matched table are empty.")
 
     # Let user know about which samples/features may have been dropped, if any.
     # Also, if we dropped any empty samples, update the sample metadata.
     filtered_sample_metadata = sample_metadata
 
-    sample_diff = orig_tbl_samples - set(filtered_table.columns)
+    sample_diff = orig_tbl_samples - set(filtered_table.ids())
     if sample_diff:
         if ordination is not None:
             empty_samples_in_ord = sample_diff & set(ordination.samples.index)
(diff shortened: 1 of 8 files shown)


empress/compression_utils.py
```diff
@@ -134,43 +128,28 @@ def compress_table(table):
...
-    sample_ids = list(binarized_table.columns)
-    s_ids_to_indices = {sid: idx for idx, sid in enumerate(sample_ids)}
-
-    indexed_b_table = binarized_table.rename(
-        index=f_ids_to_indices, columns=s_ids_to_indices
-    )
+    f_ids_to_indices = {fid: int(idx) for fid, idx in zip(feature_ids,
+                                                          feature_ids_indexed)}
+    s_ids_to_indices = {sid: int(idx) for sid, idx in zip(sample_ids,
+                                                          sample_ids_indexed)}
```
fedarko, line 139: Could the original versions of these lines (with the calls to `enumerate()`, and without declaring `feature_ids_indexed` / `sample_ids_indexed`) be kept here? If these versions are more efficient that's cool, but it doesn't seem that way to me

empress/compression_utils.py
```diff
@@ -134,43 +128,28 @@ def compress_table(table):
...
 
     compressed_table = []
+    for vec in table.iter_data(dense=False):
```
fedarko, line 142: Suggest naming the variable `sample_vec` or something similar (and/or adding an explicit `axis="sample"` to the `iter_data()` call). It isn't immediately clear whether `iter_data()` is iterating over features or samples (I mean _now_ I know it's samples because I just spent a few minutes reading through this, but you get it :)

empress/compression_utils.py
```diff
@@ -134,43 +128,28 @@ def compress_table(table):
...
+        # vec is a row from a scipy.sparse.csr_matrix
```
fedarko, line 143: I followed the example on [`iter_data()`'s docs](http://biom-format.org/documentation/generated/biom.table.Table.iter_data.html) and it looks like it's actually a column? Sorry if I'm missing something.

image

empress/compression_utils.py
```diff
@@ -134,43 +128,28 @@ def compress_table(table):
...
+        compressed_table.append(list(feature_ids_indexed[vec.indices]))
```
fedarko, line 144: If I understand this, `feature_ids_indexed` is a numpy ndarray where the first element is 0, the second is 1, and so on (and the final element is `# features - 1`). It seems like subsetting this ndarray using `vec.indices` doesn't do anything, because the 0th element of `feature_ids_indexed` will always be 0, the 5th element will always be 5, etc.

So I *think* we could just replace the interior of the `list()` call with just `vec.indices`. But if that is a horrific misunderstanding of the situation, feel free to disregard this :)

fedarko: Other question: could we replace the `list()` call entirely with `[int(v) for v in vec.indices]`? I think this would avoid the need to do the int64 -> int conversions later on below, with the added bonus of avoiding an extra traversal through the compressed table.

empress/core.py
```diff
@@ -163,7 +163,7 @@ def _validate_and_match_data(self, ignore_missing_samples,
         # sending it to tools.match_inputs() and keep using the transposed
         # table for the rest of this visualizer.
         self.table, self.samples, self.tip_md, self.int_md = match_inputs(
-            self.tree, self.table.T, self.samples, self.features,
+            self.tree, self.table, self.samples, self.features,
```
fedarko, line 166: Could you remove the comment above this call? (Since we're no longer transposing the table.)

empress/tools.py
```diff
@@ -239,7 +238,7 @@ def match_inputs(
         )
 
     padded_metadata = sample_metadata.copy()
-    if len(sm_and_table_samples) < len(ff_table.columns):
+    if len(sm_and_table_samples) < len(ff_table.ids()):
```
fedarko, line 241:
```suggestion
    if len(sm_and_table_samples) < len(table_samples):
```

Avoids recomputing `ff_table.ids()`. I guess this does technically assume that the sample IDs in the table are unique (otherwise the length of the set will be different from the length of `ff_table.ids()`), but I have to believe that that's already validated somewhere in biom...?

empress/tools.py
```diff
@@ -278,7 +277,7 @@ def match_inputs(
     #
     # All that's left to do is to filter the sample metadata to just the
     # samples that are also present in the table.
-    sf_sample_metadata = sample_metadata.loc[ff_table.columns]
+    sf_sample_metadata = sample_metadata.loc[ff_table.ids()]
```
fedarko, line 280:
```suggestion
    sf_sample_metadata = sample_metadata.loc[table_samples]
```

(I think this should work, but I guess it might alter the order of samples in the metadata somewhat. Feel free to ignore this.)

tests/python/test_compression_utils.py
```diff
@@ -322,8 +330,8 @@ def test_compress_table_basic(self):
 
         # First off, verify that compress_table() leaves the original table DF
         # untouched.
-        assert_frame_equal(table_copy, self.table_ef)
-
+        self.assertEqual(table_copy, self.table_ef)
+        print(table_copy)
```
fedarko, line 334:
```suggestion
```

tests/python/test_core.py
```diff
@@ -42,32 +43,28 @@ def setUp(self):
...
-        self.unrelated_table = pd.DataFrame(
-            {
-                "Sample1": [5, 2, 0, 2],
-                "Sample2": [2, 3, 0, 1],
-                "Sample3": [5, 2, 0, 0],
-                "Sample4": [4, 5, 0, 4]
-            },
-            index=["h", "i", "j", "k"]
-        ).T
+        self.table = biom.Table(np.array([[1, 2, 0, 4],
```
fedarko, line 46: Comment line immediately above this can probs be removed, since the table is no longer a DF

empress/compression_utils.py
wasade: This avoids recomputing the `_indexed` variables, and having the `_indexed` variables be of `np.ndarray` allows for fast indexing in the forloop

empress/compression_utils.py
wasade: TBH, I thought the original code was doing this over features... :) 

Can add

empress/compression_utils.py
wasade: Not sure I follow, these are all vectors?

empress/compression_utils.py
wasade: Good catch, I think this is correct. Will take a stab

empress/core.py
wasade: kk

empress/tools.py
wasade: Duplicate IDs are not allowed

tests/python/test_compression_utils.py
wasade: Good catch

tests/python/test_core.py
wasade: cool cool

fedarko:
It looks like the tests are failing because a few "test tables" are still being created as DataFrames:

- `self.table` in `tests/python/test_tools.py`
- `table` in `tests/python/make-dev-page.py` (looks like this can be fixed in one line, by replacing line 67 with `table = table.view(biom.Table)` (and importing `biom` ofc))

wasade:
Thanks, @fedarko! I thought I had issued a full `pytest` call before the PR but had only tested out two of the modules. More shortly.

empress/tools.py
fedarko: sick, now i can sleep better at night :D

empress/compression_utils.py
fedarko: The "vector" part wasn't what I was confused about - these are vectors, but they're column vectors not row vectors, right? since this is iterating over the samples in the table, which are stored as columns. This differs from what the comment is saying

empress/tools.py
wasade: Keeping so sample order doesn't change. It's effectively just a get on an attribute so overhead is very low.

empress/compression_utils.py
fedarko: Fair -- if it turns out that the `_indexed` variables aren't needed in the below for loop then I think it would be more efficient to use the original enumerate stuff, but if we need to keep them around then I'm cool with sticking with this approach.

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F323%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F323%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F323%2Fempress-tree.qzv)

tests/python/test_compression_utils.py
```diff
@@ -322,7 +330,7 @@ def test_compress_table_basic(self):
 
         # First off, verify that compress_table() leaves the original table DF
         # untouched.
```
fedarko, line 332:
```suggestion
        # First off, verify that compress_table() leaves the original table
        # untouched.
```

tests/python/test_core.py
fedarko: ah sorry, i meant just removing the one line about transposing the table. the qurro attribution should probably stay (i don't really mind but due to BSD licensing stuff i try to attribute this stuff).

empress/compression_utils.py
wasade: Kind of? The axes in BIOM are labeled as "sample" and "observation", and I don't believe there is any contract at the API level about whether these are rows or columns. I'll change this but I'm not sure I agree :)

tests/python/test_core.py
wasade: Okay

empress/compression_utils.py
fedarko: The main concern I had was that the comment said "row," but as far as I could tell these were actually stored as columns internally.

... not that it really impacts things, just wanted to avoid confusing people. The code is now clear as is so this is fine, sorry if this comes across as :bike: :house: 'ing

fedarko:
Ok, everything looks good! Thanks so much @wasade! (╯°□°)╯︵ ┻━┻

wasade:
woooo!!! thanks!!!

ElDeveloper:
Thanks so much @wasade!

PR #326, opened 2020-08-13
PERF: Speed-up taxonomic string preprocessing

ElDeveloper:
The string operations are vectorized which makes processing the EMP
taxonomy strings go from 45 seconds to ~1.5 seconds.

### DIFF
--- a/empress/taxonomy_utils.py
+++ b/empress/taxonomy_utils.py
@@ -100,11 +100,19 @@ def split_taxonomy(feature_metadata):
                 "(case insensitive)."
             )
 
-        # Find the maximum number of ;s within any of the taxonomy annotations
-        # Thanks Yoshiki for showing how to do this concisely :)
-        max_sc_count = feature_metadata[tax_col_name].str.count(";").max()
+        # Split the single column of taxonomy strings into n columns, where n
+        # is the highest number of taxonomic levels in any string. We have to
+        # account for leading and trailing whitespace (.str.strip()) as well as
+        # for whitespace between delimiters (.str.split()). The regular
+        # expression splits tokens separated by semicolons and up to one
+        # leading or trailing space. Importantly, using the expand keyword
+        # ensures that all rows are padded to have n tokens. The final step is
+        # to replace those empty values with "Unspecified".
+        tax_levels = feature_metadata[tax_col_name]\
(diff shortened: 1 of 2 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F326%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F326%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F326%2Fempress-tree.qzv)

empress/taxonomy_utils.py
```diff
@@ -100,11 +100,19 @@ def split_taxonomy(feature_metadata):
...
+        # Split the single column of taxonomy strings into n columns, where n
+        # is the highest number of taxonomic levels in any string. We have to
+        # account for leading and trailing whitespace (.str.strip()) as well as
+        # for whitespace between delimiters (.str.split()). The regular
+        # expression splits tokens separated by semicolons and up to one
+        # leading or trailing space. Importantly, using the expand keyword
+        # ensures that all rows are padded to have n tokens. The final step is
+        # to replace those empty values with "Unspecified".
+        tax_levels = feature_metadata[tax_col_name]\
+            .str.strip().str.split(r'\s{0,1};\s{0,1}', expand=True)
```
fedarko, line 112: If I understand correctly, the regex `\s{0,1}` means 0 or 1 instances of a whitespace character. Could we replace these with `\s*`? (basically, would a silly dataset with like 20 tabs on one side of a semicolon cause problems here?)

I tried this out in python and it seems to work ok on such a dataset, but ... I'm not sure why this is the case. Less of a suggestion, more of a question :)

fedarko:
Looks good to me! Thanks so much @ElDeveloper :)

ElDeveloper:
Thanks @fedarko!

Issue #339, opened 2020-08-18
Create js using arrays Sets


kwcantrell:
The .add() function of Set is much slower than .push of an array. See this [post](https://stackoverflow.com/questions/39007637/javascript-set-vs-array-performance) for more details.

I also created simple test to verify and it seems creating an array and then converting it to a set is much faster than just using set.add(). In the below example, 'add by set' takes ~4 seconds will 'add by index' takes ~0.5 seconds. So, if we initialize all sets using arrays, we can save a significant amount of time. This would be especially helpful for the coloring/bar plot methods. Note: we still want to use sets because .has is much faster than indexOf.
 
```
var f = function() {
    var t = [];
    var d = new Date();
    for (var i = 0; i < 10000000; i ++) {
        t[i] = 1;
    }
    var st = new Set(t);
    st.has(1) ? console.log("!!!") : console.log(":(");
    var dt = new Date();
    console.log("add by index", dt.getTime() - d.getTime());

    var t = [];
    var d = new Date();
    for (var i = 0; i < 10000000; i ++) {
        t.push(i)
    }
    var st = new Set(t);
    st.has(1) ? console.log("!!!") : console.log(":(");
    var dt = new Date();
    console.log("add by push", dt.getTime() - d.getTime());

    var t = new Set();
    var d = new Date();
    for (var i = 0; i < 10000000; i ++) {
        t.add(i);
    }
    var dt = new Date();
    console.log("add by set", dt.getTime() - d.getTime());

    var d = new Date();
    var t = new Array(10000000);
    for (var i = 0; i < 10000000; i ++) {
        t[i] = 1;
    }
    var st = new Set(t);
    st.has(1) ? console.log("!!!") : console.log(":(");
    var dt = new Date();
    console.log("add by init", dt.getTime() - d.getTime());

    var d = new Date();
    var t = new Array(10000000).fill(0);
    for (var i = 0; i < 10000000; i ++) {
        t[i] = 1;
    }
    var st = new Set(t);
    st.has(1) ? console.log("!!!") : console.log(":(");
    var dt = new Date();
    console.log("add by init and fill", dt.getTime() - d.getTime());
}
f();
```

kwcantrell:
[...]

PR #345, opened 2020-08-20
Layout js

kwcantrell:
This PR moves the layout calculations from python to js. This dramatically reduced the size of the resulting html file in the .qzv. Before a tree with ~460,000 nodes created an html file with 171mb and after removing just the coordinate information brought down the file size to 50mbs. In addition, all node names in feature metadata (compressed_tm and compressed_im) and feature ids/feature map in the biom-table (fid2idxs and f_ids), were replaced with their postorder positions. Note that if a name in the internal feature metadata table mapped to multiple nodes, then all nodes would be added to compressed_im. Finally _namesToKeys was removed from empress since all internal data structures now return/use postorder positions of nodes instead of names.

Summary Notes:  

- Layouts are calculated in js

- bp-tree now uses postorder for names/lengths array
- Empress:
    - getAttr now takes a node's postorder position instead of its _treeData array
    - _namesToKeys has been removed. All data structures (i.e. metadata/biom-table now store a nodes postorder position instead of node name. So, for example the keys for _tipMetadata/_intMetadata are now nodes postorder positions. In addition functions like getObservationUnionForSamples() no longer return a list of node names but instead returns a list of postorder positions which can then be directly used to index _treeData/metadata. Also, since internal nodes may not be unique, if multiple internal nodes with the same name exist in the feature metadata, then they will all be added to _intMetadata. Store all nodes with the same name (only if they have an entry in the feature metadata) might be a bit more memory intensive and might be worth addressing in future PRs however, this solution was the simplest to implement.
- keepUniqueKeys is a bit faster (Note: The goal was to speed up things such as sample metadata coloring however, it turns how initialization of Sets was the bottleneck for more details see #339.


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F345%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F345%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F345%2Fempress-tree.qzv)

empress/support_files/js/empress.js
```diff
@@ -297,31 +305,77 @@ define([
         this._group = new Array(this._tree.size + 1).fill(-1);
     }
 
+    /**
+     * Computes the various tree layouts and fills _treeData
+     */
+    Empress.prototype.layouts = function () {
```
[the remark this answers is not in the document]
antgonza: Agreed!

empress/support_files/js/layouts-util.js
```diff
@@ -0,0 +1,492 @@
...
+        // context.
+
+        var maxWidth = 0;
+        var maxHeight = 0;
+        var prevY = 0;
+        var xCoord = new Array(tree.size + 1).fill(0);
+        var yCoord = new Array(tree.size + 1).fill(0);
+        var highestChildYr = new Array(tree.size + 1).fill(0);
+        var lowestChildYr = new Array(tree.size + 1).fill(
+            Number.POSITIVE_INFINITY
```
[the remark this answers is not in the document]
antgonza: This is a good catch, should undefined be defined as the default for all of these values? If we agree, perhaps adding this in the coding guidelines (if there are any) would be good ...

empress/support_files/js/layouts-util.js
```diff
@@ -0,0 +1,492 @@
...
+        // normal x/y coordinates (a.k.a. Cartesian coordinates).
+        // We skip the root because we already know it's going to be at (0, 0).
+        //
+        // Unlike the rectangular / unrooted layout we need to keep track of
+        // two sets of positions for each (non-root) node: a "starting" and
+        // "end" point. In the other layouts we can just infer the starting
+        // point during drawing, but here each node's line begins at its parent
+        // node's radius but at its own angle in polar coordinates -- and to
+        // avoid doing extra work on converting between polar and Cartesian
+        // coords, we just compute both points here. (As a potential TODO, it's
```
antgonza, line 316: I normally add ToDos as issues so we don't forget; thoughts?

empress/support_files/js/layouts-util.js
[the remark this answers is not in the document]
kwcantrell: Good catch @fedarko! I agree with @antgonza, `undefined` should be the default. The only reason I set them to +/-Infinity was to avoid having something like this:
```
if (lowestChildYr[i] === undefined) {
    lowestChildYr[i] = value;
} else if (value < lowestChildYr[i] {
    lowestChildYr[i] = value;
}
```
However, it looks like if you compare a number to `undefined` then js will always return false.

empress/support_files/js/layouts-util.js
kwcantrell: @antgonza that is probably a good idea! I think we a few locations with `TODO` in the code base. I'll create an issue!

empress/support_files/js/empress.js
```diff
@@ -1527,18 +1554,19 @@ define([
...
             throw 'F. metadata coloring method "' + method + '" unrecognized.';
         }
-
         // Produce a mapping of unique values in this feature metadata
         // column to an array of the node name(s) with each value.
         var uniqueValueToFeatures = {};
         _.each(fmObjs, function (mObj) {
-            _.mapObject(mObj, function (fmRow, nodeName) {
+            _.mapObject(mObj, function (fmRow, node) {
+                // need to convert to interger
```
fedarko, line 1562:
```suggestion
                // need to convert to integer
```

empress/support_files/js/layouts-util.js
```diff
@@ -0,0 +1,499 @@
...
+
+        // We don't check if max_width == 0 here, because we check when
+        // constructing an Empress tree that it has at least one positive
+        // branch length and no negative branch lengths. (And if this is the
+        // case, then max_width must be > 0.)
+        var xScalingFactor = width / maxWidth;
+
+        // Since this will be multiplied by 0 for every node, we can set
+        // this to any real number and get the intended "effect" of keeping
+        // every node's y-coordinate at 0.
```
fedarko, line 102:
```suggestion
        // If the tree is a straight line (i.e. all internal nodes have exactly
        // one child), every node in the tree will have a y-coordinate of 0. In
        // this case, this scaling factor won't matter, since it'll get
        // multiplied by 0 for every node.
```

eesh, I had no idea what this comment meant at first. this is probably a problem since IIRC i wrote it :P this should make it clearer, hopefully...

empress/support_files/js/layouts-util.js
```diff
@@ -0,0 +1,499 @@
...
+     *     Description above + the implementation of this algorithm derived
+     *     from the Polar layout algorithm code.
+     *
+     * @param {BPTree} tree The tree to generate the coordinates for.
+     * @param {Float} width Width of the canvas where the tree will be
+     *                      displayed.
+     * @param {Float} height Height of the canvas where the tree will be
+     *                       displayed.
+     * @param {Boolean} normalize If true, then the tree will be scaled up to
+     *                            fill the bounds of width and height
```
fedarko, line 218:
```suggestion
     *                            fill the bounds of width and height.
```

empress/support_files/js/layouts-util.js
```diff
@@ -0,0 +1,499 @@
...
+     *  https://github.com/qiime/Topiary-Explorer/blob/master/src/topiaryexplorer/TreeVis.java
+     *      Derived from the "Rectangular" layout algorithm code.
+     *
+     * @param {BPTree} tree The tree to generate the coordinates for.
+     * @param {Float} width Width of the canvas where the tree will be
+     *                      displayed.
+     * @param {Float} height Height of the canvas where the tree will be
+     *                       displayed.
+     * @param {Boolean} normalize If true, then the tree will be scaled up to
+     *                            fill the bounds of width and height
```
fedarko, line 31:
```suggestion
     *                            fill the bounds of width and height.
```

empress/support_files/js/layouts-util.js
```diff
@@ -0,0 +1,499 @@
...
+            y1: y1,
+            angle: angle,
+            arcx0: arcx0,
+            arcy0: arcy0,
+            arcStartAngle: arcStartAngle,
+            arcEndAngle: arcEndAngle,
+        };
+    }
+
+    function unrootedLayout(tree, width, height, normalize = true) {
```
fedarko, line 419: Would you mind adding a docstring for this?

tests/python/test_core.py
```diff
@@ -274,24 +261,18 @@ def test_to_dict_with_feature_metadata(self):
...
         self.assertEqual(obs, dict_a_with_fm)
 
     def test_to_dict_with_metadata_nans(self):
         nan_sample_metadata = self.sample_metadata.copy()
         nan_feature_metadata = self.feature_metadata.copy()
         nan_sample_metadata.at["Sample2", "Metadata4"] = np.nan
-        nan_feature_metadata.at["h", "fmdcol1"] = np.nan
-        nan_feature_metadata.at["a", "fmdcol2"] = np.nan
+        nan_feature_metadata.at['h', "fmdcol1"] = np.nan
+        nan_feature_metadata.at['a', "fmdcol2"] = np.nan
```
fedarko, line 275: Is there a reason for changing the quotes here? I don't really mind either way, just wondering.

kwcantrell: I don't believe so. I changed the to double quotes.

tests/test-layouts-util.js
```diff
@@ -0,0 +1,430 @@
...
+        var toFixedIfy = function (arr) {
+            return _.map(arr, function (ele) {
+                return ele.toFixed(4);
+            });
+        };
+        // Given two arrays of numbers, calls toFixedIfy() on each and
+        // asserts deep equality on the results.
+        var approxDeepEqual = function (arr1, arr2, message) {
+            deepEqual(toFixedIfy(arr1), toFixedIfy(arr2), message);
+        };
```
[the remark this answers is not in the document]
kwcantrell: That a good suggestion. I'll move them to `utilities-for-testing.js` so that other  tests can take advantage of these functions.

tests/test-layouts-util.js
```diff
@@ -0,0 +1,430 @@
...
+                [0, 0, 0, tpi, 0, 0],
+                "arc end angle"
+            );
+        });
+
+        test("Test unrooted layout", function () {
+            var obs = LayoutsUtil.unrootedLayout(this.tree, 1, 1);
+            var exp = {
+                xCoord: [
+                    undefined,
```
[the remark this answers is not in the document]
kwcantrell: updated

empress/support_files/js/empress.js
```diff
@@ -17,7 +18,8 @@ define([
     CanvasEvents,
     BarplotPanel,
     util,
-    chroma
+    chroma,
+    LayoutsUtil
```
fedarko, line 22: It looks like there are still docs for old parameters of the Empress constructor (`nameToKeys`, `layoutToCoordSuffix`, `defaultLayout`, etc.) included in the comment below. Could you please remove these?

empress/support_files/js/empress.js
```diff
@@ -108,27 +133,13 @@ define([
          *       indices start at 1, for now; see #223 on GitHub.
```
fedarko, line 133: Could you update this comment to not mention nameToKeys? Since AFAIK that is removed in this PR

empress/support_files/js/empress.js
```diff
@@ -297,31 +305,97 @@ define([
...
+     */
+    Empress.prototype.getLayoutInfo = function () {
+        var data, i;
+        // Rectangular
+        if (this._currentLayout === "Rectangular") {
+            data = LayoutsUtil.rectangularLayout(this._tree, 4020, 4020);
+            this._yrscf = data.yScalingFactor;
+            for (i = 1; i <= this._tree.size; i++) {
+                // remove old layout information
+                this._treeData[i].length = this._tdToInd.visible + 1;
```
fedarko, line 319: I did not realize it was possible to truncate arrays in JS like this! Dang.

One thing that could complicate things later on is that, if `visible` is no longer the last attribute before the layout stuff in `this._tdToInd`, this code will get messed up (either silently or not-so-silently, I'm not sure). I think there are a few ways around this -- a few I can think of:

1. add a comment next to `this._tdToInd` above that `visible` must be the last attribute
2. add a class-level variable next to `this._tdToInd` called e.g. `this._numAllLayoutNodeProperties` or something, which we'd set to `3` for now (and could update as needed). We could then use this here in place of `this._tdToInd.visible + 1`.

either's ok with me (or if you can think of a better way to handle this that's also cool), but I do think this should be a bit more foolproof (I can never remember this stuff after a few months :P)

empress/support_files/js/bp-tree.js
```diff
@@ -229,7 +248,14 @@ define(["ByteArray"], function (ByteArray) {
      * @return{String}
      */
     BPTree.prototype.name = function (i) {
-        return this.names_[this.preorder(i) - 1];
+        return this.names_[this.postorder(i)];
+    };
+
+    /**
+     * Returns an array of all node names in tree
```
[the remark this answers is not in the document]
kwcantrell: Good suggestion. I added the shift() inside the function.

empress/support_files/js/empress.js
kwcantrell: Thanks for catching this. Updated doc string to reflect new changes.

fedarko:
Thanks so much @kwcantrell! Looks good to me.

PR #340, opened 2020-08-20
Refactor and test legend code; show barplot legends

fedarko:
Closes #299 and closes #331. (Also counts as progress towards #142, since now `legend.js` is tested.) Also updates the barplots section of the README accordingly.

This PR adds a pretty large amount of documentation / tests, both to new and previously-existing code.

barplotlegendswoot

### Various notes
- This fixes a weird bug where it was possible for barplot layers' HTML elements' IDs to collide in certain circumstances. See [this comment](https://github.com/fedarko/empress/blob/3ed215b4179e68685433ffbe212fca800963e3ff/empress/support_files/js/barplot-layer.js#L26-L41) for an explanation.
 - The CSS changed a decent amount in this PR, but not _that_ much -- the main reason the CSS file has a lot of diffs is I went ahead and added it to the collection of files run through prettier (so now it should be consistently styled).
- For now, gradients match Emperor in that they go from high -> low. It would be nice to flip these around if/when https://github.com/biocore/emperor/issues/782 is addressed. (For now, I assume consistency is important. also i don't know enough SVG to do this myself right now lol)

PR #346, opened 2020-08-21
Draw barplots in the circular layout

### LINKED ISSUE #297: Add support for circular layout barplots
Continuation of #201, since the #293 PR only supports rectangular layouts.

### LINKED ISSUE #343: Don't immediately draw barplots when clicking the "Draw barplots?" checkbox, to avoid crashing
This behavior isn't a problem when there's feature metadata provided (and the default barplot layer is just an empty red bar or something), but when there isn't feature metadata -- and barplots default to stacked barplots of sample metadata -- then the current behavior means that a stacked barplot of whatever the "first" sample metadata field is will be drawn. If the sample metadata field in question is not categorical and/or it has a ton of groups, then the resulting stacked barplot will take a super long time to draw and/or just straight-up crash the browser.

This is currently a problem with the EMP tree (~200k tips with the chunk of it I'm testing with): the first sample metadata field is `adiv_chao1`, and trying to draw a stacked barplot of that just straight up crashes the page in Chromium. (I get a [`SIGTRAP` error](https://en.wikipedia.org/wiki/Signal_(IPC)#SIGTRAP), bizarrely enough. Didn't know what that meant.).

By deferring drawing barplots until the user presses "Update", the user is responsible for choosing what field to use. (I guess this sort of matches the behavior Emperor plots with hundreds of thousands of samples -- IIRC trying to color by a similarly silly field there will also hang, but when the hang happens it's clear to the user _why_ things hanged.)

Changing this in the code should really just boil down to removing [this line](https://github.com/biocore/empress/blob/4f6ddd3a4e03b85703f733bd75b9771ef82157b4/empress/support_files/js/barplot-panel-handler.js#L77) of code, but we'll have to update the README as well since it mentions how a red bar pops up immediately (rather than on pressing Update).

fedarko:
Closes #297 and closes #343.

This should probably not be merged until #345 and #340 are, but this branch does work in case people want to use this functionality.

Thanks!

### Issues we should probably discuss

- Barplot lengths are still in arbitrary units, so a barplot with length "100" looks thinner in the moving pictures circular layout than in the rectangular layout. Standardizing these somewhat should be possible, but I wanted to avoid doing that until #345 is in (to avoid having to redo the work to accommodate JS layouts).

- Circular layout bars are drawn here by drawing two rectangles (reusing an image I posted [here](https://github.com/biocore/empress/issues/297#issuecomment-676949805)) --
  image
  We may want to smooth this out, and look at the number of tips in the tree to determine how many rectangles to draw (e.g. use ~15 rectangles for tiny trees like the one above, and only 1 rectangle for massive EMP-scale trees). Or we could just leave this as is until the future, when we'd ideally have a shader draw these as actual curved sections.

- Although the functions that add the coordinates for individual bars (in both rect/circular layouts) are now tested, `addFMBarplotLayerCoords()` and `addSMBarplotLayerCoords()` are still untested. Testing these is going to be a bit time-consuming -- testing the trig stuff gets really annoying really quickly (see e.g. the `_addCircularBarCoords()` test here...), so I think it makes sense to defer testing these until we agree on how many rectangles to use for approximating a curve (per the above point).

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ The plot is now updated so each branch is now colored by its Phylum level classi
 
 ### Exploring a feature’s closest common ancestors  
 
-So far, we’ve looked at our data using the default unrooted tree view. To visually locate these features’ closest common ancestors, it may be easier to switch to a different layout. From the main menu, click *Layout* then select *Circular* (or *Rectangular*). Our plot automatically switches to a rooted circular cladogram.   
+So far, we’ve looked at our data using the default unrooted tree view. To visually locate these features’ closest common ancestors, it may be easier to switch to a different layout. From the main menu, click *Layout* then select *Circular* (or *Rectangular*). Our plot automatically switches to a rooted layout.
 
 Now zoom into the longest branch of the top cluster and click on the closest external node that has a different Phylum classification (light blue).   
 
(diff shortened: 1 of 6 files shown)


fedarko:
Closing in favor of #357.

Issue #321, opened 2020-08-11
Coloring root nodes of collapsed clades?


fedarko:
We talked a bit about this a few days ago, so I'm making a note of it here for future reference. I think it might make sense to color in the root nodes of collapsed clades, since they would already be colored that way if collapsing were not done. If opinions are split on which option (color root vs. don't color root) looks better / is more useful, we could make this an option in "Tree Properties" as well.

![image](https://user-images.githubusercontent.com/4177727/89845914-a7e7f380-db34-11ea-9ecf-ba5132ccbaf1.png)

fedarko:
for reference, this seems to be configurable by removing this line:

https://github.com/biocore/empress/blob/4f6ddd3a4e03b85703f733bd75b9771ef82157b4/empress/support_files/js/empress.js#L2291-L2293

here's what this looks like on the same portion of the MP tree --

![image](https://user-images.githubusercontent.com/4177727/91108684-7b9c9e80-e62d-11ea-8ede-57f65d0f3393.png)

The main benefit is that this is nice for certain trees where the collapsed clade's area is relatively small, e.g. the MP tree in rect layout collapsed at level 1 taxonomy --

![image](https://user-images.githubusercontent.com/4177727/91108762-b4d50e80-e62d-11ea-9b18-ab38fbf18aa9.png)

![image](https://user-images.githubusercontent.com/4177727/91108738-a2f36b80-e62d-11ea-850d-9dffe31c30db.png)

notice the small red clade containing the only Archaea-classified features in the tree. In the first screenshot it's still visible (it's just really tiny), but in the second screenshot coloring the root branch red makes (like normal) makes it a lot more visible.

ElDeveloper:
Thanks for putting this together @fedarko, this makes a lot of sense. Thinking about it the collapsing shouldn't change the coloring of the branches, so we either extend the polygons to the top of the branch, or we color the branch as you suggest. I suspect leaving the branch makes for a more aesthetically pleasing representation.

PR #363, opened 2020-09-01
Support ignoring branch lengths during layout; update tree properties panel UI; some minor clade collapsing changes

### LINKED ISSUE #271: Support ignoring branch lengths during layout
Suggestion from @gibsramen last week.

This (along with #170) would be a nice use case of doing layouts in the JavaScript code (#213), since that'd allow users to compare the same tree with and without branch lengths being used quickly.

For reference, here are some comparison screenshots on the moving pictures dataset (coloring by phylum):

#### With branch lengths

image

#### Uniform branch length of 1

image

#### With branch lengths
image

#### Uniform branch length of 1
image

#### With branch lengths
image

#### Uniform branch length of 1
image

fedarko:
Closes #271. Also wound up closing #321 (ran into it along the way).

ignoooore

The main challenge here was making sure that updating the "ignore lengths?" setting didn't mess with the tree state. This mostly wasn't a problem, but I had to mess around some with how clade collapsing works (since as far as I can tell the previous clade collapsing stuff assumed that clades computed for a given layout could be cached; but here the same layout can have multiple ways of collapsing since the node coordinates vary based on the ignore lengths setting). I believe this works as it should now, but I'm not very familiar with this part of this code.

Other thing we should discuss -- the UI says that "the root node's length is always ignored", but it looks like the unrooted layout uses the root node's length at one point. Since we don't require that the root node has a valid or even specified length, we should discuss this to see what to do.
 
Thanks!

### DIFF
--- a/empress/support_files/js/bp-tree.js
+++ b/empress/support_files/js/bp-tree.js
@@ -629,14 +629,19 @@ define(["ByteArray", "underscore"], function (ByteArray, _) {
      *
      * @param {Number} start The postorder position of a node
      * @param {Number} end The postorder position of a node
+     * @param {Boolean} ignoreLengths If truthy, treat all node lengths as 1;
+     *                                if falsy, actually consider node lengths
      *
      * @return {Number} the sum of length from start to end
      */
-    BPTree.prototype.getTotalLength = function (start, end) {
+    BPTree.prototype.getTotalLength = function (start, end, ignoreLengths) {
         var curNode = start;
         var totalLength = 0;
         while (curNode !== end) {
-            totalLength += this.length(this.postorderselect(curNode));
+            var nodeLen = ignoreLengths
+                ? 1
+                : this.length(this.postorderselect(curNode));
+            totalLength += nodeLen;
             curNode = this.parent(this.postorderselect(curNode));
             if (curNode === -1) {
                 throw "Node " + start + " must be a descendant of " + end;
--- a/empress/support_files/js/bp-tree.js
+++ b/empress/support_files/js/bp-tree.js
@@ -638,11 +638,10 @@ define(["ByteArray", "underscore"], function (ByteArray, _) {
         var curNode = start;
         var totalLength = 0;
         while (curNode !== end) {
-            var nodeLen = ignoreLengths
-                ? 1
-                : this.length(this.postorderselect(curNode));
+            var popos = this.postorderselect(curNode);
+            var nodeLen = ignoreLengths ? 1 : this.length(popos);
             totalLength += nodeLen;
-            curNode = this.parent(this.postorderselect(curNode));
+            curNode = this.parent(popos);
             if (curNode === -1) {
                 throw "Node " + start + " must be a descendant of " + end;
             }
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -2244,16 +2244,12 @@ define([
         var supported = this._barplotPanel.updateLayoutAvailability(
             this._currentLayout
         );
-        // TODO: don't call drawTree() from either of these barplot
-        // funcs, since it'll get called in centerLayoutAvgPoint anyway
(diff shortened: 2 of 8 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F363%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F363%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F363%2Fempress-tree.qzv)

empress/support_files/templates/side-panel.html
```diff
@@ -238,8 +241,41 @@
...
+  <p class="side-panel-notes indented">
+    If checked, the camera will automatically focus on nodes that are selected
+    via searching by name or by clicking on an arrow in a biplot.
+  </p>
+  <hr />
+  <p class="side-panel-inner-header">
+    Layout options
+  </p>
+  <p>
+    <label for="ignore-lengths-chk">Ignore node lengths?</label>
```
ElDeveloper, line 268: I the standard term seems to be `branch length` would you mind changing to refer to branch lengths instead of node lengths?

empress/support_files/js/bp-tree.js
```diff
@@ -629,15 +629,19 @@ define(["ByteArray", "underscore"], function (ByteArray, _) {
      *
      * @param {Number} start The postorder position of a node
      * @param {Number} end The postorder position of a node
+     * @param {Boolean} ignoreLengths If truthy, treat all node lengths as 1;
+     *                                if falsy, actually consider node lengths
      *
      * @return {Number} the sum of length from start to end
      */
-    BPTree.prototype.getTotalLength = function (start, end) {
+    BPTree.prototype.getTotalLength = function (start, end, ignoreLengths) {
```
ElDeveloper, line 637: Can you add a test with this new argument?

empress/support_files/templates/side-panel.html
fedarko: Good point -- just changed. For now the description still says "root node's length" -- does this seem ok to you, or would you like me to say e.g. "root node's branch length"?

empress/support_files/js/bp-tree.js
fedarko: Done, thanks!

empress/support_files/templates/side-panel.html
ElDeveloper: Sounds good to me!

empress/support_files/js/empress.js
```diff
@@ -2207,34 +2260,11 @@ define([
...
-                    this.undrawBarplots();
-                } else if (supported && this._barplotPanel.enabled) {
-                    this.drawBarplots(this._barplotPanel.layers);
-                }
+                this.reLayout();
                 // recenter viewing window
-                // Note: this function calls drawTree()
+                // NOTE: this function calls drawTree(), which is redundant
+                // since reLayout() already called it. Would be good to
+                // minimize redundant calls to that.
```
kwcantrell, line 2267: if you haven't already, can you create an issue for this.

kwcantrell (commented):
Thanks @fedarko this looks nice! I just have one minor comment.

empress/support_files/js/empress.js
fedarko: Gotcha -- just set up https://github.com/biocore/empress/issues/367. Thanks!

ElDeveloper:
Thanks both!

Issue #197, opened 2020-06-22
Repeated box-selections in Emperor cause later selections to be "cut off" early


fedarko:
Documented in https://github.com/biocore/empress/pull/193#issuecomment-647775924. Minor, but could be confusing.

PR #370, opened 2020-09-03
ENH: Link emperor double-click event and add timer cancelation

ElDeveloper:
This adds the ability to double-click on a color category in Emperor and highlighting the tips in the tree represented by that group. This also adds a timer cancellation to prevent problems with sample selections as reported in #197.

double-click

The documentation on the callback is fairly lax but should be improved once the explanation in #319 is merged.

### DIFF
--- a/empress/support_files/js/node-click-callback.js
+++ b/empress/support_files/js/node-click-callback.js
@@ -22,3 +22,31 @@ ec.sceneViews[0].on("click", function (name, object) {
 
     empress.showNodeMenuForName(name);
 });
+
+ec.controllers.color.addEventListener("value-double-clicked", function (
+    payload
+) {
+    // cancel any ongoing timers
+    clearTimeout(empress.timer);
+
+    // if there's any coloring setup remove it, and re-enable the update button
+    sPanel.sUpdateBtn.classList.remove("hidden");
+    sPanel.fUpdateBtn.classList.remove("hidden");
+    legend.clear();
+    empress.resetTree();
+
+    var names = _.map(payload.message.group, function (item) {
+        return item.name;
+    });
+    var container = {};
+    container[payload.message.attribute] = names;
+    empress.colorSampleGroups(container);
+
+    // 4 seconds before resetting
+    empress.timer = setTimeout(function () {
+        empress.resetTree();
+        empress.drawTree();
+
+        plotView.needsUpdate = true;
+    }, 4000);
+});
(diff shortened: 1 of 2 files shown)


ElDeveloper:
HT to @antgonza for the idea 👍

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F370%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F370%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F370%2Fempress-tree.qzv)

fedarko (changes requested):
Thanks @ElDeveloper! This functionality is really handy.

Notes from review:

1. It looks like double-clicking on biplot color categories causes an error: this is what pops up in the console when I double-click the purple (`p__Proteobacteria`) Level 2 category:
   purp
   It looks like the code is trying to treat feature IDs as sample IDs, which causes the problem. I suppose ideally there would be some fancy interaction event where all nodes in this feature metadata group would be selected / colored in Empress or something, but the easiest way to handle this for now would probably just be detecting that a biplot category was selected and not doing anything.

2. I haven't been able to figure out exactly *why* this is happening, but it seems like there are still some #197-ish bugs that can occur:
   eitherbugsorimgoingcrazy

   One variant of these problems is reproducible when you double-click on `gut`, then select all `gut` samples: when this happens the samples remain highlighted --
   gutgutgut
   This also seems to be doable with `tongue` samples, as well as doable in the opposite order (select all `gut` samples then double-click on `gut`, resulting in the samples remaining highlighted indefinitely) --
   gut2gut
   Not a very important issue, since I think at this point these problems only really show up when you try to break things :), but it would be nice to at least document these as we make this stuff more and more robust.

ElDeveloper:
@fedarko thanks for the review, I think these bugs are addressed. Let me know if you get a chance to try this out.

fedarko (approved):
@ElDeveloper Thanks so much, looks good to me!

fedarko:
Thanks @ElDeveloper! :point_up_2: :point_up_2: :timer_clock:

ElDeveloper:
Thanks so much @fedarko!

PR #384, opened 2020-09-18
Some SVG export fixes; add PNG export; other misc. improvements

### LINKED ISSUE #330: Add PNG export option
I was doing some reading, and I think this shouldn't be too bad ...?

See https://webglfundamentals.org/webgl/lessons/webgl-tips.html#screenshot.

### LINKED ISSUE #334: Exported SVG images are flipped upside down

### LINKED ISSUE #348: Support drawing node circles for all nodes, regardless of if they were named in the Newick file or not
Not all that important, but could be something worth looking in to later. In QIIME 2, the tree is stored as a QZA -- the naming conventions in the underlying Newick string are probably not often considered by users, so they may not care that certain nodes have or don't have a name.

### LINKED ISSUE #349: Don't draw node circles by default for large trees?
At least on the EMP tree I was working with, toggling node circles took a noticeable amount of time (a few seconds IIRC). If this time is presumably being used on startup to draw node circles, I think we should save that time so that the initial loading period is as short as possible -- allowing users to draw node circles later, of course.

Not sure what a good threshold would be; maybe 100k tips? (We could also just disable node circles by default; IDK what would be ideal.)

fedarko:
- Fix #330: Add (initial, basic) PNG export. It's a bit clunky in that what is exported changes as the user zooms around, but it should at least help for creating the paper figures right now (since the PNG export includes barplots/collapsed clades).

- Split up SVG exporting somewhat -- now, one button exports the tree (and eventually other stuff drawn on the canvas), and another button exports the legend(s).

- Fix #334: the exported images are now no longer flipped upside-down (or at the very least the vertical orientation is consistent with Empress' drawer; see that thread for discussion)

- Fix #348: node circles are now drawn for all nodes, regardless of if they have a name or not

- Tidy up the UI for unnamed nodes: now, rather than showing `Name: null` in the selection menu, these nodes are described as follows:

    unnamed

- Fix #349: by default, node circles are not drawn.

- Always draw a circle for the green "selected node" shape, regardless of whether or not node circles are drawn. (See the comment this PR adds above `c.uniform1i(s.isSingle, 1);` in `drawer.js` for details about this. Not sure if the selected node shape turning into a square when node circles were turned off was intentional or not, but it should remain a circle now.)

- In the SVG export, don't draw a black circle at the root node any more (since #211 was addressed a while back).

- The main legend is now contained within `Empress`, instead of within `SidePanel`. This makes accessing the main legend in the SVG export easier, and makes things a bit more straightforward. (Similarly, now `Animator` doesn't have direct ownership of the legend.)

- Legend SVG exporting code is now delegated to `Legend` objects (i.e. each legend now has an `exportSVG()` method we can call to get its SVG, making combining multiple legends' SVGs not too bad)

  - Currently just the tree legend is exported, but this can already handle multiple legends -- although I need to add support for exporting continuous / length legends.

- Exported SVG legends resemble the "new"-styled legends more closely:

    level2
    level2legend

- Some other minor improvements (split up SVG exporting code, declare variables in advance, use `camelCase` instead of `snake_case`, make the node radii class attributes of the `Drawer`, etc.)

Barplots and clade collapsing are still not supported in the SVG export -- that is up next. But these changes should make doing that a lot easier.

### Things to maybe discuss

... Is the `this.NODE_CIRCLE_RADIUS` variable (formerly just `s.pointSize`, default `4.0`) actually a radius, or is it a diameter or something? Not sure what units it's in. (The SVG exporting code assumed it was a radius but IDK if that is true or not.)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F384%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F384%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F384%2Fempress-tree.qzv)

kwcantrell:
@fedarko this is great! I am still going over the PR. However, I wanted to ask if you think it would make sense to move the svg and png to an "export" class instead of the empress class?

fedarko:
@kwcantrell Sure, that would work. I think it'd make sense to still have some exporting stuff within Empress (for doing things like tallying up all the active legends), but I just moved the bulk of the exporting code to `export-util.js`, a new module (analogous to `layouts-util.js`).

kwcantrell:
Thanks @fedarko, looks good!

ElDeveloper:
Thanks both!

PR #386, opened 2020-09-22
Configurable barplot borders

### LINKED ISSUE #350: Add barplot padding/border options
- Option to have empty space between adjacent layers (doable by just increasing the `prevLayerMaxX` / `prevLayerMaxD` variable by some number every iteration through [this loop](https://github.com/biocore/empress/blob/4f6ddd3a4e03b85703f733bd75b9771ef82157b4/empress/support_files/js/empress.js#L1102-L1116)). This results in plots looking like:
  padding
  (This is of course inspired by [Anvi'o](http://merenlab.org/software/anvio/), which AFAICT does this by default for its circular barplots and looks really pretty)

- Option to surround bars with a border of a specifiable color (black would be a good default). This (along with the above option, now that i think of it) is currently doable by adding in "feature metadata" barplots before / after each barplot layer with a small length and of a specified default color, but that's cumbersome and isn't currently possible for visualizations where no feature metadata is provided.

fedarko:
Closes #350, and should at least make generating figures for the paper a lot easier.

This also removes some of the extra whitespace in the barplot panel when `Draw barplots?` is unchecked -- this was due to a `<p>` tag being empty but not hidden, which caused some extra vertical space to be taken up.

borders

### DIFF
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -6,9 +6,11 @@ define([
     "VectorOps",
     "CanvasEvents",
     "BarplotPanel",
+    "Legend",
     "util",
     "chroma",
     "LayoutsUtil",
+    "ExportUtil",
 ], function (
     _,
     Camera,
@@ -17,9 +19,11 @@ define([
     VectorOps,
     CanvasEvents,
     BarplotPanel,
+    Legend,
     util,
     chroma,
-    LayoutsUtil
+    LayoutsUtil,
+    ExportUtil
 ) {
     /**
      * @class EmpressTree
@@ -141,6 +145,13 @@ define([
         }
 
         /**
+         * @type {Legend}
+         * Legend describing the way the tree is colored.
+         * @private
+         */
+        this._legend = new Legend(document.getElementById("legend-main"));
+
+        /**
          * @type {BiomTable}
          * BIOM table: includes feature presence information and sample-level
          * metadata.
@@ -234,6 +245,12 @@ define([
          * Can be passed as input to this.thickenColoredNodes().
          */
         this._currentLineWidth = 0;
+
+        /**
+         * @type{Bool}
+         * Whether or not to draw node circles.
+         */
+        this.drawNodeCircles = false;
 
         /**
          * @type{Bool}
@@ -392,6 +409,8 @@ define([
         this._drawer.initialize();
         this._events.setMouseEvents();
         var nodeNames = this._tree.getAllNames();
+        // Don't include nodes with the name null (i.e. nodes without a
+        // specified name in the Newick file) in the auto-complete.
         nodeNames = nodeNames.filter((n) => n !== null);
         nodeNames.sort();
         this._events.autocomplete(nodeNames);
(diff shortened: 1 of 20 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F386%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F386%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F386%2Fempress-tree.qzv)

empress/support_files/js/empress.js
```diff
@@ -1876,6 +1886,82 @@ define([
...
+            halfAngleRange = Math.PI / this._tree.numleaves();
+        }
+        // Currently, this just draws a bar for every tip. This is relatively
+        // slow! For the rectangular layout, it should be possible to speed
+        // this up by figuring out the topmost and bottommost node and then
+        // drawing just two triangles (one rectangle, based on their y-values).
+        // For the circular layout, how to speed this up is less clear -- I
+        // suspect it should be possible using WebGL and some fancy
+        // trigonometry somehow, but I'm not sure.
+        for (node = 1; node < this._tree.size; node++) {
```
ElDeveloper, line 1937:
```suggestion
        for (var node = 1; node < this._tree.size; node++) {
```

empress/support_files/js/empress.js
```diff
@@ -1876,6 +1886,82 @@ define([
...
+        }
+        // ... Otherwise, we actually have to go and create bars
+        var halfyrscf, halfAngleRange;
+        if (this._currentLayout === "Rectangular") {
+            halfyrscf = this._yrscf / 2;
+        } else {
+            halfAngleRange = Math.PI / this._tree.numleaves();
+        }
+        // Currently, this just draws a bar for every tip. This is relatively
+        // slow! For the rectangular layout, it should be possible to speed
```
ElDeveloper, line 1931: Could we figure out a way to always draw 720 squares regardless of the tree size? That should give us enough resolution to make sure the circle renders smoothly. Any thoughts?

tests/test-barplots.js
```diff
@@ -56,6 +56,19 @@ require([
...
+        deepEqual(
+            empress._barplotPanel.borderLength,
+            BarplotLayer.DEFAULT_LENGTH / 2
+        );
+        // Gotta convert to numbers since getAttribute() seems to return "0"
+        deepEqual(
+            +empress._barplotPanel.borderLengthInput.getAttribute("min"),
+            +BarplotLayer.MIN_LENGTH
+        );
+    });
```
ElDeveloper, line 71: Can you add a test for `BarplotPanel.prototype.initBorderOptions`.

ElDeveloper (changes requested):
Functionality looks great, and is going to be very useful overall. One unexpected thing I noticed is that adding border around barplots adds space between the tips and the first layer. Technically the approach is valid but I think what we've noticed that would be useful for figures is the space between barplot layers, not space between barplots and the tree.

fedarko:
> One unexpected thing I noticed is that adding border around barplots adds space between the tips and the first layer. Technically the approach is valid but I think what we've noticed that would be useful for figures is the space between barplot layers, not space between barplots and the tree.

Yeah, this is because the code is technically allocating space for a border to be drawn here. It looks a bit clearer if you select black as the border color:

bord

I would kind of prefer to keep it this way, just because I think consistently drawing borders this way is more intuitive (at least to me). However I'd be down to (probably in a new PR) add in an option to toggle whether or not to draw borders before the innermost and after the outermost barplot layers.

empress/support_files/js/empress.js
fedarko: Yeah, it should definitely be possible to approximate this as a series of triangles (similar to how the clade collapsing shapes are rendered). My main concern would be ensuring that the approximation doesn't end up accidentally overlapping the next barplot -- the math might get kind of finicky, and spending a lot of time optimizing this might be too much work for its own good. @kwcantrell and I were talking earlier about ways of coercing WebGL to draw arcs using the same sort of math as in [here](https://github.com/biocore/empress/blob/3968bd42023b551aa76271b20ebfe19666762fe4/empress/support_files/js/drawer.js#L28-L31); this may be applicable to this sort of idea (draw a square centered at the tree as a really big point, and discard everything but the section within a given radius range corresponding to this barplot padding).

tests/test-barplots.js
fedarko: The current test was already kind of testing that function (the `min` attribute is set in that function), but I just added in some extra stuff that:

1. tests that the default color is properly applied to the color picker
2. tests that the default length value for the actual length input is properly set
3. cleans and adds documentation to the test code a bit
4. and mentions `initBorderOptions` in the test name, to make things clearer

fedarko:
I think changes should be good to merge in whenever, pending @ElDeveloper's approval of the two open comments. thanks!

empress/support_files/js/empress.js
ElDeveloper: I hadn't considered overlap between layers. That makes sense.

ElDeveloper:
Looks great, thanks both!

Issue #391, opened 2020-09-24
Add warning to length legends if  missing / non-numeric values


fedarko:
Similar to #362 -- this is handled for continuous color legends, just gotta extend it to here.

Issue #303, opened 2020-08-05
SVG Export Improvements


ElDeveloper:
- [ ] Add support to only export the tree that's currently visible on screen.
- [ ] Change the dimensions of the page in the Python code (currently 4020x4020) which makes the size of the SVG document far too large.
- [ ] Add support to export barplots, and their legends (once #293 is merged).
- [ ] Add support to export collapsed clades.

Anything else you can think of @sjanssen2?

sjanssen2:
1. we might want to think about the dimension and position of the legends. They are currently positioned at (0,0) and have a fixed size (relative to one hard coded variable). It might be more elegant to re-size according to drawing area.
2. we discussed to replace the 15 line arc by one real arc for the circular layout, however, this would mean some extra math that can be messed up + intelligent traversal to differentiate between arc lines and "normal" edges
3. shall we aim to write node names into SVG tags or would that bloat the file too much?

ElDeveloper:
Regarding labeling nodes, maybe we would want to think about showing labels for some nodes and just displaying those instead. But that sounds like something outside of the scope of SVG export 😄

fedarko:
a few things we should fix at some point:

- [ ] Rather than checking node color to determine whether or not to thicken the line of certain nodes, we should check the `isColored` property of the corresponding node. This is because it's totally possible for a node to get assigned the exact default color but still be technically "colored" (I doubt this is the case with any of the current color maps, but it's still a possibility ...)
  - I understand that currently `exportSvg()` works on a line-by-line basis (so it doesn't "know" which lines are associated with which nodes), but it should be possible to make this work by replicating some of the work done in `getCoords()` (and/or having `getCoords()` return some sort of structure mapping nodes to line positions in `coords` or something).

- [ ] Line width scaling changed with #275, and I forgot to update the SVG export code (sorry). We should fix this -- although a challenge is that the SVG stroke width isn't really comparable to the line widths the code currently uses (i.e. for us the default line width is now `0`, but setting a stroke width of `0` will probably cause nothing to be drawn).

Since both of these seem like fairly annoying issues to resolve, it may be reasonable to just not alter line thickness when exporting SVG images of the tree for the time being.

fedarko:
- [ ] Configurable dimensions in the exported image (e.g. fit to letter paper)

PR #392, opened 2020-09-24
Support exporting all barplot legends, including color gradient and length legends; export collapsed clade shapes and barplots

### LINKED ISSUE #421: Gradient SVGs are slightly inaccurate
This is the same error as highlighted in https://github.com/biocore/emperor/issues/788. Just wanted to reference it here so we don't forget to solve it in both repos.

fedarko:
Part of the work on #303.

Closes #421.

Required some substantial refactoring to the way gradient SVGs are stored in Empress, in order to make it easier to scale these differently within a legend.

Also as part of this PR, slightly changed the way numbers are formatted in the Tree Statistics (so now commas are used as thousands separators [at least for English locales], etc.)

TODOs:

- [x] Add more vertical padding at the end of gradient and length legends (currently the text is flush with the surrounding border, which looks a bit too close for comfort)
- [x] Show a warning (for both gradient ~~and length legends [#391]~~) if there are missing / non-numeric values in the field. Currently this warning (which is already shown for gradient legends in the interface) is not included.
- [x] Refactor legend exporting code so that there isn't a notion of "rows" any more, so each legend can arbitrarily increase its vertical space without messing up other legends (all that should be kept track of are the height and width of each legend).
- [x] Unbreak tests (the colorer and legend tests, and maybe more, are extremely broken by these changes)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F392%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F392%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F392%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F392%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F392%2Fplain.qzv)

fedarko:
These changes are not very well covered by tests, but as a initial version that supports exporting I think this may be ok.

GIF showing how this looks --

combo

ElDeveloper:
That is looking super cool!

empress/support_files/js/legend.js
```diff
@@ -269,167 +321,477 @@ define(["jquery", "underscore", "util"], function ($, _, util) {
...
+        // (... For now, at least -- see
+        // https://github.com/biocore/emperor/issues/782.)
+        innerSVG +=
+            '<text x="' +
+            textLeftX +
+            '" y="' +
+            gradientTopY +
+            '" dominant-baseline="hanging">' +
+            this._maxValStr +
+            "</text>\n";
```
kwcantrell, line 509: It looks like adding text, it might make things easier to manage in the future if we abstract this into a helper function.

kwcantrell (commented):
Thanks @fedarko! I just have a couple of comments.

empress/support_files/js/export-util.js
```diff
@@ -146,17 +336,21 @@ define(["underscore", "chroma"], function (_, chroma) {
...
             // the exported image very slightly larger -- not a huge deal. Best
             // to be safe.)
-            minX -= radius;
-            minY -= radius;
-            maxX += radius;
-            maxY += radius;
+            bb.minX -= radius;
+            bb.minY -= radius;
+            bb.maxX += radius;
+            bb.maxY += radius;
```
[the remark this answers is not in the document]
fedarko: I added a separate function to expand the bounding box on all sides by a constant distance (`_expandBoundingBox`). I think this should address the problem, lmk what you think -- I agree that moving this work to separate functions would be useful, but I think explicitly making this its own operation will be clearer than co-opting `_updateBoundingBox` into doing it for us.

empress/support_files/js/legend.js
fedarko: That's fair -- it would be nice to reduce the size of the text code here. I think I might open a new issue for this if it's ok with you, since I think refactoring things to use a helper function for all SVG text stuff will take some work.

empress/support_files/js/empress.js
```diff
@@ -324,11 +324,12 @@ define([
          * @type{Array}
          * @private
          *
-         * Stores the vertex information that is passed to WebGl
+         * Stores the vertex information that is passed to WebGL
          *
          * Format: [x, y, r, g, b, ...]
          */
         this._collapsedCladeBuffer = [];
+        this._barplotBuffer = [];
```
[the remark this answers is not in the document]
fedarko: Makes sense. I refactored things a bit so that the barplot coordinates are returned by a separate function, and the SVG exporter will call that function to get the coordinates.

This might increase the time taken to export the SVG by a bit, since now we'll have to re-compute the barplot buffer (e.g. doing color and length assignment/scaling) every time we export SVG -- but I agree that the space savings is worth it.

fedarko:
Thanks for the feedback @kwcantrell! I think all of your comments should be addressed now.

I also realized that this PR introduced a bug where gradient legends (when the color reversing was selected) weren't getting reversed -- this was due to an error in how this PR used the chroma.js `.colors()` method. The problem should be fixed and tested now.

empress/support_files/js/export-util.js
```diff
@@ -5,28 +5,209 @@ define(["underscore", "chroma"], function (_, chroma) {
...
+     * elements), return an RGB triplet representation of the color at
+     * this position.
+     *
+     * @param {Array} coords
+     * @param {Number} i
+     * @return {String}
+     */
+    function _getRGB(coords, i) {
+        return chroma.gl(coords[i + 2], coords[i + 3], coords[i + 4]).css();
+    }
```
kwcantrell, line 55: For reference, this is how I modified this function to handle the color compression.
```
getRGB(color, i) {
    var c = _.map(Colorer.unpackColor(color[i]), function (rgb) {
        return rgb / 255;
    });
    return chroma.gl(...c).css();
};
```

empress/support_files/js/export-util.js
```diff
@@ -5,28 +5,209 @@ define(["underscore", "chroma"], function (_, chroma) {
...
+     * @param {Object} boundingBox Contains four entries: minX, maxX, minY,
+     *                             maxY. These define the current bounding box
+     *                             of the SVG. If any of the points in the
+     *                             polygons to be drawn extend outside of this
+     *                             box, the box will be updated -- in any case,
+     *                             the "new" (potentially unchanged) bounding
+     *                             box will be returned by this function.
+     * @param {Array} coords Array of coordinates to represent as polygons.
+     *                       This will probably be formatted like
+     *                       [x, y, r, g, b, x, y, r, g, b, ...].
```
kwcantrell, line 124: This will need to be updated once #400 is merged

empress/support_files/js/export-util.js
```diff
@@ -5,28 +5,209 @@ define(["underscore", "chroma"], function (_, chroma) {
...
+     * @param {Number} vertexSize The number of elements in coords for each
+     *                            point. This defaults to 5, i.e. x,y,r,g,b.
```
kwcantrell, line 126: same as above

kwcantrell (commented):
Looks good, I just have a few (minor) comments.

empress/support_files/js/export-util.js
fedarko: This will be really useful when merging this in after #400, thanks!

empress/support_files/js/colorer.js
```diff
@@ -52,18 +52,38 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
...
-        // input color map is sequential / diverging)
-        this._gradientSVG = null;
+        /*** Continuous-scaling-specific things ***/
 
+        // We append a number to the gradient ID so that multiple gradients can
+        // be present on the same page without overriding each other. (It's the
+        // caller's responsibility to ensure that gradient ID suffixes are
+        // unique, at least across Colorers created for continuous scaling.)
         this._gradientIDSuffix = gradientIDSuffix;
+        this._gradientID = null;
```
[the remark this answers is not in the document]
fedarko: Good point -- I modified this so that just the gradient ID is kept. Things should be clearer now.

kwcantrell:
thanks @fedarko and @ElDeveloper

Issue #276, opened 2020-07-24
Make edge thickness consistent across layouts and visualizations


ElDeveloper:
When coloring the tree and changing the width of the lines the thickness will be dependent on the tree, and the layout. It would be nice if this wasn't the case.

See #144 and #135 for more discussion.

fedarko:
As more motivation to improve this -- the current method for scaling line widths uses the log of the number of leaves in the tree as a denominator in the equation. A problem with this is that if there's only 1 leaf in the tree (as in the silly corner-case where the tree is a straight line from the root to the single tip), then `log(1) = 0`, and none of the line widths show up as a result.

![image](https://user-images.githubusercontent.com/4177727/90089427-3ba4f580-dcd6-11ea-927c-1757fd674a2f.png)

Having the layouts computed in JS will make recording the scaling factors used a lot easier, which should be the way to fix this. Hopefully.

fedarko:
Also applies to:

- barplot lengths (I have a fork that addresses this, need to polish it and file a PR)
- node circles (ideally would be scaled with zoom level, and/or scaled to some optimal size within the SVG export)

PR #400, opened 2020-09-29
Improve draw

kwcantrell:
@antgonza if you have some spare time, can you test this pr on one of your large trees? thanks


antgonza:
Sure but could you fix the conflicts and the errors?

antgonza:
... or the plan is to test to make sure it works and then fix the rest?

kwcantrell:
I just wanted to make sure it works/see what kind of improvement it provides.

kwcantrell:
this compresses the color array for each vertex into a single number which reduces the vertex buffer by ~2/3

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F400%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F400%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F400%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F400%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F400%2Fplain.qzv)

empress/support_files/js/drawer.js
```diff
@@ -257,15 +270,78 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
+        c.bindBuffer(c.ARRAY_BUFFER, buffer);
+
+        c.vertexAttribPointer(
+            s.color,
+            COLOR_SIZE,
+            c.FLOAT,
+            c.FALSE,
+            COLOR_SIZE * Float32Array.BYTES_PER_ELEMENT,
+            COLOR_OFFSET * Float32Array.BYTES_PER_ELEMENT
+        );
```
fedarko, line 322: A lot of this stuff is reused (albeit with slight differences in the arguments to `c.vertexAttribPointer()`, I guess) between the three `bind*Buffer()` functions. Would it be possible to abstract some of this code to a separate function (taking parameters as needed) that these three functions can call?

empress/support_files/js/drawer.js
```diff
@@ -257,15 +270,78 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
         );
     };
 
+    /**
+     * Binds the buffer so WebGL can use it.
```
fedarko, line 274:
```suggestion
     * Binds the tree coordinate buffer so WebGL can use it.
```

empress/support_files/js/drawer.js
```diff
@@ -257,15 +270,78 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
+            COORD_SIZE,
+            c.FLOAT,
+            c.FALSE,
+            COORD_SIZE * Float32Array.BYTES_PER_ELEMENT,
+            COORD_OFFSET
+        );
+    };
+
+    /**
+     * Binds the buffer so WebGL can use it.
```
fedarko, line 300:
```suggestion
     * Binds the tree color buffer so WebGL can use it.
```

fedarko (changes requested):
This looks cool, thanks for getting this ready!

Read over things, and I have a few suggestions. I think the most important are:

- The structure of the `getTreeCoords` / `getTreeColor` functions in empress.js -- would it be possible to merge these back into a single function, maybe by adding a parameter that controls whether it returns colors or coordinates? I think duplicating this logic will cause some problems (although I understand that it's already kinda duplicated between `getCoords` and `thickenColoredNodes` now, though).

- Use of 256 as the "base" for the compressed colors -- since hex colors range from [0, 255], I think we should use 255 instead.

Also, I am not sure what is causing it, but it looks like there is a strange bug when turning off collapsed clades where the colors become broken and some nodes remain hidden. My guess is that this might be a problem with `getTreeCoords` and `getTreeColor` getting out of sync, but I'm not confident about that.

buug

empress/support_files/js/drawer.js
```diff
@@ -5,13 +5,22 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
-        "attribute vec3 color;",
+        "attribute float color;",
         "varying vec3 c;",
         "uniform float pointSize;",
         "",
+        "vec3 unpackColor(float f) {",
+        "  vec3 color;",
+        "  color.b = floor(f / 256.0 / 256.0);",
+        "  color.g = floor((f - color.b * 256.0 * 256.0) / 256.0);",
+        "  color.r = floor(f - color.b * 256.0 * 256.0 - color.g * 256.0);",
```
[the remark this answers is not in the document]
kwcantrell: thanks, this method is much clearer.
 see my comment above regarding the 256 issue.

empress/support_files/js/empress.js
```diff
@@ -487,135 +488,21 @@ define([
...
-        var numSamples = Math.floor(60 * Math.abs(totalAngle / Math.PI));
-        return numSamples >= 2 ? numSamples : 2;
-    };
-
-    /**
-     * Retrieves the coordinate info of the tree.
-     *  format of coordinate info: [x, y, red, green, blue, ...]
+     * We used to interlace the coorinate information with the color information
+     * i.e. [x1, y1, red1, green1, blue1, x2, y2, red2, green2, blue2,...]
+     * However, for large trees
```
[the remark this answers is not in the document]
kwcantrell: thanks for catch this

empress/support_files/js/empress.js
```diff
@@ -487,135 +488,21 @@ define([
...
 
-        // branch color
-        var color;
-
-        var coords_index = 0;
-
         var addPoint = function (x, y) {
-            coords.push(x, y, ...color);
+            coords[coords.length] = x;
+            coords[coords.length] = y;
```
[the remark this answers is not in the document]
kwcantrell: I did created a brief test and the .length method was a bit faster (although not by much). However, I agree that .push is more clear. Also, we use .push throughout empress so I will revert this back to .push.

We can make a note that adding elements to arrays via .length can be implemented as a micro optimizer (i.e. we should only worry about it once all other optimizations have been done)

tests/test-colorer.js
```diff
@@ -202,20 +202,15 @@ require([
             c = new Colorer("Dark2", eles);
```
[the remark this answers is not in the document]
kwcantrell: test added

empress/support_files/js/colorer.js
```diff
@@ -313,6 +313,9 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         return _.mapObject(this.__valueToColor, Colorer.hex2RGB);
     };
 
+    Colorer.rgbToFloat = function (rgb) {
+        return rgb[0] + rgb[1] * 256 + rgb[2] * 256 * 256;
```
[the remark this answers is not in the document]
fedarko: Dang. You're totally right -- I just spent some time playing around with this and yeah, the base needs to be 1 larger than the largest value in order for this to work. Sorry for doubting you! Thanks for pushing back on this.

tests/test-circular-layout-computation.js
```diff
@@ -203,31 +203,29 @@ require(["jquery", "BPTree", "BiomTable", "Empress"], function (
...
 
             // For the arc for node 3 start at (2,0) and ends at (-2, 0)
             // check if arc for node 3 is correct
-            ok(Math.abs(coords[30] - 2) < 1.0e-15); // start x arc position
-            ok(Math.abs(coords[31 - 0]) < 1.0e-15); //start y arc position
-            // prettier-ignore
-            ok(Math.abs(coords[625] - (-2)) < 1.0e-15); // end x arc position
-            ok(Math.abs(coords[626] - 0 < 1.0e-15)); // end y arc position
+            ok(Math.abs(coords[12] - 2) < 1.0e-15); // start x arc position
+            ok(Math.abs(coords[13]) < 1.0e-15); //start y arc position
```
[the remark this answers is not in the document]
kwcantrell: thanks, I forgot to add it back.

kwcantrell:
Thanks @fedarko for the feedback. I believe all of your comments have been addressed. 

A couple of things:
1) I fixed the collapse clades issue you pointed out (good catch)
2) I also fixed the export svg function

> The structure of the getTreeCoords / getTreeColor functions in empress.js -- would it be possible to merge these back into a single function, maybe by adding a parameter that controls whether it returns colors or coordinates? I think duplicating this logic will cause some problems (although I understand that it's already kinda duplicated between getCoords and thickenColoredNodes now, though).

I think this should get moved to a different PR. We already have multiple functions that use the same pattern so, creating a single function will require a bit of restructuring that I is a bit outside of the scope of this PR.

fedarko:
Going through now, changes look great. It looks like the clade collapsing is mostly fixed, but I think there's still some issues with collapsing and the rectangular / circular layouts. It looks like turning off collapsing doesn't re-draw the arcs in the circular layout or the vertical lines in the rectangular layout, and causes some coloring problems:

Circular layout:
hm

Rectangular layout:
hm2

It also looks like turning off coloring in general doesn't fix this, either (the arcs/lines only seem to come back when I uncheck "collapse clades"):
hm3

I think the issue may have to do with `resetTree()`, and that `getTreeCoords()` is called before `this._collapsedClades` is reset (which causes the vertical lines / arcs to not be drawn since they're still included in the now-removed collapsed clades). But I'm not 100% sure -- something similar also might be what's messing with the colors not being correct.

kwcantrell:
good catch @fedarko!

empress/support_files/js/drawer.js
```diff
@@ -222,50 +239,74 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
+    };
+
+    /**
+     * Fills the buffer used to draw the tree
+     *
+     * @param {Array} data The coordinate and color data to fill tree buffer
+     */
+    Drawer.prototype.loadTreeCoordsBuff = function (data) {
+        data = new Float32Array(data);
+        this.treeCoordSize = data.length / 2;
```
fedarko, line 297: Just to check -- since the buffer is set up with multiples of 3 elements (`[x, y, rgb, x, y, rgb, ...]`), is there a danger of the length of the buffer being an odd number (causing `this.treeCoordSize` to be a float)?

(If the buffer is just `[x, y, x, y, ...]`  then I think the parameter description for this function should be updated, since it implies that color is also included in the buffer.)

empress/support_files/js/empress.js
```diff
@@ -79,7 +79,7 @@ define([
          * @type {Array}
          * The default color of the tree
          */
-        this.DEFAULT_COLOR = [0.25, 0.25, 0.25];
+        this.DEFAULT_COLOR = Colorer.rgbToFloat([50, 50, 50]);
```
fedarko, line 82: I think the previous color was equivalent to (multiplying .25 by 256) [64, 64, 64] instead -- so this will be a slightly darker color. Is this intentional?

tests/test-legend.js
```diff
@@ -102,6 +102,10 @@ require(["jquery", "chroma", "UtilitiesForTesting", "Legend"], function (
                     // "rgb(255, 0, 0)". Fortunately, chroma.js recognizes this, so
                     // we can just chuck both colors to compare through chroma
                     // before checking equality.
+                    // Note: firefox breaks on this test. chroma.js does not
+                    // recognize "rgb(255, 0, 0" on firefox. This is not an
+                    // issue for empress as we would not pass "rgb(r, g, b)" to
+                    // chroma.js
```
fedarko, line 108: :O That's wack! Thanks so much for pointing this out. I'll open an issue.

empress/support_files/js/export-util.js
```diff
@@ -119,7 +123,7 @@ define(["underscore", "chroma"], function (_, chroma) {
 
         // create a circle for each node
         if (drawer.showTreeNodes) {
-            radius = drawer.NODE_CIRCLE_DIAMETER / 2;
+            var radius = drawer.NODE_CIRCLE_DIAMETER;
```
fedarko, line 126: I agree that the `var` is needed :), but I think we should still divide this by 2 right?

... I guess it doesn't matter too much since the "units" are kind of arbitrary, but idk.

empress/support_files/js/select-node-menu.js
```diff
@@ -491,7 +491,7 @@ define(["underscore", "util"], function (_, util) {
             var node = nodeKeys[i];
             var x = this.empress.getX(node);
             var y = this.empress.getY(node);
-            highlightedNodes.push(...[x, y, 0, 1, 0]);
+            highlightedNodes.push(...[x, y, 65280]);
```
fedarko, line 494:
```suggestion
            // Add a green circle indicating the highlighted node(s)
            highlightedNodes.push(...[x, y, 65280]);
```

(I figure it would make sense to explain that this color is green, since now it's harder to look at a color in the code and see what color it represents)

tests/test-circular-layout-computation.js
```diff
@@ -203,31 +203,34 @@ require(["jquery", "BPTree", "BiomTable", "Empress"], function (
...
+            // the arc for node 3 spans PI radians thus, the number of lines to
+            // approximate arc are 60*((radians spanned by arc) / PI) = 60
+            // the above calculation can be found at
+            // Empress.__numSampToApproximate()
```
fedarko, line 231:
```suggestion
            // Empress._numSampToApproximate()
```

fedarko (changes requested):
I haven't gone through all of the export util / empress stuff yet, but have gone through about everything else. Have a few mostly small comments. This is looking great, thanks so much!!!

empress/support_files/js/drawer.js
```diff
@@ -222,50 +239,74 @@ define(["glMatrix", "Camera"], function (gl, Camera) {
...
      *
      * @param {WebGLBuffer} buffer The Buffer to bind
+     * @param {Number} buffType The type of buffer to bind
+     * @param {Number} vertSize The size of the vertex for webgl
      */
-    Drawer.prototype.bindBuffer = function (buffer) {
-        // defines constants for a vertex. A vertex is the form [x, y, r, g, b]
-        const COORD_SIZE = 2;
+    Drawer.prototype.bindBuffer = function (buffer, buffType, vertSize) {
+        if (this.BUFF_TYPES.indexOf(buffType) == -1) {
```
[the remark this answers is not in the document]
kwcantrell: I agree, this is a much readable way.

empress/support_files/js/drawer.js
kwcantrell: This buffer is just the tree coords. I'll add a note in the doc string describing the structure of the buffer.

empress/support_files/js/empress.js
kwcantrell: I chose 50 mainly because it was a nice round number. (previously equivalent to 255*.25=63.75 since colors are in the range [0, 255]). 

The color is slightly darker. I don't mind setting it to 64 though. 
| [50, 50, 50,] | [64, 64, 64] |
|----------------- |-----------------|
|Screenshot from 2020-10-27 21-25-02 | Screenshot from 2020-10-27 21-24-43

empress/support_files/js/export-util.js
kwcantrell: The circles were not really visible when we divide by 2 (at least on my monitor with the moving pictures data set)

Linked are two svg. One that divides the diameter by 2 and one that does not. Let me know which one looks better - it could just be my monitor :).

[examples.zip](https://github.com/biocore/empress/files/5449679/examples.zip)

empress/support_files/js/empress.js
fedarko: Thanks for the explanation. My opinion is that we should probably just leave the default color at ~64 for now (if we want to update the default color it might make sense to knock that out in a separate PR or something, once these and the SVG changes are in), but ... I didn't notice that the default color was changed until I read through the code, so I don't think it will make a huge difference either way ;)

empress/support_files/js/export-util.js
fedarko: Thanks for the examples! It's subjective, but I think keeping the division by 2 looks slightly better. It does make it harder to see node circles when zoomed out, but things look less cluttered when the SVGs are zoomed in (... at least on my computer :) --

_with division by 2:_
image

_with no division by 2:_
image

Neither solution is great, honestly, since the size of node circles isn't proportional to the tree size or anything. We should ideally try to scale node circles based on the tree dimensions, similarly to the stuff for edge thicknesses described in #276.

fedarko (changes requested):
Thanks, this looks good! (And sorry for taking a few days to get back to you on this.) The bulk of this is basically ready as is, but there are some documentation / internal code things I think we should sort out first.

Besides the comments I made here, there are a few places with old documentation saying "RGB arrays" instead of RGB floats, including:

Part of `Empress.colorSampleGroups()`:

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/empress.js#L1926-L1927

Collapsed clade data:

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/empress.js#L310

Part of `Empress._colorTree()`:

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/empress.js#L2206-L2211

Part of `BarplotLayer.initFMDiv()`:

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/barplot-layer.js#L246-L250

The border color stuff in the barplot panel handler:

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/barplot-panel-handler.js#L139-L140

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/barplot-panel-handler.js#L267-L284

https://github.com/biocore/empress/blob/aaac6e316ee6b2b5453f528d38780bd2de67f65a/empress/support_files/js/empress.js#L1840-L1855

I haven't gone extremely thoroughly through things yet, so there may be additional places throughout the code/docs which exhibit this problem.

None of these are really _urgent_ issues, since only one of them impacts the user directly, but I would like to try to fix as many of them as we can in advance (to avoid the master codebase having confusing docs).

For what it's worth the barplot panel handler stuff does very slightly impact the program behavior -- the "freebie" checking stuff should be changed to check if the border color is equal to exactly white (as a float), rather than the current behavior which still assumes that the color is stored as an RGB array. The consequence of this right now is that "freebies" are detected whenever the border color is unchanged (since it defaults to [1,1,1]), but even if the user changes the color and then changes it back to exactly white (16777215), no "freebie" is given since the checks still assume this is an array. This makes drawing barplot borders slower... albeit in this one, extremely tiny corner case :P

Anyway, hope this clarifies things. Happy to zoom sometime this week and go through these, or to just defer some of this work to a new issue if you'd prefer (since we have a lot of open PRs to get through). Thanks so much!

PR #410, opened 2020-10-06
ENH: Enlarge selected samples in Emperor

### LINKED ISSUE #373: Support alternate node -> sample highlighting methods in Emperor?
Came up during discussion about paper figures. The current method of using `setEmissive()` to highlight the samples containing a tip is more visible on black than white backgrounds in Emperor, and it'd be good to support methods that work consistently regardless of the color.

Fortunately Emperor's API makes playing around with the various options here very easy :) One option would be replacing the `setEmissive()` calls with `setScale()` calls to enlarge the highlighted samples (here it's using a value of `3` for `setScale()` but this is ofc configurable) --

comb

Not sure how far down the rabbit hole we want to go with this, but we could add a section under `Tree Properties` or something that'd control what adjustments are used.

fedarko:
Closes #373. Ideally we'd have options for choosing how to visually represent selected samples, but I think enlarging is a more robust solution than the emissive stuff.

Emissives are still used for representing samples selected in Emperor using the dragging options -- the scaling stuff is only used for representing samples that contain a given selected node/nodes in Empress. I think having these things be split up looks nice, personally.

There are still the Emperor scale controls available, which would mess with things here (allowing for all sorts of interaction bugs -- e.g. the user re-scales the samples or something in Emperor only for that re-scaling to get lost when the user selects/unselects nodes in Empress), but honestly we could just hide that part of the Emperor menu and I think that'd be good enough for now...?

fancy

### DIFF
--- a/empress/support_files/js/emperor-callbacks.js
+++ b/empress/support_files/js/emperor-callbacks.js
@@ -5,18 +5,18 @@
  */
 
 empress.setOnNodeMenuVisibleCallback(function (samples) {
-    // reset emissive settings for all markers
-    ec.decViews.scatter.setEmissive(0x000000);
+    // reset scale settings for all samples
+    ec.decViews.scatter.setScale(1);
 
     // retrieve the plotting objects
     samples = ec.decModels.models.scatter.getPlottableByIDs(samples);
-    ec.decViews.scatter.setEmissive(0x8c8c8f, samples);
+    ec.decViews.scatter.setScale(2, samples);
     ec.sceneViews[0].needsUpdate = true;
 });
 
 empress.setOnNodeMenuHiddenCallback(function (samples) {
     samples = ec.decModels.models.scatter.getPlottableByIDs(samples);
-    ec.decViews.scatter.setEmissive(0x000000, samples);
+    ec.decViews.scatter.setScale(1, samples);
     ec.sceneViews[0].needsUpdate = true;
 });
 
@@ -171,6 +171,7 @@ ec.controllers.color.addEventListener("value-double-clicked", function (
     // reset emissive settings for all markers since an ongoing timer may have
     // been cancelled
     ec.decViews.scatter.setEmissive(0x000000);
+    ec.decViews.scatter.setScale(1);
     plotView.needsUpdate = true;
 
     // if there's any coloring setup remove it, and re-enable the update button
--- a/empress/support_files/js/emperor-callbacks.js
+++ b/empress/support_files/js/emperor-callbacks.js
@@ -10,10 +10,9 @@ empress.setOnNodeMenuVisibleCallback(function (samples) {
 
     // retrieve the plotting objects
     samples = ec.decModels.models.scatter.getPlottableByIDs(samples);
-    if(ec.UIState['view.usesPointCloud']) {
+    if (ec.UIState["view.usesPointCloud"]) {
         ec.decViews.scatter.setScale(10, samples);
-    }
-    else {
+    } else {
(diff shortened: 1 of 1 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F410%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F410%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F410%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F410%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F410%2Fplain.qzv)

empress/support_files/js/emperor-callbacks.js
```diff
@@ -5,18 +5,18 @@
  */
 
 empress.setOnNodeMenuVisibleCallback(function (samples) {
-    // reset emissive settings for all markers
-    ec.decViews.scatter.setEmissive(0x000000);
+    // reset scale settings for all samples
+    ec.decViews.scatter.setScale(1);
```
kwcantrell, line 9: I'm not sure if we documented this or not but this will override user defined node sizes. For example if the user either set the global node size or scaled nodes based on sample metadata then this will remove that. We had the same issue before but will the emissive property.

kwcantrell (approved):
Very cool @fedarko! I just have one comment. Its not something you need to worry about fixing in this pr because we had a similar issue before.

empress/support_files/js/emperor-callbacks.js
```diff
@@ -5,18 +5,18 @@
...
 
     // retrieve the plotting objects
     samples = ec.decModels.models.scatter.getPlottableByIDs(samples);
-    ec.decViews.scatter.setEmissive(0x8c8c8f, samples);
+    ec.decViews.scatter.setScale(2, samples);
```
ElDeveloper, line 13: Let's change this constant when emperor runs using a point cloud, something like:
```suggestion
    if(ec.UIState['view.usesPointCloud']) {
        ec.decViews.scatter.setScale(10, samples);
    }
    else {
        ec.decViews.scatter.setScale(2, samples);
    }
```

ElDeveloper (changes requested):

empress/support_files/js/emperor-callbacks.js
fedarko: Yeah, this will be a problem -- we should definitely document it, at least. I think just hiding the Scale tab in Emperor would be a good solution for now. @ElDeveloper is there a way to do this from our side of things using Emperor's API, or do you think this'd require making a PR to Emperor that supports this?

ElDeveloper: I think we can document this as a known bug.

ElDeveloper: That is rather than disabling the tab altogether.

Issue #418, opened 2020-10-13
Support "flipping" continuous color maps


fedarko:
So that e.g. viridis could go from yellow to purple rather than purple to yellow. This would add some more variety to the color options present in Empress.

The changes here would mostly involve the Colorer.js module within Empress, as well as some additional changes to the HTML to involve adding a UI element (checkbox?) that allows users to toggle whether or not the color map is reversed. I _think_ this might be as simple as changing `[min, max]` to `[max, min]` within this block of code:

https://github.com/biocore/empress/blob/a76ebf27a8c3730c60a8c5721aa899f2d3ff2e13/empress/support_files/js/colorer.js#L200-L206

This seems to be how the Chroma.js docs recommend doing this, at least -- see the section around [here](https://gka.github.io/chroma.js/#scale-nodata), right around where it says "To reverse the colors ..."

For reference: brought up by @kwcantrell this morning. We talked about reversing color maps in the context of Qurro a while back at https://github.com/biocore/qurro/issues/277, but I haven't got around to doing that yet :(

gwarmstrong:
Can I grab this one?

kwcantrell:
thanks @gwarmstrong! This will a nice feature to have.

fedarko:
(Closed by #423.)

Issue #405, opened 2020-10-01
Support coloring the tree by continuous feature metadata


fedarko:
I thought there was an open issue for this, but can't find it. Anyway, this would be useful to have: each tip (with metadata, at least) would get assigned a color based on its numeric value for a given feature metadata field, with non-numeric / missing values remaining uncolored (or colored a specific "non-numeric" color, or hidden, I guess?)

Colors could be "propagated" up the tree from the tips by doing something like averaging all child nodes' values -- this should create a pretty gradient effect and (hopefully) be a meaningful way of representing this. (Could also support e.g. taking the median of child values, and there are probably lots of different ways of representing the "gradient" coloring.)

Inspired by [this section of the ggtree manual](http://yulab-smu.top/treedata-book/chapter4.html#color-tree) which has a lot of pretty examples of trees colored using continuous data.

A lot of this should involve code reusable from continuous barplots, so hopefully this wouldn't be too much of a pain to add.

Issue #353, opened 2020-08-25
Support drawing quantitative sample metadata barplots


fedarko:
This would work by, for a given sample metadata field (e.g. pH, age, ...), computing (for each tip) the average/median/etc. value of this field out of all samples that contain this tip: then this information could be encoded either using color or length.

Initially mentioned by @ElDeveloper in #201. This could be useful for the paper, as a way of showing which tips are "associated" with which types of samples.

PR #423, opened 2020-10-14
NEW Add reverse argument for Colorer

gwarmstrong:
#418 

This PR adds functionality to Colorer to support reversing color maps. I have also integrated the reverse with the feature metadata coloring UI element.

You can see the feature in action [here](https://drive.google.com/file/d/1rGFY7TVp4eM9XoPmtvsXszomlkbFZfto/view?usp=sharing).

Notes/questions:

Custom color maps are not supported (my understanding is that the user supplies an object that performs the mapping, which is unordered, so cannot be “reversed”)

Are there other places where you would like me to implement the UI elements for reversing the color map?

Tests:

[x] Colorer with reverse=true works with discrete color map

[x] Colorer with reverse=true works with continuous color map

[x] Colorer with reverse=true works with ordinal color map

[x] Emperor.colorByFeatureMetadata switches the color values when reverse=true

### DIFF
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -124,7 +124,7 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
             palette = chroma.brewer[this.color];
         }
         if (this.reverse) {
-            palette = _.toArray(palette);
+            palette = _.clone(palette);
             palette.reverse();
         }
         for (var i = 0; i < this.sortedUniqueValues.length; i++) {
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -224,7 +224,6 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             domain = [min, max];
         }
-        console.log("domain: ", domain);
         var interpolator = chroma.scale(this.color).domain(domain);
         _.each(split.numeric, function (n) {
             scope.__valueToColor[n] = interpolator(parseFloat(n));
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -123,17 +123,13 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             palette = chroma.brewer[this.color];
         }
-        const n_unique = this.sortedUniqueValues.length;
-        var colorIndex;
-        for (var i = 0; i < n_unique; i++) {
-            if (this.reverse) {
-                colorIndex = n_unique - i - 1;
-            } else {
-                colorIndex = i;
-            }
+        if (this.reverse) {
+            palette = [...palette];
+            palette.reverse();
+        }
+        for (var i = 0; i < this.sortedUniqueValues.length; i++) {
             var modIndex = i % palette.length;
-            this.__valueToColor[this.sortedUniqueValues[colorIndex]] =
-                palette[modIndex];
+            this.__valueToColor[this.sortedUniqueValues[i]] = palette[modIndex];
         }
     };
 
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -124,7 +124,7 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
             palette = chroma.brewer[this.color];
         }
         if (this.reverse) {
-            palette = [...palette];
+            palette = _.toArray(palette);
             palette.reverse();
         }
         for (var i = 0; i < this.sortedUniqueValues.length; i++) {
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -160,7 +160,12 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
             // If there's only 1 unique value, set its color as the first in
             // the color map. This matches the behavior of Emperor.
             var onlyVal = this.sortedUniqueValues[0];
-            this.__valueToColor[onlyVal] = chroma.brewer[this.color][0];
+            var brewer = chroma.brewer[this.color];
+            if (this.reverse) {
+                this.__valueToColor[onlyVal] = brewer[brewer.length - 1];
+            } else {
+                this.__valueToColor[onlyVal] = brewer[0];
+            }
         } else {
             // ... Otherwise, do normal interpolation -- the first value gets
             // the "first" color in the color map, the last value gets the
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -28,6 +28,9 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
      *                                 color map is discrete and/or
      *                                 useQuantScale is false, then this will
      *                                 be ignored.)
+     * @param{Boolean} reverse Defaults to false. If true, the color scale
+     *                         will be reversed, with respect to its default
+     *                         orientation.
      * @return{Colorer}
      * constructs Colorer
      */
@@ -35,7 +38,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         color,
         values,
         useQuantScale = false,
-        gradientIDSuffix = 0
+        gradientIDSuffix = 0,
+        reverse = false
     ) {
         var scope = this;
 
@@ -43,6 +47,7 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         this.sortedUniqueValues = util.naturalSort(_.uniq(values));
 
         this.color = color;
+        this.reverse = reverse;
 
         // This object will describe a mapping of unique field values to colors
         this.__valueToColor = {};
@@ -118,9 +123,17 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             palette = chroma.brewer[this.color];
         }
-        for (var i = 0; i < this.sortedUniqueValues.length; i++) {
+        const n_unique = this.sortedUniqueValues.length;
+        var colorIndex;
+        for (var i = 0; i < n_unique; i++) {
+            if (this.reverse) {
+                colorIndex = n_unique - i - 1;
+            } else {
+                colorIndex = i;
+            }
             var modIndex = i % palette.length;
-            this.__valueToColor[this.sortedUniqueValues[i]] = palette[modIndex];
+            this.__valueToColor[this.sortedUniqueValues[colorIndex]] =
+                palette[modIndex];
         }
     };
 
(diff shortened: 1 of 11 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F423%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F423%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F423%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F423%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F423%2Fplain.qzv)

fedarko:
This looks sweet! Haven't dug through the changes very thoroughly yet, but will try to get that done this week. Looks like things are really well-tested, so thanks for that :)

> Are there other places where you would like me to implement the UI elements for reversing the color map?

I guess this functionality would be useful pretty much everywhere in Empress that assigns colors, right? I think this would be most useful right now for continuous color maps (currently these are only used for feature metadata barplots, but soon they'll be used for other stuff -- #405, #353, ...). Past that, the other places in Empress that assign colors I can think of are:

  - Sample metadata coloring
  - Feature metadata barplots
  - Sample metadata barplots
  - Animations

There is one question I'd like to ask -- not sure what your thoughts on this are. When reversing a categorical (or discrete idk) color map for _N_ categories, there are two options I can think of:

1. Take the first _N_ colors from the color map and then reverse them (this is what this PR currently does)
2. Take the last _N_ colors from the color map (starting with the last color in the map, and then working backwards)

Option 2 seems like it would be more intuitive to me. One reason is that the current way this works means that the start color changes depending on how many categories there are, which could be confusing:

rev

...  whereas starting from the "end" of the color map (for `Classic QIIME Colors`, this is that weird shade of turquoise) means that the first color is consistently turquoise. Another benefit is that this might simplify some of the discrete color assignment code -- since we'd be able to just reverse the `palette` array and handle everything the same way from then on, I think.

Do you have a preference on how to handle this? I guess we could implement both and make "Reversing method" or something another option, but having an option for an option might be veering into "too much stuff" territory.

Thanks so much!!! :art:

kwcantrell:
Thanks @gwarmstrong this is a much needed feature!

I agree with @fedarko, reversing the colors would be nice to have for pretty much all coloring options (or at least barplots).

tests/test-empress.js
```diff
@@ -245,15 +245,15 @@ require([
...
+                "color 1 is last qiime color"
             );
-            deepEqual(
-                chroma(cm_copy["2"]).gl().slice(0, 3),
-                chroma(cm["1"]).gl().slice(0, 3),
-                "cm_copy['2'] vs. cm['1']"
+            equal(
+                chroma(cm["2"]).hex(),
+                "#808000",
+                "color 1 is second last qiime color"
```
gwarmstrong, line 256:
```suggestion
                "color 2 is second last qiime color"
```

gwarmstrong:
The build is green and I hooked up the color reversing to sample metadata, feature metadata barplots, sample metadata barplots, and animations! Should be ready for a review.

empress/support_files/js/colorer.js
```diff
@@ -118,6 +123,10 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             palette = chroma.brewer[this.color];
         }
+        if (this.reverse) {
+            palette = [...palette];
```
fedarko, line 127: I'm guessing this line is designed to convert the palette into a JS array so `.reverse()` can be used. If so, I think it might be nice to restate this [using underscore.js](https://underscorejs.org/#toArray) as

```suggestion
            palette = _.toArray(palette);
```

I doubt that this will make a difference efficiency-wise, but I think it should be more clear to readers.

empress/support_files/js/colorer.js
```diff
@@ -157,9 +166,14 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
             // the "first" color in the color map, the last value gets the
```
fedarko, line 166: (comment applies to the `if` block above this -- GitHub won't let me comment there)

For consistency's sake, I think it might make sense to assign the "last" color in the color map to single-unique-value inputs. So e.g. if `Pink-Yellow-Green` is selected as the color scheme and the reverse option is used, then Green should be the first color used in the legend, regardless of how many unique values there are.

empress/support_files/js/empress.js
```diff
@@ -1848,19 +1855,28 @@ define([
...
+     *                         will be reversed, with respect to its default
+     *                         orientation.
      *
      * @return {Object} If there exists at least one group with unique features
      *                  then an object will be returned that maps groups with
      *                  unique features to a color. If there doesn't exist a
      *                  group with unique features then null will be returned.
      */
-    Empress.prototype.colorBySampleCat = function (cat, color) {
+    Empress.prototype.colorBySampleCat = function (cat, color, reverse = true) {
```
fedarko, line 1867: I think the default for `reverse` should be false here, right? Judging by the application behavior and what the docs comment says.

tests/test-barplots.js
```diff
@@ -298,6 +298,7 @@ require([
         notOk(layer1.colorByFM);
         equal(layer1.colorByFMField, empress._barplotPanel.fmCols[0]);
         equal(layer1.colorByFMColorMap, "discrete-coloring-qiime");
+        equal(layer1.colorByFMColorReverse, false);
```
fedarko, line 301: (QUnit's syntax is kinda funky, but the surrounding tests use `notOk()` and `ok()` to assert false/true respectively so this should probs be consistent.)

```suggestion
        notOk(layer1.colorByFMColorReverse);
```

tests/test-barplots.js
```diff
@@ -310,6 +311,7 @@ require([
         // Check sample metadata barplot defaults
         equal(layer1.colorBySMField, empress._barplotPanel.smCols[0]);
         equal(layer1.colorBySMColorMap, "discrete-coloring-qiime");
+        equal(layer1.colorBySMColorReverse, false);
```
fedarko, line 314:
```suggestion
        notOk(layer1.colorBySMColorReverse);
```

tests/test-colorer.js
```diff
@@ -297,6 +327,24 @@ require([
             equal(hexmap["2"], "#b2df8a");
             equal(hexmap["100"], "#33a02c");
         });
+        test("Test discrete color map with reverse = true", function () {
+            // CVALDISCRETETEST
```
fedarko, line 331:
```suggestion
```
(the other use of this comment in this file was [to help with searching for a specific test](https://github.com/biocore/empress/blob/a76ebf27a8c3730c60a8c5721aa899f2d3ff2e13/empress/support_files/js/empress.js#L1578-L1584); doesn't apply here i think)

tests/test-colorer.js
```diff
@@ -127,14 +127,26 @@ require([
...
         test("Test construction with a div. color map, all numeric values, and useQuantScale = true", function () {
             var eles = ["5", "2", "3", "1", "4", "0", "-5"];
             var colorer = new Colorer("PiYG", eles, true);
 
-            // Expected colors determined by trying
-            // chroma.scale("PiYG").domain([-5,5])(n); where n = 0, 1, 2, etc.
-            // (see the interactive docs at https://gka.github.io/chroma.js/ --
-            // super useful for testing this)
+            // Expected colors determined by reversing
+            // the colors in the previous test
```
fedarko, line 149: Could you restore the old comment here? The resulting color assignments for these tests shouldn't be just reversed (aside from the extreme values), since one of them has `useQuantScale` set to true while the new one has that set to false.

tests/test-empress.js
```diff
@@ -204,6 +217,7 @@ require([
                 "discrete-coloring-qiime",
                 "tip"
             );
+            let cm_copy = cm;
```
fedarko, line 220:
```suggestion
```

doesn't look like this variable is used anywhere

fedarko (changes requested):
A couple small things, but this looks really nice! Thanks so much for putting this together.

empress/support_files/js/empress.js
gwarmstrong, line 1867:
```suggestion
    Empress.prototype.colorBySampleCat = function (cat, color, reverse = false) {
```

gwarmstrong: Good catch.

tests/test-colorer.js
gwarmstrong: Sorry, probably was not editing the right comment, careless on my end...

gwarmstrong, line 149:
```suggestion
            // Expected colors determined by trying
            // chroma.scale("PiYG").domain([-5,5])(n); where n = 0, 1, 2, etc.
            // (see the interactive docs at https://gka.github.io/chroma.js/ --
            // super useful for testing this)
```

tests/test-empress.js
gwarmstrong: Yeah carry-over from developing the test, my bad.

empress/support_files/js/colorer.js
fedarko: That makes sense, thanks for clarifying your intent! I think your understanding is correct; JS is a bit funky but Objects (including Arrays) are passed by reference, so it'd be possible for the `reverse` to impact the palette. Thanks for being so cautious about this.

Ok, in this case (sorry to ask for another change here) my previous suggestion was suboptimal -- I don't think `_.toArray()` makes the most sense for creating a copy, at least from the perspective of readability. Would it be possible to just replace `_.toArray()` with `_.clone()`? This way it is as clear as possible to readers _why_ this step is being taken. Alternately, I also wouldn't mind going back to the `...` syntax instead, provided we add a comment just linking to this thread or whatever (basically just explaining that this is a necessary step, in the hopes that one of us doesn't come back to this 10 years from now and go "hmm that line probably wasn't doing anything important anyway" ;).

image

tests/test-colorer.js
fedarko: No worries!

empress/support_files/js/colorer.js
```diff
@@ -118,6 +123,10 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             palette = chroma.brewer[this.color];
         }
+        if (this.reverse) {
+            palette = _.toArray(palette);
```
gwarmstrong, line 127:
```suggestion
            palette = _.clone(palette);
```

empress/support_files/js/colorer.js
gwarmstrong: Yeah the `_.toArray -> _.clone` change makes sense.

tests/test-colorer.js
```diff
@@ -306,6 +306,26 @@ require([
...
+            var colorer = new Colorer(
+                "Viridis",
+                ["abc"],
+                undefined,
+                undefined,
+                true
+            );
+            // The last value in the color map (for viridis, dark purple)
+            // should be used.
+            equal(colorer.__valueToColor.abc, "#fee825");
```
fedarko, line 319:
```suggestion
            // The last value in the color map (for viridis, yellow)
            // should be used.
            equal(colorer.__valueToColor.abc, "#fee825");
```

fedarko (changes requested):
just a quick fix, but everything looks good! Once this is in I think this is ready to merge.

fedarko:
Thanks @gwarmstrong! :art:

ElDeveloper:
Thanks @gwarmstrong! And thanks for reviewing @kwcantrell @fedarko! 🎉

Issue #419, opened 2020-10-13
Support shearing the tree to feature metadata in tree-plot


fedarko:
Currently, `--p-shear-tree` is only available for the `qiime empress community-plot` command (where it shears the tree to just the tips contained as features in the table).

@antgonza brought up that it would be nice to support shearing the tree to the feature metadata, for example when `qiime empress tree-plot` is passed a massive tree and slightly less massive feature metadata information.

Things we will need to consider for this issue:

- Should this option also be available for `community-plot`? If so, how will this interact with the existing shear-to-table option there? (We might need to rename that parameter to something more specific than `shear-tree`, at least, just to avoid confusing users.)

    - If both shearing options are available for `community-plot`, we'll definitely want to use an explicit "order" for how shearing is done. There are probably going to be a few annoying cases where applying one shearing method causes the other shearing method to fail, or something, so testing this will be important. (For example: if the feature metadata is only defined for tips _not_ present in the table, but both shearing methods are provided, then things should explode because no tips would remain. Or, if the only features shared by the feature metadata and table are all zeroes in the table, then this should still fail, because empty features / samples are filtered out before creating the visualization.)

- Feature metadata doesn't have to refer to tips -- it is possible to pass Empress a feature metadata file with only internal node metadata. I imagine that, in this case, shearing would only be done on the tips within the feature metadata, and if there are no tips in the feature metadata then an error should be raised.

- probably other stuff i'm forgetting

After writing this out, this seems like it might be kind of complex to implement alongside the existing shearing method... so maybe just doing this for `tree-plot` at first would be reasonable.

fedarko:
Closed by #428.

PR #428, opened 2020-10-16
Add option to shear tree-plot to feature metadata

gibsramen:
Relevant to #419. Currently implemented only in `tree-plot` not `community-plot`.

Adds the `--p-shear-to-feature-metadata` parameter to `tree-plot`. This parameter, if set to `True`, shears the tree to only include tips present in the feature metadata. If the feature metadata file includes only internal nodes an error is raised. By default this option is set to `False`.

EDIT: Also - will probably update `shear-tree` option to be `shear-to-table` unless anyone has any objections.


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F428%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F428%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F428%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F428%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F428%2Fplain.qzv)

tests/python/test_core.py
```diff
@@ -492,13 +492,74 @@ def test_fm_filtering_post_shearing(self):
...
+
+        assert_frame_equal(viz.tip_md, lonely_fm.loc[["a"]])
+        self.assertTrue(viz.int_md.empty)
+
+        # feature metadata should be unchanged and be a different id instance
+        assert_frame_equal(lonely_fm, viz.features)
+        self.assertNotEqual(id(lonely_fm), id(viz.features))
+
+        self.assertIsNone(viz.ordination)
+
```
fedarko, line 555: If you don't mind, there is one more case that I think we should explicitly test here—when the tree is sheared to a set of tips in the feature metadata, but the input feature metadata also includes metadata for an internal node that gets filtered out from the shearing.

I think the default `self.feature_metadata` satisfies this case -- since it just includes metadata for `a` (tip) and `h` (internal node), and `h` is not a parent of `a`. In this case, `h`'s metadata should be filtered out of the internal node metadata.

I'm pretty sure this sort of case is already handled correctly by this PR, since `match_tree_and_feature_metadata()` is called after the feature metadata shearing. However, I don't think (sorry if i'm missing it) that any of the tests added in this PR cover this exact case -- so we should probably verify that this scenario continues to work in the future, just because this sort of stuff is always finicky.

empress/plugin_setup.py
```diff
@@ -137,13 +137,19 @@
...
     },
     input_descriptions={
         'tree': TREE_DESC
     },
     parameter_descriptions={
-        'feature_metadata': FM_DESC
+        'feature_metadata': FM_DESC,
+        'shear_to_feature_metadata': (
+            'Determines whether or not to shear the tree to only the tips '
+            'that are present as features in the feature metadata. By '
```
fedarko, line 150:
```suggestion
            'Determines whether or not to shear the tree to only the tips '
            'that are present as features in the feature metadata. Internal '
            'nodes present in the feature metadata are ignored during '
            'shearing, and can potentially be filtered out if none of their '
            'child tips are present in the feature metadata. By '
```

feel free to reword this as you see fit. just wanted to make this abundantly clear here so people don't get confused.

fedarko (changes requested):
Thanks @gibsramen! I like the test feature metadata values.

I have some minor wording suggestions, and one small test case I think we should look into, but besides that this should be good.

(...also it's really fun deleting chunks of the feature metadata and playing around with tiny subsets of the moving pictures tree :)

image

tests/python/test_core.py
gibsramen: Makes sense. Just to clarify - should the test be that `viz.int_md` is empty or that `viz.features` does not contain `h` (or both)? Currently `viz.features` is not modified at all and only `viz.tip_md` and `viz.int_md` are (if necessary). If we want to remove filtered-out nodes from the "full" feature metadata I can update things accordingly (for the other tests as well).

fedarko: This is a really good question. Hm. From the looks of it, `viz.features` already isn't being updated to be filtered to stuff that is in the tree, so I don't think you'd need to worry about filtering it here. Just checking that `viz.int_md` is empty (or at least that it only contains metadata for internal nodes still in the tree post-shearing) should be fine. I think it makes sense that this big thing of "feature metadata" gets replaced by "tip metadata" and "internal node metadata" -- it makes it clear that some stuff might get filtered out along the way.

It is a bit confusing that this is how Empress currently does things, because looking into this further we do actually update `self.samples` after `match_inputs()`... I blame Past Marcus :P If you don't mind, it'd make sense to add in a short comment right around:

https://github.com/biocore/empress/blob/3da2572b49e06214d5e66d4ae7fd209b681bc513/empress/core.py#L133-L137

... that just quickly says "this will be transformed into self.tip_md and self.int_md in self._validate_and_match_data()" or something. But that's probably out of the scope of this PR so no worries.

fedarko:
Thanks @gibsramen! :scissors:  :sheep:

PR #432, opened 2020-10-23
Fix clicking on nodes not in table; refactor and test node selection menu; fix "Add" button

### LINKED ISSUE #272: Summarizing every sample metadata column leaves the "Add" button and UI elements visible
When this happens, the "Select a metadata column to summarize" selector's value becomes empty, but the "Add" button remains clickable. Clicking on the button triggers the following error:

```
Uncaught TypeError: Cannot read property 'remove' of undefined
    at HTMLButtonElement.click (select-node-menu.js:44)
```

... which seems pretty straightforward -- [the code](https://github.com/biocore/empress/blob/4987a806ca3ce2d0a5538905f019434f1026ed0b/empress/support_files/js/select-node-menu.js#L44) is trying to remove the selected option from the selector, but there aren't any options in the selector anymore so things break.

This isn't all that problematic of an issue, but it would be nice to at least hide these UI elements when all columns have been selected. (I imagine adding UI elements to remove stuff from the table would take more work, and should probably be deferred for later.)

Here's a GIF of what this looks like:

bububug

### LINKED ISSUE #314: Clicking on nodes that aren't present in the feature table causes errors
This occurs when `--p-no-filter-unobserved-features-from-phylogeny` is passed (i.e. the tree is _not_ shorn to the features present in the table).

Clicking on such a tip causes the following error:

```
Uncaught Error: Feature ID "tipIDgoeshere" not in BIOM table.
    at BIOMTable._getFeatureIndexFromID (/_/03wu22ds7qnm/b3f410c8-e22d-41d6-b8d7-0f56fc401a3e/data/support_files/js/biom-table.js:120)
    at BIOMTable.getObsCountsBy (/_/03wu22ds7qnm/b3f410c8-e22d-41d6-b8d7-0f56fc401a3e/data/support_files/js/biom-table.js:307)
    at Empress.computeTipSamplePresence (/_/03wu22ds7qnm/b3f410c8-e22d-41d6-b8d7-0f56fc401a3e/data/support_files/js/empress.js:2574)
    at SelectedNodeMenu.showLeafNode (/_/03wu22ds7qnm/b3f410c8-e22d-41d6-b8d7-0f56fc401a3e/data/support_files/js/select-node-menu.js:246)
    at SelectedNodeMenu.showNodeMenu (/_/03wu22ds7qnm/b3f410c8-e22d-41d6-b8d7-0f56fc401a3e/data/support_files/js/select-node-menu.js:194)
    at HTMLButtonElement.click (/_/03wu22ds7qnm/b3f410c8-e22d-41d6-b8d7-0f56fc401a3e/data/support_files/js/select-node-menu.js:46)
```

The selected node menu code assumes in multiple places that the tip is in the table (`this.empress.computeTipSamplePresence()` and `SelectedNodeMenu.makeSampleMetadataTable()`, among probably other places), which is a problem. A lot of this is likely my bad.

From what I can tell, `computeIntSamplePresence()` when clicking on an internal node where no descendant tips are in the table doesn't throw any errors, but a Sample Presence Info table with all 0s is still shown in the resulting node selection menu; this should be hidden.

For reference, I think the expected behavior when clicking on a node not in the table is that the user should still be able to see the node's name and any feature metadata available for it, with the "Sample Presence Information" stuff hidden.

fedarko:
Closes #314. Now, clicking on a tip not in the table shows text to the user explaining the situation:

image

Similarly, clicking on an internal node where none of its descendant tips are in the table also shows a note:

image

This also refactors the node selection menu a decent amount. Mainly, I realized while working on #314 that a lot of the showing / hiding operations in the menu rely on some pretty complex logic and were done in a very ad-hoc way (this is primarily my fault lol), so I tried to make things more manageable. Now, more of the work is delegated to `makeFeatureMetadataTable()` / `makeSampleMetadataTable()`, so life should be easier. This PR also adds some tests for the menu -- these are by no means comprehensive, but they should be a good start.

While I was at it, I also fixed #272 and added a test for it.

Last change worth noting, I think, is that I removed `SelectedNodeMenu._checkTips()`. When more than a few tips were not present in the tree, the toast messages shown were kind of hard to read (see below); furthermore, it looks like the tip keys rather than names were being shown, anyway, which makes interpreting these messages hard for users. Since the revamped menu now has the ability to show these sorts of warnings a lot more cleanly, I delegated this to there.

_Before:_
b42

_After:_
yeeee


fedarko:
... also, for reference for whoever reviews this, here's a flowchart i drew out on all of the cases possible when showing the sample metadata stuff to the user. Turns out user interfaces are really hard!

jank

(note: i might have missed something so this might not actually be comprehensive ._.)

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F432%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F432%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F432%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F432%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F432%2Fplain.qzv)

empress/support_files/js/select-node-menu.js
```diff
@@ -45,73 +97,169 @@ define(["underscore", "util"], function (_, util) {
...
+            if (_.isNull(ctData)) {
+                // This node (or its descendant tips) isn't present in the
+                // table. Just show some text explaining the situation, and
+                // hide the table.
+                var wtext;
+                if (tipOrInt === "tip") {
+                    wtext =
+                        "This is a tip in the tree. However, it is not " +
+                        "present in the input feature table, so we " +
+                        "can't show sample presence information for it.";
```
ElDeveloper, line 163: What about a shorter version:

```suggestion
                        "This tree tip is not " +
                        "present in the input feature table, so we " +
                        "can't show sample presence information for it.";
```

empress/support_files/js/select-node-menu.js
```diff
@@ -45,73 +97,169 @@ define(["underscore", "util"], function (_, util) {
...
+                } else {
+                    wtext =
+                        "This is an internal node in the tree. None of " +
```
ElDeveloper, line 166: Same here ... `None of the descendants of this internal node in the tree ...`

empress/support_files/js/select-node-menu.js
```diff
@@ -127,51 +275,50 @@ define(["underscore", "util"], function (_, util) {
...
-                colCell.innerHTML = "<strong>" + colName + "</strong>";
-                var dataCell = featureRow.insertCell(-1);
-                dataCell.innerHTML = mdObj[nodeName][x];
+        if (this.hasFeatureMetadata) {
+            this.fmTable.innerHTML = "";
+            var mdObj;
+            // TODO: it'd be nice to add methods to Empress that return the
+            // feature metadata, so we can avoid having to access "private"
+            // attributes here. This is planned as part of the
+            // https://github.com/biocore/empress/issues/337 refactoring.
```
ElDeveloper, line 292: If this is already documented elsewhere I think it's safe to keep this comment out.

```suggestion
```

empress/support_files/js/select-node-menu.js
```diff
@@ -461,30 +564,23 @@ define(["underscore", "util"], function (_, util) {
...
-            );
-        }
-
-        /* the buffer that holds the information to highlight the tree nodes
-         * on the canvas.
-         * format [x,y,r,g,b,...] where x,y is the coorindates of the
-         * highlighted nodes
+        /* Highlight the nodes in nodeKeys on the canvas.
+         * The buffer that holds this information is formatted as
+         * [x,y,r,g,b,...] where x,y are the coords of the highlighted nodes.
```
ElDeveloper, line 583: Is this still valid considering the changes that went in last week to update the way colors are stored?

empress/support_files/css/empress.css
```diff
@@ -651,15 +651,23 @@ p.side-header button:hover,
...
+
+#menu-box-node-name-warning {
     font-weight: bold;
-    max-width: 500px;
 }
 
-#menu-box-notes {
-    max-width: 500px;
-    font-size: small;
+/* TODO: remove overlap between top/bottom borders? */
```
ElDeveloper, line 663: Yeah that would probably be nice, the borders look a bit funny IMHO. I would suggest removing them altogether `¯\_(ツ)_/¯`

ElDeveloper (changes requested):
Works great, just a few minor suggestions.

empress/support_files/js/select-node-menu.js
fedarko: Oof, no it is not! Thanks for pointing this out. I'll update the comment.

empress/support_files/css/empress.css
fedarko: I kinda like the borders, just because it makes it clear which UI elements apply to the feature metadata stuff vs. which apply to the sample metadata stuff (the distinction isn't always clear IMO without the borders, e.g. the sample metadata summarize elements). I'll try to fix the adjacent border stuff and hopefully that makes it look nicer ;P

fedarko: Actually, I spent longer than I'd like to admit trying to make the borders look the way I wanted them to and was not very successful. So... yeah, I went ahead and removed the borders :man_shrugging: 

I did change up a few things to still make the "groupings" of UI elements clear. Here's what things look like now:

Both feature and sample metadata available:
image

Just feature metadata available:
image

Just sample metadata available:
image

I think this ended up looking better than it did before, so thanks for the suggestion ;)

empress/support_files/js/select-node-menu.js
fedarko: The reason I kept things like this was to be consistent with the other "notes" text that's shown in the alternative case (when this node is in the table) in Empress, [here](https://github.com/biocore/empress/blob/e982e65b5983299cddbf3665472bfd7d5cca254f/empress/support_files/js/select-node-menu.js#L283-L287) and [here](https://github.com/biocore/empress/blob/e982e65b5983299cddbf3665472bfd7d5cca254f/empress/support_files/js/select-node-menu.js#L389-L394).

I'd be down to change this in the future (maybe adding another UI element next to `Node length:` at the top of the menu that has something like `Node type:` and then `tip` or `internal`), but I would prefer to keep this as is for now. (I don't feel _super_ strongly about this, tho, so if you'd prefer we can change this up now :)

fedarko:
Thanks @ElDeveloper! Changes should be ready for a second pass.

empress/support_files/css/empress.css
ElDeveloper: That looks super slick! ✨ 

Thanks so much for looking into this!

ElDeveloper (approved):

fedarko:
Thanks @ElDeveloper!

ElDeveloper:
Thanks @fedarko!

PR #433, opened 2020-10-26
Fix 364

### LINKED ISSUE #364: Ensure node searching is handled well for large trees
Came up in this morning's (09/01/2020) meeting.

When there are a lot of tips in a tree (i.e. EMP with 300k), the search box auto-suggestion can be very slow. For example if the node names are the sequences (TACGTCA...), searching for "TAC" takes a long time as there are potentially many nodes to be suggested.

@kwcantrell had the idea of maybe showing only the top N hits. Alternatively we could make the auto-suggestion functionality optional for users who are visualizing large trees.

kwcantrell:
Resolves #364. This implements a binary search and also limits the number of suggestions to 10 ids

Here is a link to the improved [version](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdl.dropbox.com%2Fs%2F7xrpr6z8vzoot24%2Ffast_autofill.qzv%3Fdl%3D1) and here is a link to the original [version](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fdl.dropbox.com%2Fs%2F9ybj0aqtj0nhajz%2Fslow_autofill.qzv%3Fdl%3D1).

You can test them by searching for 'taag' in the node search box.

### DIFF
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -302,7 +302,7 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
             var suggestId,
                 suggestionsAdded = 0,
                 i = _.findIndex(ids, compareQuery, true);
-            
+
             // no match was found
             if (i === -1) {
                 return;
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -292,8 +292,11 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
                 suggestionsAdded = 0,
                 i = _.findIndex(
                     ids,
-                    function (id) {
-                        return id.substr(0, query.length) == query;
+                    function (w) {
+                        return (
+                            w.substr(0, query.length).toUpperCase() ==
+                            query.toUpperCase()
+                        );
                     },
                     true
                 );
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -300,7 +300,8 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
                     },
                     true
                 );
-            for (i; i < ids.length && suggestionsAdded < 10; i++) {
+            // for (i; i < ids.length && suggestionsAdded < 10; i++) {
+            for (i = 0; i < ids.length; i++) {
                 var word = ids[i];
 
                 // if node id begins with user query, add it to suggestionMenu
@@ -323,12 +324,12 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
                     // add suggested id to the suggstions menu
                     suggestionMenu.appendChild(suggestId);
                     suggestionsAdded += 1;
-                } else {
-                    // we set i to begin at an index where the users query will
-                    // match and since ids is an assorted array, once we do not
-                    // find a match we know we will not find any more matches.
-                    break;
-                }
+                } // else {
+                //     // we set i to begin at an index where the users query will
+                //     // match and since ids is an assorted array, once we do not
+                //     // find a match we know we will not find any more matches.
+                //     break;
+                // }
             }
 
             // not all node ids were listed in the autofill box
@@ -336,7 +337,6 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
             // possible options
             if (suggestionsAdded < ids.length) {
                 suggestId = document.createElement("DIV");
-                suggestId.id = word;
 
                 suggestId.innerHTML = "<strong>...</strong>";
                 suggestionMenu.appendChild(suggestId);
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -328,7 +328,7 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
             // not all node ids were listed in the autofill box
             // create an ellipse autofill (...) to let users know there are
             // more possible options
-            if (suggestionsAdded < ids.length) {
+            if (compareQuery(ids[i])) {
                 suggestId = document.createElement("DIV");
 
                 suggestId.innerHTML = "<strong>...</strong>";
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -287,6 +287,10 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
             autocompleteContainer.appendChild(suggestionMenu);
 
             // helper function to compare query to node name
+            // returns true if the first (query.length) characters of nodeName
+            // are equal to query, ignoring case (if query.length >
+            // nodeName.length this should be false, since .substr()
+            // will just not modify the nodeName)
             var compareQuery = function (nodeName) {
                 return (
                     nodeName.substr(0, query.length).toUpperCase() ===
@@ -322,8 +326,8 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
             }
 
             // not all node ids were listed in the autofill box
-            // create an elipse autofill to let users know there are more
-            // possible options
+            // create an ellipse autofill (...) to let users know there are
+            // more possible options
             if (suggestionsAdded < ids.length) {
                 suggestId = document.createElement("DIV");
 
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -287,24 +287,22 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
             autocompleteContainer.appendChild(suggestionMenu);
 
             // helper function to compare query to node name
-            var compareQuery = function(nodeName) {
-                return nodeName.substr(0, query.length).toUpperCase() ===
-                       query.toUpperCase();
-            }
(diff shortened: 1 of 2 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F433%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F433%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F433%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F433%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F433%2Fplain.qzv)

empress/support_files/js/canvas-events.js
```diff
@@ -284,7 +288,20 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
...
             // search ids array for all possible words
-            for (var i = 0; i < ids.length; i++) {
+            var suggestId,
+                suggestionsAdded = 0,
+                i = _.findIndex(
+                    ids,
+                    function (w) {
+                        return (
+                            w.substr(0, query.length).toUpperCase() ==
+                            query.toUpperCase()
```
fedarko, line 298: It looks like this check is duplicated just below this in `canvas-events.js`. Would you mind refactoring it into a helper function, just so we only need to update this once in the future if needed?

empress/support_files/js/canvas-events.js
```diff
@@ -284,7 +288,20 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
...
+                        );
```
fedarko, line 299: ... Also, this PR isn't really the place to ask about it, but just to check -- the case insensitivity only applies to suggestions, right? It should be possible for two nodes with differently-cased names (e.g. `"abc"` and `"AbC"`) to be present in the same tree, and each should be searchable individually. I think this is already the case, but just wanted to verify (if there are problems we can open a new issue for fixing that).

empress/support_files/js/canvas-events.js
```diff
@@ -306,7 +323,23 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
...
                     // add suggested id to the suggstions menu
                     suggestionMenu.appendChild(suggestId);
-                }
+                    suggestionsAdded += 1;
+                } // else {
+                //     // we set i to begin at an index where the users query will
+                //     // match and since ids is an assorted array, once we do not
+                //     // find a match we know we will not find any more matches.
+                //     break;
+                // }
```
fedarko, line 332: Should this be un-commented? (If not, we should probably remove it.)

fedarko (changes requested):
Thanks @kwcantrell! Some suggestions -- main thing is that it looks like the functionality added is still commented out, and should be re-enabled. Also, I think the null-filtering should probably be re-enabled, as well.

empress/support_files/js/canvas-events.js
kwcantrell: Correct, case insensitivity is only used in the suggestions menu to make comparisons easier. In you above example if the user typed `"a"` in the search menu then both `"abc"` and `"AbC" would be suggested to the user. However, case insensitivity is not used when searching for a specific node. For example if the user typed `"abc"` in the search menu and then either clicked on the search button or pressed the enter key, only node `"abc"` would be shown on the empress tree.

fedarko: Awesome, thanks for clarifying! Good to know.

empress/support_files/js/empress.js
```diff
@@ -431,7 +431,6 @@ define([
         var nodeNames = this._tree.getAllNames();
         // Don't include nodes with the name null (i.e. nodes without a
         // specified name in the Newick file) in the auto-complete.
-        nodeNames = nodeNames.filter((n) => n !== null);
         nodeNames.sort();
```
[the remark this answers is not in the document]
kwcantrell: The goal was to cut back the number of computational cycles. But I agree, filtering the nulls is a one time cost and its worth doing to avoid any potential errors.

fedarko:
Thanks so much @kwcantrell! Going through the updates now :)

@antgonza Quick question -- is it possible to send a "kick"/trigger message to McHelper? Going by the provenance / functionality it looks like the QZVs for this PR haven't been updated in response to the recent commits to this branch, which is making manual testing a bit more cumbersome here.

funky

empress/support_files/js/canvas-events.js
```diff
@@ -305,8 +317,19 @@ define(["glMatrix", "SelectedNodeMenu"], function (gl, SelectedNodeMenu) {
...
                     // add suggested id to the suggstions menu
                     suggestionMenu.appendChild(suggestId);
+                    suggestionsAdded += 1;
                 }
             }
+
+            // not all node ids were listed in the autofill box
+            // create an elipse autofill to let users know there are more
+            // possible options
+            if (suggestionsAdded < ids.length) {
```
fedarko, line 327: It looks like the `...` shows up even if there is only one value shown in the menu -- I think this is because there is only one possible "option" here, i.e. the `0.000` name, so `suggestionsAdded` for the query `0.000` is immediately less than `ids.length`. I thought at first that this might be a result of the unique-filtering code you added, but from going through the code I don't think that's the reason.

I think (?) there should instead be some logic that distinguishes between when we already have 10 suggestions added and we had to cut things off (in which case we should show the `...`), and the other cases (e.g. the query is so specific that 10 or fewer items are shown, in which case we should not show the `...`)

uniqs_huh

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: Thanks -- yeah, if this gets to be a bottleneck we can definitely speed things up somehow, e.g. maybe adding a special `BPTree.getAllNonNullNodeNames()` or something like that. But for now I think this should be fine.

Filtering to unique node names makes sense, I like that this cleans things up in the menu :D

fedarko (changes requested):
A few suggestions (mostly documentation but one functionality thing). Thanks so much, this is looking awesome!

antgonza:
@fedarko, my guess is that `d9f6850` and `056f93b` happened too quickly; @kwcantrell, could you do one push with some minimal changes to see if things work fine?

empress/support_files/js/empress.js
```diff
@@ -443,6 +443,7 @@ define([
         // specified name in the Newick file) in the auto-complete.
         nodeNames = nodeNames.filter((n) => n !== null);
         nodeNames.sort();
```
fedarko, line 445: https://stackoverflow.com/a/9645447/10730311

empress/support_files/js/empress.js
```diff
@@ -442,7 +442,11 @@ define([
         // Don't include nodes with the name null (i.e. nodes without a
         // specified name in the Newick file) in the auto-complete.
         nodeNames = nodeNames.filter((n) => n !== null);
-        nodeNames.sort();
+        nodeNames.sort(function (a, b) {
+            return a.localeCompare(b, "en", { sensitivity: "base" });
+        });
+        console.log(nodeNames);
```
fedarko, line 448: Could you remove the console.log here?

Also -- do you think it might make sense to move some of this code to a new utility function that just takes a list of node names and filters nulls, sorts it case-insensitively, and filters it to unique stuff? This way, it'd be a lot easier to test this functionality directly. (I mean it works now, so it's not a big issue or anything, but it would probs make the code a bit cleaner.) Not something we have to do in this PR, necessarily.

empress/support_files/js/empress.js
```diff
@@ -442,7 +442,11 @@ define([
         // Don't include nodes with the name null (i.e. nodes without a
         // specified name in the Newick file) in the auto-complete.
         nodeNames = nodeNames.filter((n) => n !== null);
-        nodeNames.sort();
+        nodeNames.sort(function (a, b) {
+            return a.localeCompare(b, "en", { sensitivity: "base" });
+        });
```
fedarko, line 447: Cool! No idea this existed, this seems like a good way of handling this.

empress/support_files/js/empress.js
```diff
@@ -442,7 +442,11 @@ define([
         // Don't include nodes with the name null (i.e. nodes without a
         // specified name in the Newick file) in the auto-complete.
         nodeNames = nodeNames.filter((n) => n !== null);
-        nodeNames.sort();
+        nodeNames.sort(function (a, b) {
```
fedarko, line 445:
```suggestion
        // Sort node names case insensitively
        nodeNames.sort(function (a, b) {
```

fedarko (commented):
Some tiny comments, but this is pretty much fine as is. Thanks so much @kwcantrell!

Issue #439, opened 2020-10-29
Save/reload EMPress plot


kwcantrell:
It would be nice to have a mechanism to save an EMPress plot.

kwcantrell:
same as #131

PR #440, opened 2020-10-31
ENH Feature metadata coloring command pattern

gwarmstrong:
@kwcantrell and I put together a proposed refactoring of feature metadata coloring using the Command Pattern. I am tagging this as progress on #439 , but there are probably also others that could apply. We were hoping to get some feedback on the general design, and if the reception is positive, would be happy to flesh out more of the docs/tests.

But the general mechanism we introduce here is the [Command pattern](https://en.wikipedia.org/wiki/Command_pattern).  

Namely, the proposal would be that updates to Empress plots should be routed through `Command`s, where you encapsulate all of the information needed to perform an update in the create of the `Command` object. This can seen, for example in the lines:
```js
            let command = new Commands.ColorFeatureTreeCommand({
                empress: scope.empress,
                collapseClades: collapseChecked,
                lineWidth: lw,
                colBy: colBy,
                col: col,
                coloringMethod: coloringMethod,
                reverse: reverse,
            });
```
where all of the information needed to color the tree based on it's feature metadata is passed to the constructor `ColorFeatureTreeCommand`.

This command can then later be executed by calling the `execute` method, i.e, `command.execute();`, which actually performs the tree coloring based on the code in `ColorFeatureTreeCommand`'s `execute` method.

We have also included a `CommandManager` class which maintains a queue of commands to execute, and a history of commands that have been executed via the `CommandManager`. It also comes with utility functions for helping serialize and de-serialize the history. If all of the UI elements are routed through the command manager, this could be used to completely reconstruct the state of the empress plot, enabling save/load/playback feature (that admittedly could get memory intensive for long running sessions, given the current implementation).

Using the command pattern throughout this project could also be useful for a variety of other 'wishlist' items, such as programming custom animations, i.e.,
```js

command1 = new Commands.ChangeTreeLayoutCommand(...);
command2 = new Commands.ColorTreeByFeatureCommand(...);
command3 = new Commands.AddBarplootCommand(...);

commands = [command1, command2, command3];
commands.forEach(function(command) {
    setTimeout(function() {
        command.execute();
    }, 5000);
});
```

And you could have some event bus subscriber that creates commands. E.g.,

```js
function Subscriber(bus, ...) {
    ...
}

Subscriber.prototype.handle = function (event) {
    ...
    colorScheme = // something you are to glean from `event`
    let command = new Commands.ColorFeatureTreeCommand(..., colorScheme, ...);
    command.execute();
    ...
}
``` 

Which should hopefully allow for relatively extensible interaction code (e.g., interacting with Emperor)


dhakim87:
>             let command = new Commands.ColorFeatureTreeCommand({
>                 empress: scope.empress,
>                 collapseClades: collapseChecked,
>                 lineWidth: lw,
>                 colBy: colBy,
>                 col: col,
>                 coloringMethod: coloringMethod,
>                 reverse: reverse,
>             });

I'm not sure I totally follow the example.  Is there an interface that commands must follow - Some set of fields they all share?  

Why is it empress: scope.empress rather than scope: scope.empress?
Should colorFeatureTree be split into colorFeatureTree and lineWidthFeatureTree?  

It looks like colBy, col, and coloringMethod are all optional?  Color by (dictionary?), Color (fixed color?), Coloring Method (function?) - coloringMethod would have to be the name of a function I guess, in order to allow json serialization.  Or is this intended to be an enum specifying whether colorBy or color should be used?  

I think you might want a command_type: xxx added to all the commands to make parsing easy.

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F440%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F440%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F440%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F440%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F440%2Fplain.qzv)

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,95 @@
+define([], function () {
+    var defaultCommands = {
+        ColorFeatureTreeCommand: ColorFeatureTreeCommand,
+    };
+
+    function CommandManager(empress) {
+        this.empress = empress;
+        this.executed = [];
```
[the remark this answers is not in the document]
dhakim87: One use case this one makes sense for is Undo/Redo functionality, but I don't think that really works with a message bus architecture and I haven't heard it being requested.  (It also requires all commands to store enough state to be invertible)

dhakim87 (commented):
I really like this.  I think this is the right skeleton for a message passing solution.  I have minor comments on removing boilerplate and optimizations that may eventually become issues.

empress/support_files/js/side-panel-handler.js
```diff
@@ -564,12 +570,28 @@ define(["underscore", "Colorer", "util"], function (_, Colorer, util) {
...
+            // hide update button
+            scope.fUpdateBtn.classList.add("hidden");
+
+            // this came from colorMethodName
+            let colBy = scope.fSel.value;
+            let col = scope.fColor.value;
+            let coloringMethod = scope.fMethodChk.checked ? "tip" : "all";
+            let reverse = scope.fReverseColor.checked;
+            let collapseChecked = scope.fCollapseCladesChk.checked;
+            let lw = util.parseAndValidateNum(scope.fLineWidth);
```
fedarko, line 582: It's not a big deal, but for consistency's sake I think it would make sense to use `var` instead of `let` here. (It's a bit up-in-the-air right now, as documented in #222.)

Edit: ah, I just saw that we're focusing more on the high-level stuff for now... Yeah, for now this is fine :)

empress/support_files/js/side-panel-handler.js
```diff
@@ -564,12 +570,28 @@ define(["underscore", "Colorer", "util"], function (_, Colorer, util) {
...
+                empress: scope.empress,
+                collapseClades: collapseChecked,
+                lineWidth: lw,
+                colBy: colBy,
+                col: col,
+                coloringMethod: coloringMethod,
+                reverse: reverse,
+            });
+            scope.commandManager.push(command);
+            scope.commandManager.executeAll();
```
fedarko, line 594: This approach is fine, but it seems to me like we'll have to use the idiom of "add this command and then immediately try to run it" a lot throughout the codebase. Do you think it would make sense to add a convenience function for the CommandManager which takes care of both tasks, for these cases? I'm imagining something like `scope.commandManager.pushAndExecute(command);`. (Not sure if this is completely not kosher with the Command pattern, so feel free to ignore if so :)

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,95 @@
...
+        object.forEach(function (commandObject) {
+            var commandConstructor = scope.commands[commandObject.command];
+            var args = commandObject.arguments;
+            args.empress = scope.empress;
+            var newCommand = new commandConstructor(args);
+            scope.push(newCommand);
+        });
+    };
+
+    function ColorFeatureTreeCommand(arguments) {
```
fedarko, line 52: It seems to me like the individual command classes (which as @dhakim87 suggested would be derived from a Command base class) should have their own concrete requirements for arguments, although I guess the base class wouldn't have any such requirements.

Hm, I'm not sure what the best way to set this up would be. Knowing JS, there are probably like a billion ways to do this, and you've probably already thought of some of these :) To add to the noise, one idea that should work would be (assuming a Command base class will be eventually added) adding a method to that base class that goes through the stuff in `arguments` and checks it against, say, a Command-subclass-specific array of needed arguments, and raises an error if some needed stuff isn't specified (using some of the property stuff Dan suggested -- and/or we can probably use some of [underscore.js](https://underscorejs.org/#keys) to do this easily, e.g. `_.keys()` and `_.intersection()`).

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,95 @@
...
+            var commandObject = {
+                command: command.constructor.name,
+                arguments: command.toObject(),
+            };
+            objects.push(commandObject);
+        });
+        return objects;
+    };
+
+    CommandManager.prototype.parse = function (object) {
```
fedarko, line 41: One of those "not a huge deal but" things -- I think we should try to avoid `object` as a variable name, just because `Object` is a keyword in JS.

I'm also a bit confused by how this function works, since it seems to me on first reading that the `object` parameter is actually an Array of Command objects (since `forEach()` isn't a property of Objects in JS). If this is the case, I would suggest renaming `object` to something like `commands` (I know docs aren't a big focus of the initial work here but this would be good to make clear).

fedarko (commented):
This looks great, thanks so much! I have some comments -- they're mostly small-scale things, though. By and large I think the overall structure looks great. Having this in will clean things up a lot :)

empress/support_files/js/commands.js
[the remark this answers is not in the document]
gwarmstrong: Right so this `CommandManager` is intended just to queue up commands for empress only, and hopefully be able to replay some actions on empress. We would require a separate class to manage interactions with an event bus.

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,95 @@
...
+        this.toExecute = [];
+        this.commands = defaultCommands;
+    }
+
+    CommandManager.prototype.push = function (command) {
+        this.toExecute.push(command);
+    };
+
+    CommandManager.prototype.executeNext = function () {
+        let command = this.toExecute.shift();
```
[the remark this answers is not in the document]
gwarmstrong: Hmmmm, any pointers to places where people have written an actual linked list/dequeue in javascript? Most of the implementations I have found just use shift 😆

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,95 @@
...
+        command.execute();
```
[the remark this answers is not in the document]
gwarmstrong: ^ended up following this and switching to `execute(empress)` and it is a lot cleaner.

empress/support_files/js/side-panel-handler.js
gwarmstrong: Good idea an incorporated now.

empress/support_files/js/commands.js
gwarmstrong: Make sense and updated now!

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,285 @@
+define(["underscore"], function (_) {
+    function CommandManager(empress) {
```
fedarko, line 2: Not all that urgent, but it would be good to add a comment here similar to what the `Command` class has below (just to make it clear that this is also a "class", at least kinda). There are examples of this (for non-explicitly-declared JS classes in Empress) in `bp-tree.js` and `empress.js`.

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,285 @@
...
+        this.empress = empress;
+        this.executed = [];
+        this.toExecute = [];
+    }
+
+    CommandManager.prototype.push = function (command) {
```
fedarko, line 8: I guess now that we're at the stage of getting ready to merge this in, it'd be nice to add some basic docs for these functions (sorry, I realize this is going to be like 4 separate comments ._.)

At the very least, it'd be good to be very clear that a first-in first-out queue is used (... at least that is what I think is done from reading this).

empress/support_files/js/commands.js
```diff
@@ -0,0 +1,285 @@
...
+
+        execute(controller) {
+            this.commands.forEach(function (command) {
+                command.execute(controller);
+            });
+        }
+    }
+
+    /**
+     * Composite pipeline for performing multipel steps of coloring a tree by feature metadata
```
fedarko, line 237:
```suggestion
     * Composite pipeline for performing multiple steps of coloring a tree by feature metadata
```

fedarko (commented):
I haven't finished going through all of the new functionality yet (will do that later this week), but here's what I have by now.

empress/support_files/js/commands.js
gwarmstrong: Added in https://github.com/biocore/empress/pull/440/commits/00baae9f230c9f34603d819a246e119fcbf22725

empress/support_files/js/commands.js
gwarmstrong: Added in https://github.com/biocore/empress/pull/440/commits/00baae9f230c9f34603d819a246e119fcbf22725

Issue #279, opened 2020-07-28
Support drawing the tree as ultrametric


fedarko:
Suggestion from @esayyari during this morning's meeting.

This would be similar to #271, but -- rather than each branch's length being equal -- the distance from the root to each tip is equal across all tips.

This would probably be another good thing to add in after JS layouts are supported (#213).

**Edit:** [here's](http://phylobotanist.blogspot.com/2018/04/time-calibrated-or-at-least-ultrametric.html) a link to a really nice blog post describing a bunch of different ways to make trees ultrametric. This is probably going to require exposing a few different options to the user in the interface, rather than a single "Ultrametric or not ultrametric?" checkbox.

wasade:
Unit normalization may also make sense (e.g., div all branches by max tip->root distance)

gwarmstrong:
It looks like branch lengths are set in the layout functions in layouts-util.js in [rectangularLayout](https://github.com/biocore/empress/blob/13438f9646b43ed79651624b827b92ca25ae9e4e/empress/support_files/js/layouts-util.js#L133), [circularLayout](https://github.com/biocore/empress/blob/13438f9646b43ed79651624b827b92ca25ae9e4e/empress/support_files/js/layouts-util.js#L365), and [unrootedLayout](https://github.com/biocore/empress/blob/13438f9646b43ed79651624b827b92ca25ae9e4e/empress/support_files/js/layouts-util.js#L543).

All of those lines are the same and could probably be moved to a function call that determines node lengths, i.e., 
```js
function getNodeLengths(ignoreLength, tree, position, precomputedUltrametricLength=null) {
  var nodeLen;
  if (precomputedUltrametricLength === null) {
     nodeLen = ignoreLength ? 1 : tree.length(position);
  } else {
    nodeLen = precomputedUltrametricLength[position];
  }
  return nodeLen
}
```
Where `precomputedUltrametricLength` is computed before the call to this function, depending on the tree should be drawn as ultrametric. Passed as `null` (or not passed) if the tree should not be ultrametric.

gibsramen:
> Edit: here's a link to a really nice blog post describing a bunch of different ways to make trees ultrametric. This is probably going to require exposing a few different options to the user in the interface, rather than a single "Ultrametric or not ultrametric?" checkbox.

[...]

PR #444, opened 2020-11-10
Ultrametric option

gwarmstrong:
@gibsramen and I threw together a method for rendering trees as ultrametric (see #279 )

Here are some screenshots of the ultrametric trees:
Rectangular
Screen Shot 2020-11-09 at 4 26 57 PM
Circular
Screen Shot 2020-11-09 at 4 32 36 PM
Unrooted
Screen Shot 2020-11-09 at 6 54 35 PM

I am unsure why the unrooted layout does not appear to have equal root to tip distances is unclear to me and I could use some guidance here.

This method relies on the algorithm as described in comments and below:
```
         The lengths for intermediate nodes are effectively "stretched" until
         their deepest descendant hits the deepest level in the whole tree.

         E.g., if we are at the node represented by * in the tree below:

         |--------------------------maxDistance-------------------------|
         |--distanceAbove--|           |---distanceBelow---|
                            |-length--|                     |-remainder-|
                                                    ____
                                        ___________|
                            *__________|           |_______
          __________________|          |__
                            |
                            |___________________________________________

         then the branch will be extended so that its deepest tip has the
         same depth as the deepest tip in the whole tree,
         i.e., newLength = length + remainder
         however, below it is equivalently calculated with
         newLength = maxDistance - distanceAbove - distanceBelow

         E.g.,
         |--------------------------maxDistance-------------------------|
         |--distanceAbove--|                        |---distanceBelow---|
                            |-length--||-remainder-|
                                                                 ____
                                                     ___________|
                            *_______________________|           |_______
          __________________|                       |__
                            |
                            |___________________________________________

        Repeated in a pre-order traversal, this will result in an ultrametric tree

```

## Questions for reviewers
* Do you know why the unrooted layout renders as it does?
* I did not spend a ton of time thinking out how this should be exposed to the user. Currently it is a checkbox under the layout page, and if it is check the "Ignore branch lengths" checkbox is ignored. How do you think we should code the interaction between these two functionalities?

### DIFF
--- a/empress/support_files/css/empress.css
+++ b/empress/support_files/css/empress.css
@@ -683,6 +683,10 @@ p.side-header button:hover,
     margin-left: 2em;
 }
 
+.warning {
+    color: #ff0000;
+}
+
 .selected-metadata-choice {
     background-color: #333 !important;
 }
--- a/empress/support_files/css/empress.css
+++ b/empress/support_files/css/empress.css
@@ -683,10 +683,6 @@ p.side-header button:hover,
     margin-left: 2em;
 }
 
-.warning {
-    color: #ff0000;
-}
-
 .selected-metadata-choice {
     background-color: #333 !important;
 }
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -362,25 +362,13 @@ define([
      * Also updates this._maxDisplacement.
      */
     Empress.prototype.getLayoutInfo = function () {
-        var data, i, lengthGetter;
+        var data, i;
         // set up length getter
-        var ultraMetricLengths;
         var branchMethod = this.branchMethod;
-
-        if (branchMethod === "ultrametric") {
-            ultraMetricLengths = LayoutsUtil.getUltrametricLengths(this._tree);
-            lengthGetter = function (i) {
-                return ultraMetricLengths[i];
-            };
-        } else if (branchMethod === "ignore") {
-            lengthGetter = function (i) {
-                return 1;
-            };
-        } else if (branchMethod === "normal") {
-            lengthGetter = null;
-        } else {
-            throw "Invalid branchMethod: '" + branchMethod + "'.";
-        }
+        var lengthGetter = LayoutsUtil.getLengthMethod(
+            branchMethod,
+            this._tree
+        );
 
         // Rectangular
         if (this._currentLayout === "Rectangular") {
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -276,6 +276,12 @@ define([
         this.ignoreLengths = false;
 
         /**
+         * @type{Bool}
+         * Whether to make branch lengths ultrametric during layout or not
+         */
+        this.makeUltrametric = false;
+
+        /**
          * @type{String}
          * Leaf sorting method: one of "none", "ascending", or "descending"
          */
@@ -356,7 +362,17 @@ define([
      * Also updates this._maxDisplacement.
      */
     Empress.prototype.getLayoutInfo = function () {
-        var data, i;
+        var data, i, lengthGetter;
+        // set up length getter
+        var ultraMetricLengths;
+        if (this.makeUltrametric) {
+            ultraMetricLengths = LayoutsUtil.getUltrametricLengths(this._tree);
+            lengthGetter = function (i) {
+                return ultraMetricLengths[i];
+            };
+        } else {
+            lengthGetter = null;
+        }
         // Rectangular
         if (this._currentLayout === "Rectangular") {
             data = LayoutsUtil.rectangularLayout(
@@ -364,7 +380,9 @@ define([
                 4020,
                 4020,
                 this.ignoreLengths,
-                this.leafSorting
+                this.leafSorting,
+                undefined,
+                lengthGetter
             );
             this._yrscf = data.yScalingFactor;
             for (i = 1; i <= this._tree.size; i++) {
@@ -385,7 +403,10 @@ define([
                 4020,
                 4020,
                 this.ignoreLengths,
-                this.leafSorting
+                this.leafSorting,
+                undefined,
+                undefined,
+                lengthGetter
             );
             for (i = 1; i <= this._tree.size; i++) {
                 // remove old layout information
@@ -409,7 +430,9 @@ define([
                 this._tree,
                 4020,
                 4020,
-                this.ignoreLengths
+                this.ignoreLengths,
+                undefined,
+                lengthGetter
             );
             for (i = 1; i <= this._tree.size; i++) {
                 // remove old layout information
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -405,7 +405,6 @@ define([
                 this.ignoreLengths,
                 this.leafSorting,
                 undefined,
-                undefined,
                 lengthGetter
             );
             for (i = 1; i <= this._tree.size; i++) {
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -325,9 +325,9 @@ define([
          * @type{Array}
          * @private
          *
-         * Stores the vertex information that is passed to WebGl
+         * Stores the vertex information that is passed to WebGL
          *
-         * Format: [x, y, r, g, b, ...]
+         * Format: [x, y, RGB, ...]
          */
         this._collapsedCladeBuffer = [];
 
@@ -488,7 +488,40 @@ define([
     };
 
     /**
-     * Retrives the coordinate info of the tree.
+     * Exports a SVG image of the active legends.
+     *
+     * Currently this just includes the legend used for tree coloring, but
+     * eventually this'll be expanded to include all the barplot legends as
+     * well.
+     *
+     * @return {String} svg
+     */
+    Empress.prototype.exportLegendSVG = function () {
+        var legends = [];
+        // Add the legend used for coloring the tree (if shown).
+        if (this._legend.isActive()) {
+            legends.push(this._legend);
+        }
+        // Add all the active legends from all the barplot layers.
+        // NOTE: Since we expect there to be many more barplot legends than
+        // just the one tree-coloring legend, we could potentially save a tiny
+        // bit of time by just setting legends to the output of getLegends()
+        // and then calling unshift() to add this._legend to the start of the
+        // array. However, I don't think the speed gain would be worth making
+        // this code much less readable ._.
+        if (this._barplotsDrawn) {
+            legends.push(...this._barplotPanel.getLegends());
+        }
+        if (legends.length === 0) {
+            util.toastMsg("No active legends to export.", 5000);
+            return null;
+        } else {
+            return ExportUtil.exportLegendSVG(legends);
+        }
+    };
+
+    /**
+     * Retrieves the coordinate info of the tree.
      *
      * We used to interlace the coordinate information with the color information
      * i.e. [x1, y1, red1, green1, blue1, x2, y2, red2, green2, blue2,...]
@@ -741,31 +774,6 @@ define([
     };
 
     /**
-     * Exports a SVG image of the active legends.
-     *
-     * Currently this just includes the legend used for tree coloring, but
-     * eventually this'll be expanded to include all the barplot legends as
-     * well.
-     *
-     * @return {String} svg
-     */
-    Empress.prototype.exportLegendSVG = function () {
-        var legends = [];
-        if (!_.isNull(this._legend.legendType)) {
-            legends.push(this._legend);
-        }
-        // TODO: get legends from barplot panel, which should in turn get them
-        // from each of its barplot layers. For now, we just export the tree
-        // legend, since we don't support exporting barplots quite yet (soon!)
-        if (legends.length === 0) {
-            util.toastMsg("No active legends to export.", 5000);
-            return null;
-        } else {
-            return ExportUtil.exportLegendSVG(legends);
-        }
-    };
-
-    /**
      * Exports a PNG image of the canvas.
      *
      * This works a bit differently from the SVG exporting functions -- instead
@@ -805,9 +813,8 @@ define([
     /**
      * Retrieves the node coordinate info (for drawing node circles).
      *
-     * @return {Array} Node coordinate info, formatted like
-     *                 [x, y, red, green, blue, ...] for every node circle to
-     *                 be drawn.
+     * @return {Array} Node coordinate info, formatted like [x, y, RGB float]
+     *                 for every node circle to be drawn.
      */
     Empress.prototype.getNodeCoords = function () {
         var tree = this._tree;
@@ -1353,10 +1360,34 @@ define([
     };
 
     /**
-     * Draws barplots on the tree.
+     * Computes the coordinate data needed for drawing a collection of barplot
+     * layer(s), as well as additional information needed for populating the
+     * corresponding barplot legends.
      *
-     * @param {Array} layers An array of BarplotLayer objects. This can just be
-     *                       the "layers" attribute of a BarplotPanel object.
+     * Similar to this.getCoords().
+     *
+     * @param {Array} layers Collection of BarplotLayer objects. Layers will be
+     *                       drawn starting from the edge of the tree and going
+     *                       outwards: the first layer in the array will be the
+     *                       innermost and the last will be the outermost
+     *                       (ignoring barplot border layers, which may be
+     *                       added depending on the BarplotPanel's state).
+     *
+     * @returns {Object} Contains three entries:
+     *                   -coords: An array of coordinate data, in the format
+     *                    [x, y, RGB...]
+     *                   -colorers: An Array of the same length as the number
+     *                    of barplot layers containing in each position either
+     *                    a Colorer object (for layers for which a color legend
+     *                    should be shown) or null (for layers for which no
+     *                    color legend should be shown).
+     *                   -lengthExtrema: An Array of the same length as the
+     *                    number of barplot layers containing in each position
+     *                    either another Array of two elements (the minimum and
+     *                    maximum value to be shown in a length legend) or
+     *                    null (for layers for which no length legend should be
+     *                    shown).
+     *
      * @throws {Error} If any of the following conditions are met:
      *                 -One of the layers is of barplot type "fm" and:
      *                    -A field with < 2 unique numeric values is used to
@@ -1367,9 +1398,11 @@ define([
      *                     scaleLengthByFMMax attribute is smaller than its
      *                     scaleLengthByFMMin attribute
      */
-    Empress.prototype.drawBarplots = function (layers) {
+    Empress.prototype.getBarplotData = function (layers) {
         var scope = this;
-        var coords = [];
+
+        // The main thing that will be returned by this function
+        var barplotBuffer = [];
 
         // Add on a gap between the closest-to-the-root point at which we can
         // start drawing barplots, and the first barplot layer. This could be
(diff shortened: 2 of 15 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F444%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F444%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F444%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F444%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F444%2Fplain.qzv)

fedarko:
This looks awesome! Will go through things in depth tomorrow.

> I did not spend a ton of time thinking out how this should be exposed to the user. Currently it is a checkbox under the layout page, and if it is check the "Ignore branch lengths" checkbox is ignored. How do you think we should code the interaction between these two functionalities?

If these options are mutually exclusive, I think using radio buttons for this might make sense for now. Alternately, we could set up something where clicking the "Make tree ultrametric?" checkbox (analogous to the "Draw barplots?" checkbox) hides the "ignore lengths" stuff and opens up a `<div>` with options for which ultrametric method to use. (But since there's just one ultrametric method supported -- for now at least -- I think just using radio buttons would be the easiest.)

kwcantrell:
Thanks @gwarmstrong and @gibsramen, this is really cool.

> Do you know why the unrooted layout renders as it does?

I'll be able to explain this better during our meeting today. Basically, the above algorithm works with circular/rectangular because branch lengths will always travel from a common point/direction. For example, in rectangular layout, branch lengths will always stretch the tree in the +x direction. Similarly in circular layout, all branch lengths will project out from a common point. This "standardizes"  branch lengths. For example, say we have two sets of branches a->b and c->d, 'a' has a length of 1, 'b' has a length of 5, and both 'c' and 'd' have a length of 3. This means 'b' and 'd' will have a "total aggregate" length of 6. Thus, in rectangular layout,  the x coordinate of 'b' and 'd' will be 6 and in circular layout, the l2 distance of 'b' and 'd' from the root will be equivalent. Thus, by setting the "total aggregate" length of all tips to be the same will make the tree appear ultrametric. However, in unrooted layout, you can essentially think the direction a node travels is "random". Basically, just because two nodes have the same "total aggregate" length, not necessarily mean they will be the same distance from the root (i.e. the l2 distance of 'b' and 'd' from the root will not be equivalent). 

**tldr** in unrooted making the "total aggregate" length of all tips equivalent does not guarantee they will be the same distance from the root.

gwarmstrong:
> **tldr** in unrooted making the "total aggregate" length of all tips equivalent does not guarantee they will be the same distance from the root.

Ah okay, I think my confusion can be explained away as a misunderstanding of where the root was. So to clarify, the root of the tree in the unrooted layout is probably somewhere around here?
Screen Shot 2020-11-09 at 6 54 35 PM

kwcantrell:
> Ah okay, I think my confusion can be explained away as a misunderstanding of where the root was. So to clarify, the root of the tree in the unrooted layout is probably somewhere around here?

Correct, so the "aggregate" branch lengths of the tips are the same however the l2 distances are different.

empress/support_files/js/empress.js
```diff
@@ -356,15 +362,42 @@ define([
...
+            };
+        } else if (branchMethod === "ignore") {
+            lengthGetter = function (i) {
+                return 1;
+            };
+        } else if (branchMethod === "normal") {
+            lengthGetter = null;
+        } else {
+            throw "Invalid branchMethod: '" + branchMethod + "'.";
+        }
```
fedarko, line 383: This is a bit confusing to me -- it looks like this is somewhat redundant with the stuff in `LayoutsUtil._determineLengthGetter()`. Would it be possible to move more of this logic into `LayoutsUtil`? I think just passing the `branchMethod` to the layout functions (replacing the `ignoreLengths` option, as you mentioned below), and having them pass that to `_determineLengthGetter()`, might make sense.

(...if that would be too much work for right now feel free to ignore this, we could open another issue to take care of it later :)

empress/support_files/js/empress.js
```diff
@@ -356,15 +362,42 @@ define([
...
                 4020,
                 4020,
-                this.ignoreLengths,
-                this.leafSorting
+                // since lengths for "ignoreLengths" are set by `lengthGetter`,
+                // we don't need (and should likely deprecate) the ignoreLengths
+                // option for the Layout functions since the layout function only
+                // needs to know lengths in order to layout a tree, it doesn't
+                // really need encapsulate all of the logic for determining
+                // what lengths it should lay out.
```
fedarko, line 396: Yeah, that's fair. My take on this is that I think figuring out lengths junk should be the job of the stuff in LayoutsUtils, just so we can avoid having logic in the main `Empress` class as much as we can (since it's already probs too big for its own good, and also to make testing easier).

empress/support_files/js/layouts-util.js
```diff
@@ -32,6 +32,112 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
+                        maxNodeToTipDistance[child] + tree.length(child);
+                    if (childMaxLen > maxDist) {
+                        maxDist = childMaxLen;
+                    }
+                }
+                maxNodeToTipDistance[nodeIndex] = maxDist;
+            }
+        }
+        /*
+         This loop is responsible for determining new branch lengths
```
fedarko, line 75:
```suggestion
         This loop is responsible for determining new branch lengths.
```

empress/support_files/js/layouts-util.js
```diff
@@ -32,6 +32,112 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
+        for (i = 1; i <= tree.size; i++) {
+            nodeIndex = tree.preorderselect(i);
+            children = tree.getChildren(nodeIndex);
+            for (j = 0; j < children.length; j++) {
+                child = children[j];
+                var totalDistance = maxDistance;
+                var distanceAbove = depths[nodeIndex];
+                var distanceBelow = maxNodeToTipDistance[child];
+                lengths[child] = totalDistance - distanceAbove - distanceBelow;
+                depths[child] = distanceAbove + lengths[child];
```
fedarko, line 123:
```suggestion
                var distanceAbove = depths[nodeIndex];
                var distanceBelow = maxNodeToTipDistance[child];
                lengths[child] = maxDistance - distanceAbove - distanceBelow;
                depths[child] = distanceAbove + lengths[child];
```

empress/support_files/js/layouts-util.js
```diff
@@ -32,6 +32,112 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
+            }
+        }
+        return lengths;
+    }
+
+    function _determineLengthGetter(lengthGetter, ignoreLengths, tree) {
```
fedarko, line 129: Could you please add a comment explaining the types of the arguments? (I know it's an internal function so it's not a huge deal, but it'd help make the code easier to read)

fedarko (changes requested):
Thanks @gwarmstrong and @gibsramen! This looks solid (and the algorithm seems to me like a really elegant way of doing this). I have some suggestions, but most of them are cosmetic things -- this is going to be super useful.

empress/support_files/js/empress.js
gwarmstrong: The difference between this chunk and `_determineLengthGetter` is that `_determineLengthGetter` maintains the default behavior of the Layout utilities, whereas this function maps keywords to specific methods. This would definitely tidier if it was its own function though.

empress/support_files/js/empress.js
gwarmstrong: I think that passing a function for the lengths is nice because it opens the door for us to do some pretty cool stuff later, without having to make modifications to the `LayoutsUtil` module and make those functions signatures more unmanageable.

E.g., say we wanted a feature where you could adjust branch lengths on the fly, you could define some function like 
```
var userSetLengths = // some Object containing lengths that the user set explicitly
lengthGetter = function(i) {
    if (i in userSetLengths) {
        return userSetLengths[i];
    } else {
        return this._tree.length(i);
}
```

And it shouldn't really be `LayoutsUtil`'s job to support this.

gwarmstrong: What if we moved the logic from above into an `Object` of predefined length getters that lives in `LayoutsUtil` but leave the `Layout` functions extensible by functions.

E.g., this block would look something more like
```
lengthGetter = LayoutsUtil.lengthGetters[branchMethod];
if (this._currentLayout === "Rectangular") {
    data = LayoutsUtil.rectangularLayout(
        ...,
        lengthGetter
    )
}
```

empress/support_files/css/empress.css
```diff
@@ -683,6 +683,10 @@ p.side-header button:hover,
     margin-left: 2em;
 }
 
+.warning {
+    color: #ff0000;
+}
```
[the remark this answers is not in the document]
gwarmstrong: Sure thing!

gibsramen:
Profiled the time to ultrametrify (officially coining this term) the EMP tree using `performance.now()`. Looks like, on average, the computation time of the ultrametric lengths is around 500+-50ms. This was for both rectangular & circular layouts.

Sandwiched [this block of code](https://github.com/gwarmstrong/empress/blob/8c3656872999474176dc4bbd16f1d0c98c48a9c6/empress/support_files/js/empress.js#L369-L384) per @gwarmstrong with time calls and took the difference.

empress/support_files/templates/side-panel.html
```diff
@@ -133,15 +133,43 @@
...
+             class="empress-input">
+    </p>
+    <p class="side-panel-notes indented">
+      The branch lengths shown correspond to the branch lengths in the original tree.
+    </p>
+    <p class="side-panel-notes warning">
+      WARNING: the 'Ignore branch lengths' and 'Make ultrametric' options display
+      modified branch lengths that can be useful for aesthetics. The resulting
+      trees should NOT be interpreted with respect to their branch lengths.
+    </p>
```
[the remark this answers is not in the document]
gwarmstrong: So I think I addressed (2) in https://github.com/biocore/empress/pull/444/commits/867eb96bf19c6a7329ff790e1263ba325b484f29

Is this https://github.com/biocore/empress/pull/444/commits/cd13f594ca996d0bec57009262e541aec266bd3a what you meant for 1?

empress/support_files/templates/side-panel.html
```diff
@@ -133,15 +133,43 @@
...
-  <p>
-    <label for="ignore-lengths-chk">Ignore branch lengths?</label>
-    <input id="ignore-lengths-chk" type="checkbox" class="empress-input">
-  </p>
-  <p class="side-panel-notes indented">
-    If checked, branch lengths are ignored when drawing the tree. Note that the
-    root node is never drawn with any length, regardless of if this
-    is checked or not.
+  <p class="side-panel-notes">
+    Change how the branch lengths are determined.
```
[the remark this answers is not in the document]
gwarmstrong: think https://github.com/biocore/empress/commit/867eb96bf19c6a7329ff790e1263ba325b484f29 resovles this?

tests/test-layouts-util.js
```diff
@@ -86,6 +86,28 @@ require([
...
+                5: 2,
+                2: 1,
+                1: 2,
+                11: 2,
+                0: 1,
+            };
+
+            var obs = LayoutsUtil.getUltrametricLengths(this.tree);
+            deepEqual(obs, expLengths);
+        });
```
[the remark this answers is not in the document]
gwarmstrong: Sure. So I just wrote a test in https://github.com/biocore/empress/pull/444/commits/c4801e52ca0c27ee3c7e510c60881697fa576cdf for  tree that was already ultrametric and asserted that its branch lengths were not changed. Makes sense since the "extend a branch until it is the same length as the deepest branch" should just keep the branches the same length (and cases of this were kind of already covered in the existing test).

While (seemingly) unlikely, it is also possible that "Ignore branch lengths" does not change someone's tree. So I think it might be worthwhile to open a separate issue that does check for whether the branch length transformation actually changes anything and give the user a warning based on that. But that is kind of contingent on how the `lengthGetter` discussion is resolved.

empress/support_files/js/empress.js
gwarmstrong: Let me know if this commit https://github.com/biocore/empress/pull/444/commits/80209027a1b69e1d08ebc1fd43ce7dd44d991a55 resolve this.

empress/support_files/js/layouts-util.js
gwarmstrong: should be resolved in https://github.com/biocore/empress/pull/444/commits/054de769fafa387b851fd8db1acb6de5d62c2cb8

empress/support_files/templates/side-panel.html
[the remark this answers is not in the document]
fedarko: It's closer to what I was thinking of, but a bit different. My preference (just a preference, I'm not too married to it) would be for the following changes, relative to how things look now:

1. Rather than `Layout options`, the header should just say `Branch lengths` (since clade sorting is also a type of layout option I guess)
2. Being consistent with the level of detail for each subsection -- my vote would be for just hiding the `Change how the branch lengths are determined.` text unless the warning is being shown. The reasons for this are A) I think this description kinda goes without saying and B) the other control panel sections structured similarly to this (e.g. `Clade sorting`; the stuff in the Settings tab, e.g. `Node circles`) don't have this sort of description, so it's a bit confusing.

... My comments are veering kinda close to bikeshedding, so none of this is a big deal :) This already looks great tbh

empress/support_files/templates/side-panel.html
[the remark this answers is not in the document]
fedarko: Yeah, (2) seems addressed. I think my original comment was kinda misleading -- by "replace" I think I meant not even showing the `Change how the branch lengths are determined.` text in the first place, and just sticking the warning in its place when that warning shows up. So the effective behavior would be the same, except for when "Normal" is selected in which case no description text would be shown.

Does that make sense / seem reasonable? This is described a bit in my comment above at https://github.com/biocore/empress/pull/444#discussion_r524919761, but the tldr is we currently only really use these "notes" for the headers of individual tabs (e.g. sample metadata coloring) and for warnings, not for additional sections within tabs.

I'm rereading that last paragraph now and I sound like a crazy person, sheesh. UH it's basically fine already lmao this is just me being picky :P

tests/test-layouts-util.js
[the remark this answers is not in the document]
fedarko: Thanks! That makes sense to me. Agreed that popping up these warnings is probably out-of-scope for this PR. Fortunately, the text documenting these branch length options does a good job describing things -- so if a user really tries to open up Empress with a tree with uniform-length nodes, it should hopefully be pretty obvious why nothing is changing when they click `Ignore branch lengths`.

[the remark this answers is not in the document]
fedarko: Opened #447, so in the future if we ever have enough free time to work on it then we can discuss that there (... writing "free time" at this point in the quarter makes me think of sitcom laugh track noises playing in the background, but you get the idea :P)

empress/support_files/js/layouts-util.js
```diff
@@ -32,6 +32,149 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
+    /**
+     * Sets the default length getter, if necessary. Otherwise passes through the length getter.
+     *
+     * @param {Function|null} lengthGetter determines the length at a given node index
+     * @param {Boolean} ignoreLengths indicates whether branch lengths should be ignored.
+     * @param {BPTree} tree Tree for setting branch lengths in default length getter
+     * @returns {Function} map from node index to branch length
+     * @private
+     */
+    function _determineLengthGetter(lengthGetter, ignoreLengths, tree) {
```
fedarko, line 166: Right now, my understanding is that -- in the actual execution of Empress -- `lengthGetter` will always be a function, and that it'll only be `null` during certain tests. Is this correct?

If so, I think (similarly to the stuff at https://github.com/biocore/empress/pull/444/files#r524571518) we should probably just cut out the middleman and remove this function (ideally within this PR, but if not that is ok). If this would explode like an intractable number of tests we may be able to figure out another way around that.

...Sorry for asking you to add docs for this function only to ask you to remove it a week later ._.

fedarko (changes requested):
Some requested changes; overall this is looking great. Exciting stuff :D :deciduous_tree: :evergreen_tree: :tanabata_tree:

empress/support_files/js/layouts-util.js
gwarmstrong: All good 😃

gwarmstrong: Covered in https://github.com/biocore/empress/pull/444/commits/fdff5e432ae597f84ec668acfd74b62f081cac93 ?

I could go one step further and remove all instances using the default `lengthGetter`, and it would also require re-ordering some arguments in all of the places, I believe.

fedarko: I think this looks fine as is. Thanks!

empress/support_files/js/side-panel-handler.js
```diff
@@ -83,16 +83,65 @@ define(["underscore", "Colorer", "util"], function (_, Colorer, util) {
...
+                    } else {
+                        scope.branchLengthWarningContainer.classList.add(
+                            "hidden"
+                        );
+                    }
+                }
+                // we want to make sure empress knows whether or not to ignore lengths
+                // at the moment; this is most critical for empress._collapseClade,
+                // since the lengths for the layout methods will be determined
+                // by setting empress.branchMethod
```
fedarko, line 128: Could you add a quick note pointing to https://github.com/biocore/empress/issues/448, and mentioning why this works for ultrametric trees? Just so we don't accidentally break this in the future

empress/support_files/templates/side-panel.html
```diff
@@ -131,18 +131,49 @@
...
+    </p>
+    <p>
+      <label for="ignore-lengths-chk">Ignore branch lengths</label>
+      <input type="radio" id="ignore-lengths-chk" name="branches-radio" value="ignore"
+             class="empress-input">
+    </p>
+    <p class="side-panel-notes indented">
+      If checked, branch lengths are ignored when drawing the tree. Note that the
+      root node is never drawn with any length, regardless of if this
+      is checked or not.
```
fedarko, line 160:
```suggestion
      If checked, all branches are drawn with the same length. Note that the
      root node is never drawn with any length, regardless of if this
      is checked or not.
```

Feel free to change the wording -- I realized just now that saying this "ignores" branch lengths might be slightly misleading, since technically the make ultrametric setting also ignores the branch lengths in the sense that it modifies them. But this is extremely minor so not a big deal

fedarko (approved):
Two very, very small comments (one for dev documentation and one for user documentation). Basically good to merge as is.

empress/support_files/js/side-panel-handler.js
fedarko, line 128:
```suggestion
                // we want to make sure empress knows whether or not to ignore lengths
                // at the moment; this is most critical for empress._collapseClade,
                // since the lengths for the layout methods will be determined
                // by setting empress.branchMethod. Note that we don't pass information
                // about whether or not the tree is ultrametric; this is ok for now,
                // because lengths are just used to determine the "depth" of collapsed
                // clades, and by definition all tips in an ultrametric tree end at the
                // same "length" from the root of the tree. However, we should fix
                // this in the future; see https://github.com/biocore/empress/issues/448.
```

fedarko:
In the interest of getting a release out today, @ElDeveloper and I decided to just make the two small changes and merge this in. Thanks so much @gwarmstrong and @gibsramen for your work on this!!! :straight_ruler: :evergreen_tree:

gwarmstrong:
Thanks @fedarko !!

ElDeveloper:
# 📏 🎄

PR #450, opened 2020-11-17
Disable browser autocomplete in node search menu

### LINKED ISSUE #445: Set autocomplete="off" on the search bar
This is a minor issue, but when using the node search bar Chrome attempts to be helpful and tries to show autocomplete options -- this gets in the way of the autocomplete menu that Empress already shows:

autocmpokple

From https://gist.github.com/niksumeiko/360164708c3b326bd1c8, it looks like using `autocomplete="off"` on the search bar `<input>` hides the browser's autocomplete from popping up, letting Empress' actually useful autocomplete be shown.

acomp2

fedarko:
Really tiny change -- just closes #445.

### DIFF
--- a/empress/support_files/templates/side-panel.html
+++ b/empress/support_files/templates/side-panel.html
@@ -1,8 +1,11 @@
 <!-- The Header bar -->
 <div id="autocomplete-container">
   <p class="side-header">
+    <!-- Attempt to stop the browser's autocomplete suggestions, since Empress
+         shows its own autocomplete menu (see #445 on GitHub)
+    -->
     <input type="text" id="quick-search" placeholder="Search by node name..."
-           style="flex-grow: 1;" class="empress-input">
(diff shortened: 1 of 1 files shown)


kwcantrell:
Thanks @fedarko! I don't chrome so I did know this was a thing. One quick comment, do you think we should add that to the empress-input css? That way this wont be an issue if we add more input boxes.

fedarko:
That's a good point -- looked it up and I don't think there's a way to disable the autocomplete stuff with CSS (https://stackoverflow.com/a/2186406/10730311), since the `autocomplete` property is defined in the HTML. I think the node search bar is the only `text` input that is used in Empress, so we should be safe for now, but maybe we could try using something like [this solution](https://stackoverflow.com/a/8651838/10730311) in the future if we end up having many `text` inputs?

kwcantrell:
Okay, sounds good. Thanks @fedarko

fedarko:
Thanks @kwcantrell! :car:

PR #456, opened 2020-11-25
NEW warn when layout options (approximately) do not change branch lengths

### LINKED ISSUE #447: Show a warning when a branch length option (e.g. ignore lengths; make ultrametric) doesn't actually change the tree
Discussed with @gwarmstrong in #444 at https://github.com/biocore/empress/pull/444#discussion_r520968631. Basically, if the user has a tree that is already ultrametric, then clicking the "make ultrametric" option should show a warning -- and if the tree's nodes all already have the same length, then clicking "ignore branch lengths" should also show a warning (I thiiink that even if the uniform length is 2 or something the drawing will still look identical to the uniform length of 1 used under the hood in the layout computations, due to how the camera/etc are scaled).

I think the most likely cause of users running into this situation in practice will be Qemistree users coming in with ultrametric trees and wondering why the `Make ultrametric` option doesn't change anything. Fortunately, the UI for these options (esp. with the improvements coming soon in #444) should make this situation pretty obvious, so I don't think this is a very high-priority issue.

gwarmstrong:
Fixes #447 .

This PR follows up on #444 to raise a toast message if the tree is not changed by options like "make ultrametric" and "ignore branch lengths".

When working on a tree from qemistree (thanks @gibsramen ), I found that the branch lengths could be modified by small amounts by the make ultrametric code, but this resulted in no visible difference between the two trees. As a result, I introduced a tolerance for how similar branch lengths should be in order to make this warning. Though this could result in a warning being falsely raised in the case where the branch lengths in the tree are all very small.

Screen Shot 2020-11-25 at 11 05 42 AM

### DIFF
--- a/empress/support_files/js/layouts-util.js
+++ b/empress/support_files/js/layouts-util.js
@@ -373,7 +373,8 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             }
             anyDifferent =
                 // this expression assumes nodeLen and tree.length will both be positive
-                anyDifferent || Math.abs((tree.length(prepos) / nodeLen) - 1) > TOL;
+                anyDifferent ||
+                Math.abs(tree.length(prepos) / nodeLen - 1) > TOL;
         }
         if (!anyDifferent && checkLengthsChange) {
             noLengthsChangedMsg();
--- a/empress/support_files/js/layouts-util.js
+++ b/empress/support_files/js/layouts-util.js
@@ -160,6 +160,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
      * @param {String} methodName Method for determing branch lengths.
      *                            One of ("ultrametric", "ignore", "normal").
      * @returns {Boolean} Indicates whether the check should be performed.
+     * @throws {Error} If methodName is invalid.
      */
     function shouldCheckBranchLengthsChanged(methodName) {
         var methods = {
@@ -175,7 +176,8 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
     }
 
     var NO_LENGTHS_CHANGED_MSG =
-        "No branch lengths modified by layout options.";
+        "It doesn't look like any branch lengths were changed " +
+        "by the current method of modifying branch lengths.";
     var NO_LENGTHS_CHANGED_DURATION = 3000;
     var TOL = 0.000001;
 
--- a/empress/support_files/js/layouts-util.js
+++ b/empress/support_files/js/layouts-util.js
@@ -159,7 +159,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
      *
      * @param {String} methodName Method for determing branch lengths.
      *                            One of ("ultrametric", "ignore", "normal").
-     * @returns {Function} A function that maps node indices to branch lengths.
+     * @returns {Boolean} Indicates whether the check should be performed.
      */
     function shouldCheckBranchLengthsChanged(methodName) {
         var methods = {
--- a/empress/support_files/js/layouts-util.js
+++ b/empress/support_files/js/layouts-util.js
@@ -373,10 +373,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             if (maxWidth < xCoord[node]) {
                 maxWidth = xCoord[node];
             }
-            anyDifferent =
-                // this expression assumes nodeLen and tree.length will both be positive
-                anyDifferent ||
-                Math.abs(tree.length(prepos) / nodeLen - 1) > TOL;
+            anyDifferent = updateAnyDifferent(anyDifferent, nodeLen, tree, prepos);
         }
         if (!anyDifferent && checkLengthsChange) {
             noLengthsChangedMsg();
@@ -619,8 +616,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
 
             var nodeLen = lengthGetter(prepos);
             radius[node] = radius[parent] + nodeLen;
-            anyDifferent =
-                anyDifferent || Math.abs(nodeLen - tree.length(prepos)) > TOL;
+            anyDifferent = updateAnyDifferent(anyDifferent, nodeLen, tree, prepos);
         }
         if (!anyDifferent && checkLengthsChange) {
             noLengthsChangedMsg();
@@ -748,6 +744,23 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
     }
 
     /**
+     *
+     * @param {Boolean} anyDifferent Indicates if any branches have different
+     *                               transformed lengths that differ from
+     *                               their original length in the tree.
+     * @param {Float} branchLen Transformed length of the branch.
+     * @param {BPTree} tree That has (potentially) been transformed.
+     * @param {Number} n Node index in tree.
+     * @returns {Boolean} Indicates if any of the branches checked in the tree
+     *                    so far are different than the transformed length.
+     */
+    function updateAnyDifferent(anyDifferent, branchLen, tree, n) {
+        anyDifferent =
+            anyDifferent || Math.abs(branchLen - tree.length(n)) > TOL;
+        return anyDifferent;
+    }
+
+    /**
      * Unrooted layout.
      *
      * REFERENCES
@@ -837,9 +850,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             minX = Math.min(minX, x2Arr[node]);
             maxY = Math.max(maxY, y2Arr[node]);
             minY = Math.min(minY, y2Arr[node]);
-
-            anyDifferent =
-                anyDifferent || Math.abs(nodeLen - tree.length(n)) > TOL;
+            anyDifferent = updateAnyDifferent(anyDifferent, nodeLen, tree, n);
         }
         if (!anyDifferent && checkLengthsChange) {
             noLengthsChangedMsg();
--- a/empress/support_files/js/layouts-util.js
+++ b/empress/support_files/js/layouts-util.js
@@ -619,7 +619,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             var nodeLen = lengthGetter(prepos);
             radius[node] = radius[parent] + nodeLen;
             if (checkLengthsChange && !anyDifferent) {
-                anyDifferent = isTransformedLenDifferent(nodeLen, tree, n);
+                anyDifferent = isTransformedLenDifferent(nodeLen, tree, prepos);
             }
         }
         if (checkLengthsChange && !anyDifferent) {
@@ -758,7 +758,7 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
      *                    different from its original length.
      */
     function isTransformedLenDifferent(branchLen, tree, n) {
-        return Math.abs(branchLen - tree.length(n)) > TOL;
+        return Math.abs(tree.length(n) / branchLen - 1) > TOL;
     }
 
     /**
--- a/empress/support_files/js/layouts-util.js
+++ b/empress/support_files/js/layouts-util.js
@@ -608,7 +608,6 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
 
             var nodeLen = lengthGetter(prepos);
             radius[node] = radius[parent] + nodeLen;
-            // anyDifferent = anyDifferent || nodeLen !== tree.length(prepos);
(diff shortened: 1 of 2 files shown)


empress/support_files/js/layouts-util.js
```diff
@@ -568,6 +608,12 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
 
             var nodeLen = lengthGetter(prepos);
             radius[node] = radius[parent] + nodeLen;
+            // anyDifferent = anyDifferent || nodeLen !== tree.length(prepos);
```
gwarmstrong, line 611:
```suggestion
```

empress/support_files/js/layouts-util.js
```diff
@@ -776,6 +827,13 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             minX = Math.min(minX, x2Arr[node]);
             maxY = Math.max(maxY, y2Arr[node]);
             minY = Math.min(minY, y2Arr[node]);
+
+            // anyDifferent = anyDifferent || nodeLen !== tree.length(n);
```
gwarmstrong, line 831:
```suggestion
```

empress/support_files/js/layouts-util.js
```diff
@@ -154,6 +154,31 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
     }
 
+    /**
+     * Determines whether a given method should check if branch lengths were changed.
+     *
+     * @param {String} methodName Method for determing branch lengths.
+     *                            One of ("ultrametric", "ignore", "normal").
+     * @returns {Function} A function that maps node indices to branch lengths.
+     */
+    function shouldCheckBranchLengthsChanged(methodName) {
```
kwcantrell, line 164: this looks like it returns a boolean?

empress/support_files/js/layouts-util.js
```diff
@@ -334,6 +364,11 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             if (maxWidth < xCoord[node]) {
                 maxWidth = xCoord[node];
             }
+            anyDifferent =
+                anyDifferent || Math.abs(nodeLen - tree.length(prepos)) > TOL;
```
kwcantrell, line 368: perhaps instead of using an absolute number for the tolerance, we can use a percentage changed? Using a percentage changed instead of an absolute number would "normalize" the tolerance across all branch lengths.

empress/support_files/js/layouts-util.js
```diff
@@ -776,6 +826,12 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             minX = Math.min(minX, x2Arr[node]);
             maxY = Math.max(maxY, y2Arr[node]);
             minY = Math.min(minY, y2Arr[node]);
+
+            anyDifferent =
+                anyDifferent || Math.abs(nodeLen - tree.length(n)) > TOL;
+        }
+        if (!anyDifferent && checkLengthsChange) {
+            util.toastMsg(NO_LENGTHS_CHANGED_MSG, NO_LENGTHS_CHANGED_DURATION);
```
kwcantrell, line 834: might want to make this a helper function since it this same block is used in multiple functions.

kwcantrell (commented):
Thanks @gwarmstrong! This looks good. I just have a couple minor comments.

empress/support_files/js/layouts-util.js
gwarmstrong: Fixed here https://github.com/biocore/empress/pull/456/commits/1e86bc9b7372f505a04c7dd653dbd3e40b8779ba

empress/support_files/js/layouts-util.js
gwarmstrong: Good idea. Fixed here: https://github.com/biocore/empress/pull/456/commits/22f12fe1920e6d4afb5702399b47d1de0edbbe20

empress/support_files/js/layouts-util.js
gwarmstrong: Changed here: https://github.com/biocore/empress/pull/456/commits/77c03da0e439f7282914719c81b5dd08cadd6c25

empress/support_files/js/layouts-util.js
```diff
@@ -334,6 +371,13 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             if (maxWidth < xCoord[node]) {
                 maxWidth = xCoord[node];
             }
+            anyDifferent =
+                // this expression assumes nodeLen and tree.length will both be positive
+                anyDifferent ||
+                Math.abs(tree.length(prepos) / nodeLen - 1) > TOL;
```
fedarko, line 377: I have three suggestions:

1. It's not going to make a difference except (maaaybe) for huge trees, but division is kind of expensive. It seems to me like the final line here could be rewritten simpler as
   
   ```js
   Math.abs(tree.length(prepos) - nodeLen) > TOL;
   ```
   This should be more efficient and easier to read (unless I'm missing something).

2. If `checkLengthsChange` is false, then we don't need to bother with updating `anyDifferent` at all. Therefore, I suggest that we wrap the `anyDifferent = ` statement inside an `if (checkLengthsChange)` branch, so that we don't spend time updating this unless it's necessary.

3. Even if `checkLengthsChange` is true, then we can stop trying to update `anyDifferent` as soon as we find one case where the length is different (i.e. as soon as we set `anyDifferent` to true). Therefore, I suggest that we modify the aforementioned if statement to be something like `if (checkLengthsChange && !anyDifferent)`.

Does this all make sense? Let me know if I missed something.

empress/support_files/js/layouts-util.js
```diff
@@ -568,6 +617,11 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
 
             var nodeLen = lengthGetter(prepos);
             radius[node] = radius[parent] + nodeLen;
+            anyDifferent =
+                anyDifferent || Math.abs(nodeLen - tree.length(prepos)) > TOL;
+        }
+        if (!anyDifferent && checkLengthsChange) {
+            noLengthsChangedMsg();
```
fedarko, line 624: In order to limit code reuse, it might be worth moving the line that updates anyDifferent to a separate function, so that way future modifications to that are just stored in one place.

(...but I get it if you'd prefer to not bother with this for now)

empress/support_files/js/layouts-util.js
```diff
@@ -154,6 +154,38 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
+     * Determines whether a given method should check if branch lengths were changed.
+     *
+     * @param {String} methodName Method for determing branch lengths.
+     *                            One of ("ultrametric", "ignore", "normal").
+     * @returns {Boolean} Indicates whether the check should be performed.
+     */
+    function shouldCheckBranchLengthsChanged(methodName) {
+        var methods = {
+            ultrametric: true,
+            ignore: true,
```
fedarko, line 167: One slight problem with the current approach is that it only works when all the node lengths are 1 - if all the nodes have lengths of 2 or something, then this isn't identified (even though these cases look identical). This is discussed a bit in #447.

...That being said these corner cases are rare and not very important, so I'm cool with not worrying about this.

empress/support_files/js/layouts-util.js
```diff
@@ -154,6 +154,38 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
...
+        };
+        if (methodName in methods) {
+            return methods[methodName];
+        } else {
+            throw "Invalid method: '" + methodName + "'.";
+        }
+    }
+
+    var NO_LENGTHS_CHANGED_MSG =
+        "No branch lengths modified by layout options.";
```
fedarko, line 178:
```suggestion
        "It doesn't look like any branch lengths were changed by the current method of modifying branch lengths.";
```

Rationale for this:

1. Gives us leeway, indicating that these messages aren't perfect (both due to tolerance things and due to missing the ignore lengths stuff i mentioned above)

2. By saying "the current method", we avoid confusing the user -- these toast messages show up every time the user changes the layout while an "alternate branch length method" (eg ultrametric) is selected, so wording things this way makes it clearer that the actual layout algorithm isn't the reason for these messages.

fedarko: (the styling code will probably want this string to be broken onto two separate lines, sorry if it yells at you)

fedarko (changes requested):
I have a few small changes that I think would be useful, but none of them are very large. Thank you for getting this ready, and let me know if I can help out with anything!

empress/support_files/js/layouts-util.js
gwarmstrong, line 178:
```suggestion
        "It doesn't look like any branch lengths were changed " +
        "by the current method of modifying branch lengths.";
```

empress/support_files/js/layouts-util.js
gwarmstrong: What evidence do you have for division being expensive in js?
I found these benchmarks https://www.akselipalen.com/2021/01/13/benchmark-of-elementary-mathematical-operations-in-node-js/
and ran them locally and got results similar to the author's.

The way you wrote the line is what I had originally, but see the comment in my initial post and @kwcantrell 's [follow-up](https://github.com/biocore/empress/pull/456#discussion_r533660491) regarding the insufficiency of absolute tolerance (maximum difference). The relative tolerance (i.e., compare difference of the division between branch lengths) proposed by Kalen solves the problem where, e.g., "make ultrametric" makes a large relative change to small branch lengths. E.g., if all branches in an unbalanced tree of max depth3 are 0.01, and tolerance is 0.1, then "make ultrametric" would change a branch length by at most 0.03, which is less than the tolerance and would trigger the "no changes" warning. But if we do the division, we realize that a branch changing from 0.01 to 0.03 is tripled in length, so we do not want to raise the warning.

Further, I think we can solve the "all branch lengths are 2" problem you [mentioned above](https://github.com/biocore/empress/pull/456/files#r564135982) with the division approach by changing the line to something like

```
                Math.abs(tree.length(prepos) / nodeLen - lengthFactor) > TOL;
```

Where `lengthFactor` is determined by the first call to `tree.length(prepos) / nodeLen`.

gwarmstrong: Good points on the logic surrounding `checkLengthsChange` and `anyDifferent`!

empress/support_files/js/layouts-util.js
gwarmstrong: Updated here: https://github.com/biocore/empress/pull/456/commits/cec5ed6dadaa2df2eb49b60e18a77396cb9fcff3

empress/support_files/js/layouts-util.js
gwarmstrong: Changed the control flow here: https://github.com/biocore/empress/pull/456/commits/000a5b841cbf1c474d6ebaf40c22ff2000561ed1

fedarko: The point about relative tolerance makes a lot of sense—sorry for missing that discussion.

Re: runtime of arithmetic operations, I was thinking along the lines of https://en.wikipedia.org/wiki/Computational_complexity_of_mathematical_operations. That article is language-independent and JS is high-level enough that I can imagine that division vs. subtraction probably wouldn't make much of a difference, similarly to the benchmarks.

In tandem with the control flow changes, I imagine that prematurely optimizing this is probably not worth the effort right now... So I'm cool with leaving this as is (and if people complain about this being a bottleneck with like trillion-node trees we can cross that bridge when we get to it ;) Thanks for your insight on this!

fedarko (changes requested):
This is basically fine. I have one suggested change to the any-different-updating function (not passing any different, which simplifies things a bit), and we may want to change the checking back from subtraction to division... but not a huge deal either way. Sorry for nitpicking things here. I'm ok with merging this in provided these changes are addressed (either via the suggestions enclosed in this review, or in another way if you'd prefer).

Thanks so much! Sorry for taking a few days to get to this.

empress/support_files/js/layouts-util.js
```diff
@@ -334,6 +373,12 @@ define(["underscore", "VectorOps", "util"], function (_, VectorOps, util) {
             if (maxWidth < xCoord[node]) {
                 maxWidth = xCoord[node];
             }
+            if (checkLengthsChange && !anyDifferent) {
+                anyDifferent = isTransformedLenDifferent(nodeLen, tree, n);
```
fedarko, line 377:
```suggestion
                anyDifferent = isTransformedLenDifferent(nodeLen, tree, prepos);
```

Looks like I broke this sorry ._.  tested locally and confirmed this change should fix things with the rectangular layout

fedarko: Committing change due to discussion with @gwarmstrong

PR #464, opened 2020-12-11
Add option to adjust gap between tree and first barplot layer

### LINKED ISSUE #453: Allow users to adjust amount of separation between first barplot layer and the farthest tip
Brought up by @gwarmstrong at this morning's meeting.

Having the ultrametric PR in, as well as #366 and #308 in, definitely reduces the need for this -- but if demand for this remains in the future we may want to work on it.

For reference, a solution to this issue would involve adding an extra `<input type="number">` to the barplot section of the side panel that allows users to configure the `1.1` value listed here (or something similar -- there is probably a more user-friendly way of exposing this functionality):

https://github.com/biocore/empress/blob/ae42a086bbc664529b5c02b66e8f9b4bab5d9a5d/empress/support_files/js/empress.js#L1490-L1499

gibsramen:
Closes #453 

Adds a new element to the side panel that allows users to manually set the distance between the tree and the first barplot layer. Currently works by specifying a percentage increase to `_maxDisplacement`. ~Default value is 10 (1 + 10% = 1.1) to keep with current default behavior.~ Default value changed to 5 as we decided to decrease the default gap.

OLD: `var maxD = 1.1 * this._maxDisplacement;`

NEW: `var maxD = (1 + this._displacementFrac) * this._maxDisplacement;`

~A new function `Empress.prototype.changeBorderGap` just divides the input value by 100. This can likely be rewritten to not use a whole function - depends on what people think looks more clean.~

## Current Status

- [x] Pick a better display name than `Separation length`
- [x] Update relevant documentation in `empress.js`
- [x] Disallow negative values

gap_movie

### DIFF
--- a/empress/support_files/js/barplot-panel-handler.js
+++ b/empress/support_files/js/barplot-panel-handler.js
@@ -55,6 +55,7 @@ define([
         this.borderContent = document.getElementById("barplot-border-content");
         this.borderCheckbox = document.getElementById("barplot-border-chk");
         this.borderOptions = document.getElementById("barplot-border-options");
+        this.borderGapInput = document.getElementById("barplot-custom-gap");
         this.borderColorPicker = document.getElementById(
             "barplot-border-color"
         );
@@ -129,6 +130,7 @@ define([
 
         // And define behavior for how to add in barplots
         this.updateButton.onclick = function () {
+            scope.empress.changeBorderGap(scope.borderGap);
             scope.empress.drawBarplots(scope.layers);
         };
 
@@ -142,6 +144,7 @@ define([
         // ... and to having a length of whatever the default barplot layer
         // length divided by 10 is :)
         this.borderLength = BarplotLayer.DEFAULT_LENGTH / 10;
+        this.borderGap = 10;
 
         // Now, initialize the border options UI accordingly
         this.initBorderOptions();
@@ -158,8 +161,17 @@ define([
             }
         };
 
+        // Define behavior for border gap input
+        this.borderGapInput.value = this.borderGap;
+        console.log("Default set!");
+        console.log(this.borderGapInput.value);
+        $(this.borderGapInput).change(function () {
+            scope.borderGap = this.value;
+        });
+
         // To get things started off with, let's add a layer
         this.addLayer();
+
     }
 
     /**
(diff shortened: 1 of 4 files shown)


empress/support_files/templates/side-panel.html
```diff
@@ -221,6 +221,10 @@
       </p>
     </div>
     <div id="barplot-border-content" class="hidden">
+      <p>
+        <label for="barplot-custom-gap">Separation length</label>
```
ElDeveloper, line 225: Only thing I can come up with is to make this a bit more verbose.

```suggestion
        <label for="barplot-custom-gap">Distance between tree and barplots</label>
```

empress/support_files/js/barplot-panel-handler.js
```diff
@@ -143,6 +144,16 @@ define([
         // length divided by 10 is :)
         this.borderLength = BarplotLayer.DEFAULT_LENGTH / 10;
 
+        // Initialize default spacing between tree and first barplot layer
+        // as well as change behavior.
+        this.borderGap = 10;
```
ElDeveloper, line 149: Per our call earlier today, the suggestion is to make this space small by default - just making a note here to have a record.

ElDeveloper (approved):
Thanks so much @gibsramen

fedarko (approved):
Thanks so much @gibsramen!!!

Per discussion with Gibs earlier today I modified this slightly in the following ways:

1. Use the same "units" for the initial gap as for barplot lengths, making things a bit more consistent for the user. The default gap value is now 10, which matches the default barplot border length.

2. Store the current gap value only in `BarplotPanelHandler`, and have `Empress` retrieve this value when needed rather than keeping track of it in `Empress._displacementFrac`. The rationale for this is that it reduces the amount of data `Empress` needs to keep track of, and matches some of the other ways in which the barplot functionality works (e.g. how the border color / length are also retrieved from the `BarplotPanelHandler`).

Provided that these changes look good, I think we can merge this in. Sorry for taking so long to get to this.

ElDeveloper:
@gibsramen, if this looks good to you let me know and I'll be happy to merge 
this to the master branch.

gibsramen:
LGTM :)

ElDeveloper:
Thanks so much both!

PR #465, opened 2020-12-15
GitHub actions

kwcantrell:
this pr "translates" our travis yml file to a github actions yml file.

Its currently setup to trigger when ever a pull request on the master branch is create or when pull requests are updated.

I currently have the step that adds McHelper commented out. We should be able to uncomment it out once we resolve the issue with McHelper.


gibsramen:
Thanks for this, @kwcantrell! One thing I just realized is that we will probably also want to update the README to remove the Travis build embedding (https://travis-ci.org/biocore/empress). Do you know if there's an equivalent for GH actions?

.github/workflows/main.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Set up conda
+        uses: s-weigand/setup-conda@v1
+        with:
+          update-conda: true
+          python-version: 3.6
+          conda-channels: anaconda, conda-forge
+
+      - name: Create Qiime enviroment
+        run: |
+          wget https://data.qiime2.org/distro/core/qiime2-2020.6-py36-linux-conda.yml
```
wasade, line 55: Why 2020.6?

wasade (commented):
I hadn't seen `conda run` before, that is nifty. I took a different approach with microsetta public API (https://github.com/biocore/microsetta-public-api/blob/master/.github/workflows/python-package-conda.yml#L37-L40), where the environment is activated, but this seems much much cleaner!

.github/workflows/main.yml
kwcantrell: I was using the latest version (https://raw.githubusercontent.com/qiime2/environment-files/master/latest/staging/qiime2-latest-py36-linux-conda.yml) and it was working fine but all of a sudden it started to break the build and causing it to fail. So I switched to 2020.6 simply because that is the version I am using on my local machine.

kwcantrell:
> Thanks for this, @kwcantrell! One thing I just realized is that we will probably also want to update the README to remove the Travis build embedding at the top. Do you know if there's an equivalent for GH actions?

Thats a good point, I didn't even think about that. The CONTRIBUTING.md also needs to be updated.

ElDeveloper (approved):

ElDeveloper:
Thanks @kwcantrell, I agree with @gibsramen regarding updating the contribution documentation.

gibsramen:
Looks like we can update with GitHub's own [status badges](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge).

kwcantrell:
Thanks everyone for the feedback. The PR should now be good to merge.

ElDeveloper:
Thanks so much @kwcantrell!! 🚀 

Let's start with this config file and see what else is revealed along the way 👍

Issue #462, opened 2020-12-11
Speed up how node names are processed on initialization


fedarko:
Just making a record of this here so we don't forget it. (@kwcantrell and @ElDeveloper did all the work of finding this!!!)

Relates to the tree shown in [this twitter thread](https://twitter.com/yoshikivb/status/1336848912364175362). Basically, when there are a ton of node names, this block of code:

https://github.com/biocore/empress/blob/1b7a8d3a8da2c2a980ffc12f0f04ea6220d7e5e5/empress/support_files/js/empress.js#L463-L472

... slows things down a lot, because string comparisons are slow.

There are many possible ways we could handle this (better algorithms / data structures in JS, preprocessing stuff in Python, etc.) -- it's probably most important right now just to implement _something_ that fixes this, since this is a pretty significant bottleneck.

PR #466, opened 2020-12-16
remove sort/binarysearch

kwcantrell:
This addresses #462.

Here I removed the sorting of node names/filtering non-uniq names in order to speed up the load time for larger trees. In addition, I know perform a basic linear search when filling the name suggestion menu.


kwcantrell:
Github actions worked!

fedarko (approved):
Looks good to me, thanks @kwcantrell! (Sorry for taking so long to get to this...)

PR #468, opened 2020-12-16
Rerun GitHub Actions on push events to master; update dev documentation to mention a weird thing with JS tests

### LINKED ISSUE #467: It looks like GH actions jobs aren't getting triggered when we merge stuff into master
See https://github.com/biocore/empress -- although the CLI PR had a (successful) job triggered, it didn't get rerun when it was merged in to master. This means that the CI badge in the README still shows up as `no status`.

I think we can fix this by adding `push` to the YML file (alongside `pull_request`) as one of the actions that triggers a build.

fedarko:
Should fix #467 (I think ... I don't really have a way of testing this).

Also updates `CONTRIBUTING.md` based on discussion with @gibsramen (we figured out the reason #464 was failing, and the errors being emitted from that reason were pretty obscure -- hopefully this helps clarify things for future devs).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,11 +34,16 @@ make test
 If you just want to run the Python or JavaScript tests, you can run
 `make pytest` or `make jstest` respectively.
 
-### Important note about JavaScript tests
+### Notes about JavaScript tests
 If you add HTML elements to EMPress' HTML code that the JavaScript code relies
 on, you will also need to add these elements to the `tests/index.html` file.
 Ideally this would not be necessary, but for now it is. (Failing to do this can
 cause obscure error messages from the JS tests.)
+
+Also, loading the test `index.html` fails on Firefox for some of the `Legend`
+tests: see https://github.com/biocore/empress/issues/435. To circumvent this,
+if you need to load the tests in the browser (rather than using puppeteer),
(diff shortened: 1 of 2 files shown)


ElDeveloper (approved):

kwcantrell:
I did not include the push event because github actions in order to save computation time on the github servers. But it makes sense that it is needed in order to update the status badge. Plus, it will also rerun the on the merged code and check for any weird errors that might pop up.

CONTRIBUTING.md
```diff
@@ -34,6 +34,12 @@ make test
 If you just want to run the Python or JavaScript tests, you can run
 `make pytest` or `make jstest` respectively.
 
+### Important note about JavaScript tests
+If you add HTML elements to EMPress' HTML code that the JavaScript code relies
+on, you will also need to add these elements to the `tests/index.html` file.
+Ideally this would not be necessary, but for now it is. (Failing to do this can
+cause obscure error messages from the JS tests.)
```
kwcantrell, line 41: Do you think we should also include a note about issue 435 that causes the legend test to fail if you load tests/index.html into a firefox browser?

fedarko: That makes sense. Just added a note.

kwcantrell:
Thanks @fedarko!

Issue #470, opened 2020-12-17
Can Empress._to_dict be made public?


gwarmstrong:
The `empress.core.Empress._to_dict` method is really handy for serializing an empress object, and we would like to do this on https://github.com/biocore/microsetta-public-api/ .

As noted by @wasade on https://github.com/biocore/microsetta-public-api/pull/84 it is not great to call a private method. Is there any reason this method should not be used as, or made into a public method of the `core.Empress` object?

kwcantrell:
I don't see any issue with making Empress._to_dict() public. The original purpose of the empress python object was to create the html page so there was no need to expose the Empress._to_dict() function as using it would be outside the scope of the object. However, it might be better to create a wrapper function for it as Empress._to_dict() returns a dict which contains references to internal variables. So before we return the dict, we would want to duplicated the objects first.

ElDeveloper:
+1, this is a good idea and there shouldn't be any issues with that.

gwarmstrong:
> So before we return the dict, we would want to duplicated the objects first.

Maybe something like a `copy=True`

fedarko:
Looks like #472 addressed this, so I think we can close this (pls yell at me if not)

PR #472, opened 2020-12-21
make _to_dict() public

kwcantrell:
In response to #470.

Rather than adding a parameter to return a copied version of the empress dictionary object, I decided to provided a warning in the doc string stating that to_dict returns references to internal objects. As @gwarmstrong mentioned in #470, its seems kind of wasteful to copy the object and since python does not technically have private variables I decided to simply return the dictionary object as is.


ElDeveloper (approved):
Looks great, thanks @kwcantrell!

PR #476, opened 2021-01-30
REL: bump version number to 1.1.0-dev, and also add a PyPI version number badge while we're at it

fedarko:


ElDeveloper:
Thanks @fedarko

PR #481, opened 2021-02-03
Add optional abbreviated parameters for the standalone CLI; add some more context to the README

fedarko:
This makes the CLI a bit easier to use if you're in a hurry, since potentially generating a visualization can be as short as:

```bash
$ empress tree-plot -t tree.nwk -o viz
```

I also updated the README a bit to document this, and to make it a bit fancier in general. (One of the changes is consistently using hyphens in filenames for some of the early examples. I didn't modify the QIIME 2 tutorial since the filenames there [IIRC] were designed to be consistent with Q2's tutorial files, some of which have underscores instead of hyphens.)

Not all of the parameters in the standalone CLI have short equivalents (e.g. `--filter-extra-samples` will still have to be written out in full). I think this is reasonable, since arguably part of the rationale for many of these parameters is requiring the user to acknowledge e.g. "yes I recognize that I am asking the program to filter out certain samples right now".

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,20 +53,18 @@ The standalone version of Empress takes the following filetypes as inputs. (Note
 ### Example standalone usage
 
 ```bash
-# Using "long" parameter names
+# Option 1: Using "long" parameter names
 empress tree-plot \
     --tree tree.nwk \
     --feature-metadata feature-metadata.tsv \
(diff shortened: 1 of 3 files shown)


gibsramen (approved):
LGTM :)

Issue #371, opened 2020-09-03
Support only coloring by the most frequent values in a field


fedarko:
I think this would mainly be useful for categorical feature metadata tree coloring / barplots, but it may also be useful for sample metadata barplots. Basically, this would have the user specify some k (let's say k = 10), and then this would only assign the 10 most common values colors and then map all other values to "Other" or something. Would be a nice way of making plots with massive amounts of unique values more interpretable (e.g. when coloring by species / genus in Empress, it's usually really hard to see what's going on because there are just so many different species / genera present even in moderately-sized 16S datasets).

I don't think this would be (easily) applicable to sample metadata tree coloring or animations, since those work by figuring out unique portions of the tree to certain groups -- and subsuming things into an "Other" group would probably cause a bunch of misleading "Other" blobs to get highlighted on the tree.

Idea (and "k" notation :) from @gibsramen.

fedarko:
It would also be nice to allow the user to specify certain values that should be excluded (and always included in `Other`) -- for example, `s__`, `Unspecified`, etc.

Issue #337, opened 2020-08-18
Identify common (sample) metadata values, and replace them with unique integers


fedarko:
@kwcantrell brought this up in this morning's meeting. Currently sample metadata is stored as follows in the HTML (this is from the moving pictures dataset, formatting modified for ease of reading):

```js
["barcode-sequence", "body-site", "year", "month", "day", "subject", "reported-antibiotic-usage", "days-since-experiment-start"],
[["AGTGCGATGCGT", "gut", "2009.0", "3.0", "17.0", "subject-1", "No", "140.0"],
 ["ATGGCAGCTCTA", "gut", "2008.0", "10.0", "28.0", "subject-2", "Yes", "0.0"],
 ["CTGAGATACGCG", "gut", "2009.0", "1.0", "20.0", "subject-2", "No", "84.0"],
 ["CCGACTGAGATG", "gut", "2009.0", "3.0", "17.0", "subject-2", "No", "140.0"],
 ["CCTCTCGTGATC", "gut", "2009.0", "4.0", "14.0", "subject-2", "No", "168.0"],
 ["ACACACTATGGC", "gut", "2009.0", "1.0", "20.0", "subject-1", "No", "84.0"],
 ["ACTACGTGTGGT", "gut", "2009.0", "2.0", "17.0", "subject-1", "No", "112.0"],
 ["AGCTGACTAGTC", "gut", "2008.0", "10.0", "28.0", "subject-1", "Yes", "0.0"],
 ["ACGATGCGACCA", "left palm", "2009.0", "1.0", "20.0", "subject-1", "No", "84.0"],
 ["AGCTATCCACGA", "left palm", "2009.0", "2.0", "17.0", "subject-1", "No", "112.0"],
 ["ATGCAGCTCAGT", "left palm", "2009.0", "3.0", "17.0", "subject-1", "No", "140.0"],
 ["CACGTGACATGT", "left palm", "2009.0", "4.0", "14.0", "subject-1", "No", "168.0"],
 ["CATATCGCAGTT", "left palm", "2008.0", "10.0", "28.0", "subject-2", "Yes", "0.0"],
 ["CGTGCATTATCA", "left palm", "2009.0", "1.0", "20.0", "subject-2", "No", "84.0"],
 ["CTAACGCAGTCA", "left palm", "2009.0", "3.0", "17.0", "subject-2", "No", "140.0"],
[...]

PR #482, opened 2021-02-06
[DON'T MERGE THIS PLS] Include ancestor information in processed taxonomy feature metadata

### LINKED ISSUE #422: How to better display higher taxonomic unknown levels?
While discussing large trees and coloring by taxonomy, we realized that by default Q2 assigns "Unassigned" to sequences that didn't get an assignment, for example:
Screen Shot 2020-10-13 at 10 47 43 AM
and that Empress by default will fill missing levels with "Unspecified"  and that this could become problematic when looking at higher levels, like this:
Screen Shot 2020-10-13 at 10 47 00 AM
as it merges the assignments that were "Unassigned" in level 1, with those without a classification at that specific level.

The questions are:
1. What should we do about it? There were a couple of suggestions: (a) repeat the latests level available level in the missing ones, basically have "Unassigned" in all the levels or "k__Bacteria" if those are the highest level available; and/or (b) do 1 but also add the level we are looking in the label, from the example above: "Unassigned [L2]" in all the levels or "k__Bacteria [L2]"
2. Should we do this by default or control via a parameter?

### LINKED ISSUE #473: Respect ancestors in feature metadata coloring/propagation
I brought this up in https://github.com/biocore/empress/issues/471#issuecomment-749300846, but for the sake of everyone's sanity this should probably be its own issue. This isn't a very pressing issue (the only common case it should run into is stuff like `s__` for Level 7), but there are corner-cases where real levels of taxa can have the same name but different ancestors. In these cases, we should assign these values distinct colors and not treat them as "equal" when doing propagation.

> As a sidenote: this discussion brings up the mildly wonky point that, currently, EMPress treats each feature metadata field (including the various levels of taxonomy) as its own independent thing, ignoring other metadata fields. This means that, for example, if you color by `Level 7` (species) in a 16S dataset using the default QIIME color map, you'll probably see a lot of clades of the tree colored as red due to all of the tips in the clade sharing a species classification of `s__`, even if they're from different genera/families/etc:
>
> yike
> 
> Addressing this would definitely be possible, by for example representing the values in each `Level N` string as the full taxonomy to that point (e.g. setting `Level 7` to `k__Bacteria; p__Firmicutes; c__Somecoolclass; o__Ogeezimrunningoutoftaxonomynamesiknow; f__Isanyonereadingthis; g__Himom; s__` instead of just `s__`) -- in some ways this is similar to a point @antgonza raised a few weeks ago in #422.

fedarko:
Closes #473. And closes #422, I guess!

Rather than representing a species-level "classification" as something like `"s__"`, we now represent it in the feature metadata as something like `"k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales; f__Bacteroidaceae; g__Bacteroides; s__"`.

This has a few nice side effects, including:

1. Stuff like `s__`, `g__`, `Unspecified`, etc. no longer form blobs of highly-common values when coloring the tree by taxonomy at a given level. Previously, coloring by level 6 or 7 using the default `Classic QIIME Colors` color map would show a ton of ugly red blobs -- this was because the first value alphabetically was `g__` / `s__` / etc., which was getting assigned red. See, for example, the outer layer (Level 7 in the Moving Pictures dataset) in the following screenshots, and how the red blobs go away due to this change:

    __Taxonomy levels 1 through 7 (kingdom through species), moving pictures dataset, before:__
    lessfancy

    __Taxonomy levels 1 through 7 (kingdom through species), moving pictures dataset, after:__
    fancy

2. Propagation (when coloring the tree itself by taxonomy; see #473 for details) is now limited to values of a taxonomy field that share ancestry. So now we don't have the problem of e.g. a bunch of unrelated `s__` values being merged together in the feature metadata coloring for level 7.
    
    *It is still possible for propagation to be done on `s__` (or on `Unspecified`, or other weird taxonomy values)*, but only between values with the same ancestry, if that makes sense. So we no longer consider `k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales; f__Bacteroidaceae; g__Bacteroides; s__` to be equal to `k__Archaea; p__somephylum; c__idk; o__; f__; g__; s__`, although both are equal to themselves.

    __Coloring the tree by level 7 (species), moving pictures dataset, before:__
    before7

    __Coloring the tree by level 7 (species), moving pictures dataset, after:__
    after7

## Caveats
This does inflate the size of the feature metadata stored in an EMPress visualization, probably by a decent amount. I haven't done thorough benchmarking yet, but I'd imagine it is pretty substantial.

This also arguably makes part of the UI look inelegant (see for example the legend in the final screenshot above -- users will have to scroll through the legend to see things like species names).

We could _potentially_ solve at least the size problem by not changing the way we store taxonomy feature metadata, and only considering ancestor info in the JS code when assigning colors / doing propagation or whatever, but that seems to me like it might be a headache to implement. However, personally: I think the benefits of this PR outweigh the downsides, and that future changes like #337 and #371 (PR for the latter coming soon :tm:) will offset these problems nicely.

### DIFF
--- a/empress/taxonomy_utils.py
+++ b/empress/taxonomy_utils.py
@@ -125,6 +125,14 @@ def split_taxonomy(feature_metadata):
                 ),
                 TaxonomyWarning
             )
+        else:
+            # Solve #473 -- add ancestor taxonomy strings
+            # (producing stuff like "k__Bacteria; p__Firmicutes; c__Bacilli"
+            # for class-level taxonomy levels).
+            # This disambiguates things like "s__" or "Unspecified" in the UI,
+            # at the cost of increasing metadata size.
+            for t in range(1, len(tax_levels.columns)):
+                tax_levels[t] = tax_levels[t - 1] + "; " + tax_levels[t]
 
         # Our use of expand=True means that tax_levels will be a DataFrame with
         # the same index as feature_metadata but with one column for each
--- a/tests/python/test_taxonomy_utils.py
+++ b/tests/python/test_taxonomy_utils.py
@@ -62,12 +62,12 @@ class TestTaxonomyUtils(unittest.TestCase):
             split_fm.loc["f1"],
             pd.Series({
                 "Level 1": "k__Bacteria",
-                "Level 2": "p__Bacteroidetes",
-                "Level 3": "c__Bacteroidia",
-                "Level 4": "o__Bacteroidales",
-                "Level 5": "f__Bacteroidaceae",
-                "Level 6": "g__Bacteroides",
-                "Level 7": "s__",
+                "Level 2": "k__Bacteria; p__Bacteroidetes",
+                "Level 3": "k__Bacteria; p__Bacteroidetes; c__Bacteroidia",
+                "Level 4": "k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales",
+                "Level 5": "k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales; f__Bacteroidaceae",
+                "Level 6": "k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales; f__Bacteroidaceae; g__Bacteroides",
+                "Level 7": "k__Bacteria; p__Bacteroidetes; c__Bacteroidia; o__Bacteroidales; f__Bacteroidaceae; g__Bacteroides; s__",
                 "Confidence": 0.95
             }, name="f1")
         )
@@ -75,12 +75,12 @@ class TestTaxonomyUtils(unittest.TestCase):
             split_fm.loc["f2"],
             pd.Series({
                 "Level 1": "k__Bacteria",
-                "Level 2": "p__Proteobacteria",
-                "Level 3": "c__Gammaproteobacteria",
-                "Level 4": "o__Pasteurellales",
-                "Level 5": "f__Pasteurellaceae",
-                "Level 6": "g__",
-                "Level 7": "s__",
+                "Level 2": "k__Bacteria; p__Proteobacteria",
+                "Level 3": "k__Bacteria; p__Proteobacteria; c__Gammaproteobacteria",
+                "Level 4": "k__Bacteria; p__Proteobacteria; c__Gammaproteobacteria; o__Pasteurellales",
+                "Level 5": "k__Bacteria; p__Proteobacteria; c__Gammaproteobacteria; o__Pasteurellales; f__Pasteurellaceae",
+                "Level 6": "k__Bacteria; p__Proteobacteria; c__Gammaproteobacteria; o__Pasteurellales; f__Pasteurellaceae; g__",
+                "Level 7": "k__Bacteria; p__Proteobacteria; c__Gammaproteobacteria; o__Pasteurellales; f__Pasteurellaceae; g__; s__",
                 "Confidence": 0.8
             }, name="f2")
         )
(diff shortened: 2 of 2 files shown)


empress/taxonomy_utils.py
```diff
@@ -125,6 +125,14 @@ def split_taxonomy(feature_metadata):
...
                 TaxonomyWarning
             )
+        else:
+            # Solve #473 -- add ancestor taxonomy strings
+            # (producing stuff like "k__Bacteria; p__Firmicutes; c__Bacilli"
+            # for class-level taxonomy levels).
+            # This disambiguates things like "s__" or "Unspecified" in the UI,
+            # at the cost of increasing metadata size.
+            for t in range(1, len(tax_levels.columns)):
+                tax_levels[t] = tax_levels[t - 1] + "; " + tax_levels[t]
```
kwcantrell, line 135: I would imagine that this would use quite a bit of memory. We wouldn't necessarily have to store all taxonomic info at each level. Instead, when we fetch taxonomy info in empress we can iterate over the levels (i.e. if we fetch level 7 then for each node we fetch level 1-7). This would allow us to implement this feature without using extra memory.

Essentially, using this for loop within empress to retrieve level information.

kwcantrell (commented):
Thanks @fedarko! I like the idea of this pr however, I am concerned with how much memory it would use. We should either benchmark how much additional memory this would use or come up with a way to dynamically retrieve taxonomic information within empress.

empress/taxonomy_utils.py
ElDeveloper: > This does inflate the size of the feature metadata stored in an EMPress visualization, probably by a decent amount. I haven't done thorough benchmarking yet, but I'd imagine it is pretty substantial.

It would be best if we could avoid further increasing the size of what we write to disk. We are already hitting some issues with browsers loading large files. 😟 

> ... that seems to me like it might be a headache to implement ...

Would you mind sharing what you think are some of the headache-inducing things associated with the JS option? I really like @kwcantrell's suggestion of doing this "on-demand" rather than upfront.

fedarko: > It would be best if we could avoid further increasing the size of what we write to disk. We are already hitting some issues with browsers loading large files. :worried:

I definitely feel that!

I had initially thought, when I was working on this Friday, that modifying the JS code to do this might be too much work—I imagined this involving, like, modifying a bunch of separate things in the UI code that use feature metadata to detect if the metadata field being used is a taxonomy column, and if so fetch the ancestor data. But the point Kalen makes about just taking this for loop and moving it into the JS code makes sense, and coming at this after a few days I think this should be easier than I'd thought.

After doing some digging, I think the only place in the JS code we'd need to really modify would be [`Empress.getUniqueFeatureMetadataInfo()`](https://github.com/biocore/empress/blob/5d14e07d747783edb2a38fcca82e754b9956f97f/empress/support_files/js/empress.js#L2192) -- this function is called by the feature metadata coloring and by the feature metadata barplot code.

I think we can set things up so the Python code (that does the taxonomy splitting) will just mark each new taxonomy column somehow, and pass this information to the JS -- and then we can modify `getUniqueFeatureMetadataInfo()` to detect these columns and fetch the ancestor info if needed. This should avoid having to inflate the feature metadata size in the QZV.

...OK, I will give this a shot in a separate PR -- hopefully it should get us the best of both worlds :)

kwcantrell: @fedarko 9 I might be missing something but I don't see why we would need to modify the python code. Don't we already label taxonomy columns as "Level n"? So if a user selects "Level 7" then all we would have to do is convert the last char the the column name (I.e. 7) and use that to determine how many levels we would need to grab.

fedarko: That'd work for probably the vast majority of datasets, but there is a slim case that the user doesn't specify taxonomy information and passes in feature metadata that still contains `Level` columns—in this case trying to use "ancestor" level information would likely be undesired.

We [throw an error](https://github.com/biocore/empress/blob/5d14e07d747783edb2a38fcca82e754b9956f97f/empress/taxonomy_utils.py#L91-L101) if the user's feature metadata already has `Level` columns _and_ a  `Taxonomy` column, but not in the case where the metadata has `Level` columns without a `Taxonomy` column.

So in short I think modifying the Python code would be necessary to avoid a few corner-cases, or at the very least to make the changes to the JS code easier.

kwcantrell: I see. In that case, I say we just add a boolean variable on the python side which flags if a `Taxonomy` column was passed into empress send it to JS to determine whether or not columns labeled as `Level` can be considered taxonomy rather than trying to create a special marker for each taxonomy column.

fedarko:
Supplanted by #487.

PR #483, opened 2021-02-10
New color map

kwcantrell:
This PR adds a new custom diverging color map (green to light orange to purple) that is great for coloring tree branches as a lot of the other diverging color maps are either hard to see the separation of values and/or the middle values are colored to light which makes the branches hard to see. 
log_ratio
You can see that this color map does a good job separating the two extreme values (i.e. endpoints) from the middle values while still making everything visible on screen.

### DIFF
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -246,9 +246,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             domain = [min, max];
         }
-        // var interpolator = chroma.scale(this.color).domain(domain);
-        // var interpolator = chroma.scale(["blue", "yellow", "red"]).domain(domain);
-        var interpolator = chroma.scale(["green", "#fed8b1", "purple"]).domain(domain);
+        var interpolator = chroma.scale(this.color).domain(domain);
+        // var interpolator = chroma.scale(["green", "#fed8b1", "purple"]).domain(domain);
         _.each(split.numeric, function (n) {
             scope.__valueToColor[n] = interpolator(parseFloat(n));
         });
@@ -609,6 +608,7 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         { id: "BrBG", name: "Brown-Blue-Green", type: Colorer.DIVERGING },
         { id: "PuOr", name: "Purple-Orange", type: Colorer.DIVERGING },
         { id: "PRGn", name: "Purple-Green", type: Colorer.DIVERGING },
+        { id: "PrYlGn": name: "Purple-Light Orange-Green", type: Colorer.DIVERGING}
     ];
     return Colorer;
 });
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -177,14 +177,15 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
 
         var interpolator;
         if (this.color === Colorer.__PR_OR_GR) {
-            interpolator = chroma.scale(Colorer.__prOrGr).domain([0, 1]);
+            interpolator = chroma.scale(Colorer.__prOrGr);
         } else {
-            interpolator = chroma.scale(this.color).domain([0, 1]);
+            interpolator = chroma.scale(this.color);
         }
 
         if (this.sortedUniqueValues.length === 1) {
             // If there's only 1 unique value, set its color as the first in
             // the color map. This matches the behavior of Emperor.
+            interpolator = interpolator.domain([0,1]);
             var onlyVal = this.sortedUniqueValues[0];
             if (this.reverse) {
                 this.__valueToColor[onlyVal] = interpolator(1);
@@ -564,8 +565,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         "#808000",
         "#008080",
     ];
-    Colorer.__PR_OR_GR = "PrOrGn";
-    Colorer.__prOrGr = ["green", "#fed8b1", "purple"];
+    Colorer.__PR_OR_GR = "GnOrPr";
+    Colorer.__prOrGr = ["#008000", "#fed8b1", "#800080"];
 
     // Used to create color select option and chroma.brewer
     //Modified from:
@@ -619,8 +620,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         { id: "PuOr", name: "Purple-Orange", type: Colorer.DIVERGING },
         { id: "PRGn", name: "Purple-Green", type: Colorer.DIVERGING },
         {
-            id: "PrOrGn",
-            name: "Purple-Light Orange-Green",
+            id: "GnOrPr",
+            name: "Green-LightOrange-Purple",
             type: Colorer.DIVERGING,
         },
     ];
--- a/empress/support_files/js/colorer.js
+++ b/empress/support_files/js/colorer.js
@@ -140,9 +140,7 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         if (this.color === Colorer.__QIIME_COLOR) {
             palette = Colorer.__qiimeDiscrete;
         } else {
-            // palette = chroma.brewer[this.color];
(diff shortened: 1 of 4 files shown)


empress/support_files/js/colorer.js
```diff
@@ -605,6 +618,11 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         { id: "BrBG", name: "Brown-Blue-Green", type: Colorer.DIVERGING },
         { id: "PuOr", name: "Purple-Orange", type: Colorer.DIVERGING },
         { id: "PRGn", name: "Purple-Green", type: Colorer.DIVERGING },
+        {
+            id: "PrOrGn",
+            name: "Purple-Light Orange-Green",
```
fedarko, line 623: I think we may want to reverse the name of this (calling it `Green-Light Orange-Purple` instead of the other way around). The rationale is that the other Diverging color maps' names go in ascending order (see e.g. `Red-Blue` and `Pink-Yellow-Green` below), whereas the "low" color for this map is purple instead of green.

image

image

image

fedarko, line 623: Also, this is much much much less important, but it might make sense to remove the space in between `Light` and `Orange` (so it's just named `Purple-LightOrange-Green` or something) -- currently, none of the other color map names that have hyphens also have a space, so to me the UI looked at first like something went wrong and this was two color maps' names merged together (`Purple-Light` and `Orange-Green` combined into one). But feel free to disagree with this one, I can see how the space helps readability also

Screenshot_2021-02-10_01-13-11

empress/support_files/js/colorer.js
```diff
@@ -553,6 +564,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         "#808000",
         "#008080",
     ];
+    Colorer.__PR_OR_GR = "PrOrGn";
+    Colorer.__prOrGr = ["green", "#fed8b1", "purple"];
```
fedarko, line 568:
```suggestion
    Colorer.__prOrGr = ["#008000", "#fed8b1", "#800080"];
```
It's probably irrational on my part, but using hex codes just seems somehow "safer" to me -- in case Chroma.js would decide randomly that the default `green` should be another color, or whatever. (I doubt that'd ever happen, but IDK.)

Another motivation for using hex codes instead of color names is that it is consistent with how `Classic QIIME Colors` is declared above -- this could make this color map more portable to other projects that don't use Chroma.js.

empress/support_files/js/colorer.js
```diff
@@ -175,15 +175,21 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
             );
         }
 
+        var interpolator;
+        if (this.color === Colorer.__PR_OR_GR) {
+            interpolator = chroma.scale(Colorer.__prOrGr).domain([0, 1]);
+        } else {
+            interpolator = chroma.scale(this.color).domain([0, 1]);
+        }
```
fedarko, line 183:
```suggestion
        if (this.color === Colorer.__PR_OR_GR) {
            interpolator = chroma.scale(Colorer.__prOrGr);
        } else {
            interpolator = chroma.scale(this.color);
        }
```

fedarko: (Rationale for this: we don't need to set the domain to [0, 1] unless there is just one unique value, so this avoids doing this extra work if we can.)

empress/support_files/js/colorer.js
```diff
@@ -175,15 +175,21 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
...
+
         if (this.sortedUniqueValues.length === 1) {
             // If there's only 1 unique value, set its color as the first in
```
fedarko, line 186:
```suggestion
            interpolator = interpolator.domain([0, 1]);
            // If there's only 1 unique value, set its color as the first in
```

Technically I don't even think we need to add this -- the default domain of a scale in chroma is [0, 1] (see [here](https://gka.github.io/chroma.js/#chroma-scale)). However, I do like explicitly setting it as is currently done in this PR, both in case chroma'd change this in the future and to make the code more understandable.

fedarko (changes requested):
This looks dope! A few small finicky changes I think would improve clarity, but this should be good to go without much tweaking. Thanks for adding this!

empress/support_files/js/colorer.js
kwcantrell: That makes sense. I also felt it looked a bit odd with the space but I wasn't sure what the best way to write it. I'll remove the space since the space looks a bit odd in this context.

empress/support_files/js/colorer.js
kwcantrell: I'll add the hex codes. I doubt chroma.js would change the default colors but agree that we shouldn't rely on chroma.js to maintain its default colors.

empress/support_files/js/colorer.js
kwcantrell: @fedarko thanks for reviewing this! I think I have addressed all of your comments.

kwcantrell:
@fedarko thanks for reviewing this! I think I have addressed all of your comments.

empress/support_files/js/colorer.js
```diff
@@ -553,6 +565,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         "#808000",
         "#008080",
     ];
+    Colorer.__PR_OR_GR = "GnOrPr";
+    Colorer.__prOrGr = ["#008000", "#fed8b1", "#800080"];
```
fedarko, line 569: Would it be possible to change these to reflect the new name? (This'd also involve updating the other uses of these variables throughout this module.)

This is not very high priority, since at this point it's just to avoid confusing future people looking at this code who wonder why the names are switched.

```suggestion
    Colorer.__GN_OR_PR = "GnOrPr";
    Colorer.__gnOrPr = ["#008000", "#fed8b1", "#800080"];
```

fedarko (commented):
Thanks @kwcantrell! This looks great -- I have one suggestion, but we could probably merge this in even without it.

empress/support_files/js/colorer.js
kwcantrell: good catch! I'll change them.

empress/support_files/js/colorer.js
```diff
@@ -553,6 +565,8 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         "#808000",
         "#008080",
     ];
+    Colorer.__GR_OR_PR = "GnOrPr";
```
fedarko, line 568:
```suggestion
    Colorer.__GN_OR_PR = "GnOrPr";
```

Just to make this consistent with `Colorer.__gnOrPr` -- if we use different abbreviations i suspect it'll cause problems down the line

empress/support_files/js/colorer.js
```diff
@@ -244,7 +251,12 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
         } else {
             domain = [min, max];
         }
-        var interpolator = chroma.scale(this.color).domain(domain);
+        var interpolator;
+        if (this.color === Colorer.__GR_OR_PR) {
```
fedarko, line 255:
```suggestion
        if (this.color === Colorer.__GN_OR_PR) {
```

empress/support_files/js/colorer.js
```diff
@@ -175,15 +175,22 @@ define(["chroma", "underscore", "util"], function (chroma, _, util) {
             );
         }
 
+        var interpolator;
+        if (this.color === Colorer.__GR_OR_PR) {
```
fedarko, line 179:
```suggestion
        if (this.color === Colorer.__GN_OR_PR) {
```

fedarko (commented):
Sorry to nitpick but I have one more suggestion (I don't think it's needed but it will probably cause confusion if left unfixed). Aside from that, this looks good!

kwcantrell:
> Sorry to nitpick but I have one more suggestion (I don't think it's needed but it will probably cause confusion if left unfixed). Aside from that, this looks good!

Nitpick away :)

It should be fixed now.

fedarko:
Perfect, thanks so much!

PR #484, opened 2021-02-10
Gradient Coloring Option for Feature Metadata

kwcantrell:
~~Address #405 This PR adds a gradient color option for feature metadata (similar to gradient option for barplots). There is one small issue I wasn't sure how to resolve though which you can see in the image below (look at the legend). 
Screenshot from 2021-02-09 21-38-50
Basically, the gradient color bar extends beyond the initial size of the legend which triggers the scroll bar. I would like to have both the legend title and gradient color bar visible without triggering the scroll bar but there is some odd css that I can't figure out.  @fedarko would you have any idea?~~

I will address the above issues in another PR. I want to get the basics of this PR merged in with master so that it would be possible to recreate the figure for the stacked faiths pd manuscript.  This PR in its current form allows users to use a gradient color map on feature metadata. It will also display an error message to the user if they try to use gradient coloring on non-numeric data/only one unique number  (similar to how barplots handle it). There is still an issue where gradient color triggers a scroll bar in the legend. But I will address that in a follow up PR (unless @fedarko or @ElDeveloper think it should be resolved here).


fedarko:
> Basically, the gradient color bar extends beyond the initial size of the legend which triggers the scroll bar. I would like to have both the legend title and gradient color bar visible without triggering the scroll bar but there is some odd css that I can't figure out.

I think I found out a way to (sort of) fix this problem, but I'm not yet sure why it works. Could you try removing `overflow: auto;` from the `#legend-main` CSS? It looks like this fixes the problem for me in Firefox (without this fix, I get the same problem in Firefox).

I am really not sure why the overflow thing breaks it in Firefox, since this same setting (overflow auto) is also used for the barplot layer legends, and those look fine in Firefox. I thought there was something weird going on with the [`resize` CSS setting](https://developer.mozilla.org/en-US/docs/Web/CSS/resize) that was used for `#legend-main`, but removing `resize` entirely (and just keeping the `overflow: auto` thing) doesn't fix the problem -- only removing the overflow auto thing seems to fix it.

My best guess right now is that something about the overflow setting (in combination with the other CSS in effect on the main legend element) results in Firefox thinking that a scrollbar should be present, which maybe somehow messes with the SVG / element dimensions.

Strangely, this fix has the bizarre side effect (...just in Chrome, as far as I can tell) of making the default width of a continuous legend too wide. Although I guess this is a better problem to have :)

whoops

Notably-- removing overflow: auto will **only** be useful for the main legend when a continuous scale is being used. This is because we rely on overflow (and the ability to scroll stuff) when displaying categorical legends with a bunch of values -- without overflow these'll go off the legend. If we wind up going with this as a solution, I think we'd need to add/remove the `overflow: auto` CSS thing when changing the legend type somehow.

empress/support_files/js/side-panel-handler.js
```diff
@@ -333,11 +337,15 @@ define(["underscore", "Colorer", "util"], function (_, Colorer, util) {
         var col = this.fColor.value;
         var coloringMethod = this.fMethodChk.checked ? "tip" : "all";
         var reverse = this.fReverseColor.checked;
+        var continuous =
+            !this.fContinuousDiv.classList.contains("hidden") &&
+            this.fContinuousChk.checked;
         this.empress.colorByFeatureMetadata(
```
fedarko, line 343: It's possible to try to use continuous coloring when a non-continuous field (eg a taxonomy column) is selected. Currently this results in nothing happening, which will probably confuse the user. We should detect the error thrown and alert the user, similarly to how continuous coloring is handled in the barplot code:

https://github.com/biocore/empress/blob/5d14e07d747783edb2a38fcca82e754b9956f97f/empress/support_files/js/empress.js#L1839-L1865

empress/support_files/templates/side-panel.html
```diff
@@ -99,6 +99,13 @@
         If checked, collapse all clades whose members share the same feature
         metadata value.
       </p>
+      <div id="feature-continous-div" class="hidden">
+        <p>
+            <label for="feature-continuous-color-chk">Continuous values?</label>
+            <input id="feature-continuous-color-chk" type="checkbox" value="false"
+                   class="empress-input">
+        </p>
+      </div>
```
fedarko, line 108: I'm not sure right now what the best way to handle this would be, but the continuous values option should be mutually exclusive with the clade collapsing option, right? I don't think it will ever make sense to combine the two (...i might be wrong though :)

The fact that "Collapse Clades" works automatically when it's checked -- unlike the other feature metadata coloring options -- might complicate the UI, and make solving this problem harder than it should be.

It might be easiest to make checking the "Collapse Clades" option hide the continuous values checkbox, and vice versa. Then, we could change the collapse clades thing to only apply when the user clicks "Update", which would prevent weird state problems in the UI.

empress/support_files/js/empress.js
```diff
@@ -2275,14 +2276,19 @@ define([
         var colorer = new Colorer(
             color,
             sortedUniqueValues,
-            undefined,
-            undefined,
+            continuous,
+            continuous ? -1 : undefined,
```
fedarko, line 2280: Using -1 as the gradient ID suffix here is a super clever way to handle this! Could you add a brief comment (just a line or so) above this explaining the motivation for this? Just so this will be easier for future devs to understand :)

empress/support_files/templates/side-panel.html
fedarko: (Maybe we should create a new issue titled "Make clade collapsing only apply when `Update` is clicked" or something, address that first, and then come back to this?)

kwcantrell: Clade collapsing shouldn't update automatically when its clicked. I'll create an issue for this.

kwcantrell:
I simplified this PR a bit. Please look at updated PR description.

empress/support_files/js/side-panel-handler.js
kwcantrell: This should now be similar to how continuous coloring is handled in the barplot code.

empress/support_files/templates/side-panel.html
kwcantrell: (At least for now) I think clade collapsing should be allowed since it will only collapse clades that have the same value. In the case of continuous data it might actually be interesting if you see a large collapsed clade. I think if anything, we should modify clade collapsing with continuous data and collapse clades if they are within some +-lambda of each other. We can revisit this issue later.

empress/support_files/js/empress.js
kwcantrell: Description has been added. I tried to use the same language that is found in Colorer.js.

empress/support_files/js/empress.js
```diff
@@ -2257,7 +2257,8 @@ define([
         cat,
         color,
         method,
-        reverse = false
+        reverse = false,
+        continuous = false
```
fedarko, line 2261: Could you mention the `continuous` parameter in the function documentation, after `reverse`?

empress/support_files/js/empress.js
```diff
@@ -2271,18 +2272,47 @@ define([
...
+            // If the Colorer construction failed (should only have
+            // happened if the user asked for continuous values but the
+            // selected field doesn't have at least 2 unique numeric
+            // values), then we open a toast message about this error and
+            // then raise it again to send error to console.
+            var msg =
+                "Error with assigning colors: '" +
+                cat +
+                "' is either non-numeric" +
+                " or has less than 2 unique numeric values";
```
fedarko, line 2303:
```suggestion
                " has less than 2 unique numeric values.";
```

I don't think there is a big distinction between these cases -- all non-numeric columns have less than 2 unique numeric values, by definition. This matches the error used for continuous barplots.

empress/support_files/js/empress.js
```diff
@@ -2271,18 +2272,47 @@ define([
...
+                sortedUniqueValues,
+                continuous,
+                // Colorer will create a special gradient ID using the number
+                // we pass into this parameter. This allows empress to display
+                // multiple gradients at the same time without them overriding
+                // each other. Currently, the barplots are setup to start at
+                // 0. So we set this to -1 this value will never be -1 in
+                // the barplots. Thus, by setting this value to -1 we can
+                // display the feature metadata gradient along side the
+                // barplot gradients.
```
fedarko, line 2289:
```suggestion
                // each other. Currently, the barplots are set up to start at
                // 0. So, we set this value to -1 here to avoid conflict with
                // the barplot gradients; this allows us to display the
                // feature metadata gradient alongside the
                // barplot gradients.
```

empress/support_files/js/empress.js
```diff
@@ -2324,7 +2357,11 @@ define([
                         to a set of keys (i.e. tree nodes) that are unique to
                         each group.
      */
-    Empress.prototype._projectObservations = function (obs, ignoreAbsentTips) {
+    Empress.prototype._projectObservations = function (
+        obs,
+        ignoreAbsentTips,
+        isContinuous = false
```
fedarko, line 2363: Since it looks like this isn't used in this PR yet, could we remove this parameter?

empress/support_files/js/empress.js
```diff
@@ -2271,18 +2272,47 @@ define([
...
+            // selected field doesn't have at least 2 unique numeric
+            // values), then we open a toast message about this error and
+            // then raise it again to send error to console.
+            var msg =
+                "Error with assigning colors: '" +
+                cat +
+                "' is either non-numeric" +
+                " or has less than 2 unique numeric values";
+            util.toastMsg(msg, 5000);
+            throw msg;
```
fedarko, line 2305: There is a small bug I noticed that happens when this error is thrown after the tree is already colored.

What happens is `resetTree()` is called before we determine whether or not to throw this error (I think [this is the problematic line](https://github.com/biocore/empress/blob/f732852ce72633b7c241f5dc85021cf3f8a19293/empress/support_files/js/side-panel-handler.js#L294) -- `colorByFeatureMetadata` is called on line 300), so if this error gets thrown then it removes the tree coloring when the user next clicks on the tree. See a GIF of this below.

dang

This wasn't a problem before because to my knowledge nothing in sample / feature metadata coloring would have the potential to throw an error. However, it makes sense we'd have to deal with this eventually.

I think the way to fix this is moving the call to `resetTree()` to after `colorByFeatureMetadata()` is able to successfully create a `Colorer` object here (we'd need to also do this for `colorBySampleCat()`). We probably also want to move the `        updateBtn.classList.add("hidden");` to after calling `colorMethodName`, so that if something goes wrong while trying to color the tree the `Update` button is still available.

fedarko (changes requested):
Thanks @kwcantrell! This is looking good.

empress/support_files/js/empress.js
```diff
@@ -2271,18 +2272,47 @@ define([
...
+        } catch (err) {
+            // If the Colorer construction failed (should only have
+            // happened if the user asked for continuous values but the
+            // selected field doesn't have at least 2 unique numeric
+            // values), then we open a toast message about this error and
+            // then raise it again to send error to console.
+            var msg =
+                "Error with assigning colors: '" +
+                cat +
+                "' is either non-numeric" +
```
fedarko, line 2302:
```suggestion
                'Error with assigning colors: ' +
                'the feature metadata field "' +
                cat +
                '" is either non-numeric' +
```

empress/support_files/js/empress.js
fedarko, line 2303:
```suggestion
            var msg =
                'Error with assigning colors: ' +
                'the feature metadata field "' +
                cat +
                '" has less than 2 unique numeric values.';
```

empress/support_files/js/empress.js
kwcantrell: added

empress/support_files/js/empress.js
kwcantrell: removed

empress/support_files/js/empress.js
kwcantrell: Good catch! I added a try catch block in `_updateColoring` method in side-panel-handler.js that will reset the tree/legend/update-button if an error is thrown during the `colorBy...` methods

ElDeveloper:
@kwcantrell looks like the js style checker failed. Just flagging for your visibility.

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F484%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F484%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F484%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F484%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F484%2Fplain.qzv)

empress/support_files/js/empress.js
kwcantrell: Instead of modifying where resetTree() is called, a error message is shown and discrete coloring is used (also the `Continous values?` checkbox is unselected).

kwcantrell:
Thanks @fedarko and @ElDeveloper! I think I have addressed all of your comments so this pr should be good to merge.

empress/support_files/js/side-panel-handler.js
```diff
@@ -291,21 +295,28 @@ define(["underscore", "Colorer", "util"], function (_, Colorer, util) {
         lwInput,
         updateBtn
     ) {
-        this.empress.resetTree();
+        try {
```
ElDeveloper, line 298: Is there any way we can limit the lines executed in the try block to just the line that might fail or better yet, check upfront?

ElDeveloper (approved):
Just one comment/question. Thanks so much @kwcantrell

empress/support_files/js/side-panel-handler.js
kwcantrell: thanks, I meant to remove the try-catch

ElDeveloper (approved):
Thanks @kwcantrell!

@fedarko feel free to merge after you review.

empress/support_files/js/empress.js
```diff
@@ -2479,8 +2521,11 @@ define([
 
         // color tree
         this._colorTree(obs, cm);
-
-        this.updateLegendCategorical(cat, keyInfo);
+        if (continuous) {
+            this._legend.addContinuousKey(cat, keyInfo);
+        } else {
```
fedarko, line 2526: A less important problem, but one that may annoy users. Currently, resizing the main legend CSS when a gradient is shown breaks things:

arghcss

This isn't a problem for the barplot legend gradients since those aren't resizable, unlike the main legend.

There are probably multiple good ways of handling this (I just spent some time looking into this). For example, we could prevent resizing the legend CSS, but that will be inconvenient when the default size of the legend is big (e.g. when coloring by Level 7 or something).

For now, one easy solution might be setting [the height of the legend SVG here](https://github.com/biocore/empress/blob/5b8a340bbff046f9d1d7f992248db87224b13c7a/empress/support_files/js/legend.js#L182) to a constant height, e.g. `15em` (matching the `barplot-layer-legend` class), instead of `100%`. This should only be done for the main legend and not for the barplot layer legends -- doing it for those will mess up the formatting and make the default size of the SVGs too large -- so we would need to add a conditional to the Legend JS that checks if this is for the main legend or a barplot layer (this would overlap with the discussion above about making the Legend class aware of if it's used for the main legend or for barplots).

This isn't _urgent_ (I'd be ok merging this PR in without addressing this), so if you'd like to delay it to later that's ok.

fedarko (changes requested):
Thanks @kwcantrell! This looks great. Most of the comments/suggestions I've made are minor, but there is one important one we should address before merging (in the context of the warning for non-numeric values).

empress/support_files/js/empress.js
```diff
@@ -2479,8 +2521,11 @@ define([
 
         // color tree
         this._colorTree(obs, cm);
-
-        this.updateLegendCategorical(cat, keyInfo);
+        if (continuous) {
+            this._legend.addContinuousKey(cat, keyInfo);
```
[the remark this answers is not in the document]
kwcantrell: Rather than creating special flags or other things that will increase the complexity of the code, I create a generic message that is used in both barplots and feature metatadata

empress/support_files/js/empress.js
kwcantrell: I vote to hold this off. I'd like to get this PR merged in ASAP so we can do a new release.

fedarko: Sounds good - just opened #515.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: I think I disagree with this, sorry -- IMO it is important to let the user know that non-numeric values will cause no bars to show up in a barplot, but this is a very different effect from what happens when non-numeric values occur during tree coloring (in these cases the corresponding tip is still colored, it just isn't colored differently).

I guess we could defer this to later if you'd prefer to get this merged in ASAP -- the current error message is preferable to a generic one, IMO, since it is at least clear about the non-numeric barplot case.

fedarko:
@kwcantrell I just went over the changes. It looks like the only two TODOs left are the non-numeric warning and testing `continousFailedFunc`, and both of those could be deferred to separate issues if you'd like. Let me know if you'd like me to just merge this as is, and I can do that / open issues as needed.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
kwcantrell: I created a new message for the feature metadata coloring.
Screenshot from 2021-06-11 01-26-47

empress/support_files/js/empress.js
```diff
@@ -2464,18 +2483,57 @@ define([
...
+            // values), then we open a toast message about this error and
+            // use discrete coloring instead.
+            continuous = false;
+            var msg =
+                'Error with assigning colors: the feature metadata field "' +
+                cat +
+                '" has less than 2 unique numeric values, so it cannot be ' +
+                "used for continuous coloring. " +
+                "Using discrete coloring instead.";
+            util.toastMsg(msg, 5000);
```
fedarko, line 2517:
```suggestion
            util.toastMsg("Feature metadata coloring error", msg, 5000);
```

Since the #519 PR is in now :)

empress/support_files/js/empress.js
```diff
@@ -2487,8 +2545,11 @@ define([
 
         // color tree
         this._colorTree(obs, cm);
-
-        this.updateLegendCategorical(cat, keyInfo);
+        if (continuous) {
+            this._legend.addContinuousKey(cat, keyInfo);
```
fedarko, line 2549:
```suggestion
            this.resizeLegend();
            this._legend.addContinuousKey(cat, keyInfo);
```

(now that #512 is merged in)

empress/support_files/js/empress.js
```diff
@@ -158,12 +158,26 @@ define([
...
          * @type {Legend}
          * Legend describing the way the tree is colored.
          * @private
          */
         this._legend = new Legend(document.getElementById("legend-main"));
+        this._legend.setMissingNonNumericWarning(
+            "Some value(s) in this field were missing and/or not numeric. " +
+                "These value(s) are not included in the gradient, and the " +
+                "associated nodes have been left as the default color."
+        );
```
fedarko, line 180:
```suggestion
            "Some value(s) in this field were missing and/or not numeric. " +
                "These value(s) are not included in the gradient, and the " +
                "associated nodes have been left as the default color.",
            "Missing / non-numeric value(s)' associated nodes left as " +
                "default color."
        );
```

Currently, no short message is set for the main legend, which messes with the legend SVG export:

legs2
_top = main legend; bottom = barplot legend (both using modified data with some missing/non-numeric values)_

This fixes that.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
fedarko: Looks good, thanks a ton!

tests/test-empress.js
```diff
@@ -353,6 +353,122 @@ require([
...
+                "midValStr",
+                "minValStr",
+                "missingNonNumerics",
+                "pageSVG",
+            ];
+
+            // make sure color gradient returns correct keyInfo
+            var resultProperties = util.naturalSort(Object.keys(cm));
+            deepEqual(resultProperties, properties);
+
```
fedarko, line 378:
```suggestion

            deepEqual(cm.gradientID, "Gradient-1");
            deepEqual(cm.maxValStr, "2");
            deepEqual(cm.midValStr, "1.5");
            deepEqual(cm.minValStr, "1");
            deepEqual(cm.missingNonNumerics, false);
```

(While we're at it, may as well test a few easy things.)

empress/support_files/js/legend.js
```diff
@@ -742,6 +744,28 @@ define(["jquery", "underscore", "util"], function ($, _, util) {
         };
     };
 
+    Legend.prototype.setMissingNonNumericWarning = function (
```
fedarko, line 747: This works -- and I'm OK with merging it in as is, as long as the other changes I've suggested here are addressed -- but I think we could probably structure this code more clearly. To me, it is confusing that the `Legend` defaults to the barplot messages, and that these messages have to be passed in from `Empress` object (but the barplot messages are already located here).

I propose we restructure this so that the `Legend` class (on creation) takes as input an optional `legendUse` parameter of `tree` or `barplot`, which then adjusts what the full and short missing non-numeric warnings are (so all of the warning text will live in `legend.js`). That said: I'm totally OK with deferring this to a future issue so we can get this PR in soon.

fedarko (changes requested):
I have a few suggestions for fixes. I think once all of these are in we can merge this in. Thanks so much @kwcantrell!

fedarko:
One more note sorry --

> I fixed the bug were the gradient bar would overflow the legend. I changed containerSVG.setAttribute("height", "100%"); to containerSVG.setAttribute("height", "80%"); in addContinousKey in legend.js

This works, thank you! However, the bug still seems to be a problem (albeit much less noticeable) when the warning message is shown:

Peek 2021-06-11 19-33

It's def not something we gotta fix right now (the `80%` solution should work great for most use cases!), but I just wanted to bring this up so that we know where to start when eventually working on #515.

empress/support_files/js/empress.js
kwcantrell: 👍

empress/support_files/js/legend.js
kwcantrell: I agree, this was a fairly lazy solution. The goal was to modify the existing code as little as possible hence why it defaults to the barplot messages.

kwcantrell: Instead of adding a special parameter, I will make Legend an abstract class and add SampleFeatureColoringLegend and BarplotLegend that will inherit it. This will also make it a lot easier to change the behavior of the different legend types.

kwcantrell:
I fixed the legend gradient bar issue by placing the gradient svg within a div container. It now will be initialized as 80% of the legend size but will not be resized if the user resizes the legend.

I also abstracted the legend class and added SampleFeatureColorLegend and BarplotLegend.

fedarko (changes requested):
Changes look good -- the suggestions here are only minor things. The warning messages / legend style fixes are dope.

However, I found two bugs with this PR that we need to fix before merging. Will comment with that.

fedarko:
Looks good.

Did some more testing of this. Found two bugs we need to fix: 

### 1. Turning feature metadata coloring off and on again

Although the `Color Map` is reset when turning off feature metadata coloring, the `Continuous values?` checkbox isn't. This makes it possible for the user to try to trigger continuous coloring using the default categorical scale, which throws an error:

```
colorer.js:366 Uncaught Error: No gradient data defined for this Colorer; check that useQuantScale is true and that the selected color map is not discrete.
    at Colorer.getGradientInfo (colorer.js:366)
    at Empress.colorByFeatureMetadata (empress.js:2521)
    at SidePanel._colorFeatureTree (side-panel-handler.js:345)
    at SidePanel._updateColoring (side-panel-handler.js:304)
    at HTMLButtonElement.fUpdateBtn.onclick (side-panel-handler.js:641)
```

GIF of this happening:

gc

There are probs multiple ways to fix this, but I think one of the easiest might be resetting `Continuous values?` to unchecked and hiding it whenever Feature Metadata Coloring is turned off.

### 2. Nodes left as the "default color" in the gradient are not assigned a proper color, and are left out of the tree export

This is a weird one. I was playing around with the tree exporting, and noticed that the tree nodes corresponding to non-numeric values seemed to be invisible in Chromium's SVG viewer:

image

In Inkscape, however, these same nodes seemed visible (same region, same SVG file):

image

It turns out that these nodes are getting assigned an invalid color, and I guess different SVG renderers will interpret this invalid color arbitrarily. I'm copying in one of the lines from the SVG file:

```svg
<line x1="1923.2021484375" y1="564.703125" x2="1928.596923828125" y2="566.2871704101562" stroke="rgb(NaN,NaN,NaN)"  />
```

Furthermore, I think this problem isn't just with the tree exporting -- these nodes actually **aren't getting assigned colors**, and we've been lucky in that this lack of color somehow gets turned into `rgb(0, 0, 0)` (which *differs* from the actual default color of `rgb(64, 64, 64)`). I don't know if it's easy to tell, but notice how a node with missing data actually subtly changes color when enabling continuous feature metadata coloring in the GIF below, even though it should remain the default color: this is because it's getting changed from the default dark gray color to black (`rgb(0, 0, 0)`).

subtle

I believe we need to fix this problem by adding another parameter to the `Colorer` class constructor, that indicates whether or not to include color info for missing values in the `getMapRGB` / `getMapHex` functions (and if so, what that default color should be). We could probably handle this in a single optional parameter named `nonNumericColor` or something (which should default to `null` or `undefined`). If this parameter is specified (could be a hex string or an RGB integer, whatever's easiest -- when supplied to `Colorer` by `Empress`, it should be directly derived from `Empress.DEFAULT_COLOR`) and gradient coloring is used, then all non-numeric values should be added (with this color) in `scope.__valueToColor` here:

https://github.com/biocore/empress/blob/df3a4da5a5b23d465937112dfa2050f8957bb0ed/empress/support_files/js/colorer.js#L260-L265

The code should be extendable with something like:

```js
if (this._missingNonNumerics) {
    // (this next line will depend on if nonNumericColor is a hex
    // string, RGB tuple, RGB integer, etc.)
    var nonNumericColorChroma = chroma(nonNumericColor);
    _.each(split.nonNumeric, function (nn) {
        scope.__valueToColor[nn] = nonNumericColorChroma;
    });
}
```

We should test this, also (I think extending the current `test-colorer` stuff to handle this extra parameter should be feasible).

I guess the bright side is that having this parameter in here will make it easier in the future to let users change the color of missing / non-numeric barplots -- so there's that.

ElDeveloper:
@kwcantrell can you resolve the conflicts?

kwcantrell:
> Furthermore, I think this problem isn't just with the tree exporting -- these nodes actually aren't getting assigned colors, and we've been lucky in that this lack of color somehow gets turned into rgb(0, 0, 0) (which differs from the actual default color of rgb(64, 64, 64)).

Good catch! I like the idea of adding a parameter to let users choose what color to assign non-numeric values in the case of continuous coloring. However, that would be a bit out of scope for this pr. So to fix the 2nd bug you found, I choose to simply remove non-numeric values from `obs` in `empress.colorByFeateureMetadata`. You solution does work however, it would lead to empress thickening the non-numeric lines (which I believe should not happen). But I think we should create an issue for choosing the non-numeric and link your comment since ultimately it would be a better solution.

ElDeveloper:
@fedarko would you be able to look at @kwcantrell's answer to your comment? What else do we need to get this branch merged?

ElDeveloper:
@kwcantrell, I was able to confirm that the issues that @fedarko pointed out were no longer there (missing data branches are assigned a 64,64,64 color value during export, and the exception is no longer raised. However, I did find a minor glitch related to the continuous values checkbox. It seems like the box is leftover after the menu is reset when the feature-metadata coloring is turned off. Once you click that while a non-continuous colorscheme is selected the checkbox disappears. After the checkbox disappears everything continues to work well. See this animated GIF:

pr

------

Just in case, this is the code I ran to generate an additional 2 columns with continuous metadata:

```python
import qiime2 as q2, pandas as pd, numpy as np
df = q2.Artifact.load('docs/moving-pictures/taxonomy.qza').view(pd.DataFrame)

# continuous category
df['seven_little_numbers'] = np.random.randint(0, 7, len(df))

# continuous category with missing values
df['six_little_numbers'] = df['seven_little_numbers'].copy()
df.loc[df.six_little_numbers == 0, 'six_little_numbers'] = 'I am not a number, I am a free man'

df.to_csv('docs/moving-pictures/feature-metadata.tsv', sep='\t')
```

```bash
qiime empress community-plot \
	--i-tree docs/moving-pictures/rooted-tree.qza \
	--i-feature-table docs/moving-pictures/table.qza \
	--m-feature-metadata-file docs/moving-pictures/feature-metadata.tsv \
	--o-visualization docs/moving-pictures/empress-tree.qzv
```

PR #485, opened 2021-02-12
Fixes error when no feature metadata file is provided

ahdilmore:
When I previously ran empress without supplying a feature metadata file, I received the error "All objects passed were None." This occurs on line 167 when pandas attempts to concatenate two Nones. I added a line to check whether both self.tip_md and self.int_md are both None to avoid raising the error.

### DIFF
--- a/empress/core.py
+++ b/empress/core.py
@@ -164,7 +164,10 @@ class Empress():
 
                 # if there are no matches set to None so Emperor can ignore
                 # the feature metadata
-                feature_metadata = pd.concat([self.tip_md, self.int_md])
+                if (self.tip_md is None) and (self.int_md is None): 
+                    feature_metadata = pd.DataFrame()
+                else: 
+                    feature_metadata = pd.concat([self.tip_md, self.int_md])
                 arrows = self.ordination.features.index
                 if (feature_metadata.index.intersection(arrows).empty or
                    feature_metadata.empty):
--- a/empress/core.py
+++ b/empress/core.py
@@ -164,9 +164,9 @@ class Empress():
 
                 # if there are no matches set to None so Emperor can ignore
                 # the feature metadata
-                if self.tip_md is None and self.int_md is None: 
+                if self.tip_md is None and self.int_md is None:
                     feature_metadata = pd.DataFrame()
-                else: 
+                else:
                     feature_metadata = pd.concat([self.tip_md, self.int_md])
                 arrows = self.ordination.features.index
                 if (feature_metadata.index.intersection(arrows).empty or
--- a/empress/core.py
+++ b/empress/core.py
@@ -164,7 +164,7 @@ class Empress():
 
                 # if there are no matches set to None so Emperor can ignore
                 # the feature metadata
-                if (self.tip_md is None) and (self.int_md is None): 
+                if self.tip_md is None and self.int_md is None: 
                     feature_metadata = pd.DataFrame()
                 else: 
                     feature_metadata = pd.concat([self.tip_md, self.int_md])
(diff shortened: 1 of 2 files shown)


empress/core.py
```diff
@@ -164,7 +164,10 @@ def __init__(self, tree, table=None, sample_metadata=None,
 
                 # if there are no matches set to None so Emperor can ignore
                 # the feature metadata
-                feature_metadata = pd.concat([self.tip_md, self.int_md])
+                if (self.tip_md is None) and (self.int_md is None):
```
ElDeveloper, line 167: Parenthesis are not needed here, and the PEP-8 linter will error:

```suggestion
                if self.tip_md is None and self.int_md is None: 
```

ElDeveloper (changes requested):
Thanks @ahdilmore! Would you be able to add a new test-case that without the changes reproduces the error, and with the changes in place passes? I left one suggestion which should take care of the failing build.

ElDeveloper (approved):
Thanks so much @ahdilmore

tests/python/test_core.py
```diff
@@ -243,6 +243,14 @@ def test_init_with_ordination(self):
         # emperor is instantiated as needed but not yet setup
         self.assertTrue(isinstance(viz._emperor, Emperor))
 
+    def test_init_with_ordination_features(self):
+        viz = Empress(self.tree, self.table, self.sample_metadata,
+                      ordination=self.pcoa, shear_to_table=False)
```
fedarko, line 248:
```suggestion
                      ordination=self.biplot, shear_to_table=False)
```
From some manual testing, it looks like the bug you discovered (thank you!) only appears when a biplot is passed as the ordination -- a normal ordination without features (like `self.pcoa`) doesn't seem to trigger this problem (but please correct me if I've misunderstood this).

Therefore I think it would make sense to use a biplot in this test.

tests/python/test_core.py
```diff
@@ -243,6 +243,14 @@ def test_init_with_ordination(self):
...
+        '''check that empress does not break when ordination has features
+        but empress itself does not'''
```
fedarko, line 250: Could you please move this docstring to the top of this function (above the `viz = Empress...` line)?

tests/python/test_core.py
```diff
@@ -243,6 +243,14 @@ def test_init_with_ordination(self):
...
+        viz.ordination.features = self.feature_metadata
+        self.assertIsNone(viz.features)
```
fedarko, line 252: Sorry, I am not sure I understand these lines of code. Since the `All objects passed were None.` error was being triggered when creating an `Empress` object (which this test already does), I think we can remove these lines and the test should still be ok. (But again, please let me know if I'm missing something!)

fedarko (changes requested):
Thanks for catching this, @ahdilmore! I have a few suggestions that I think will improve the test you added -- please let me know if you have any questions.

tests/python/test_core.py
ahdilmore: Yes, you're correct, thank you!!

tests/python/test_core.py
ahdilmore: I think that the line you added above is redundant with this. 

This line was intended to ensure that the feature_metadata stayed empty even when the ordination had features. In the unmodified code, we received an error concatenating two empty objects when the ordination has features. Line 252 was intended to check that the error was not raised. 

Like you said, I think now that line 248 has been changed, this can be removed.

ahdilmore: While the test is in essence to ensure that the Empress object is created, I'm going to keep line 252 to avoid failing style checks raised when the Empress object is created but not used.

fedarko: Got it. This looks good right now, thanks! I think we could bypass those style checks by removing the `viz = ` part of line 249, but I think leaving just the `assertIsNone` in here makes sense. Thank you for following up on all of this!

fedarko:
Thanks @ahdilmore!

ElDeveloper:
Thanks @ahdilmore and @fedarko 🎉

Issue #479, opened 2021-02-03
Show node circles by default


ElDeveloper:
While turning these off is a nice performance improvement. This can lend itself to misinterpretations.

For example these two leafs have the same node length (see the node detail window), but are visually of much different lengths:

<img width="823" alt="long" src="https://user-images.githubusercontent.com/375307/106683446-60896f80-6579-11eb-814b-9b65e8c1eee6.png">

<img width="807" alt="short" src="https://user-images.githubusercontent.com/375307/106683449-61ba9c80-6579-11eb-9bcd-f79c58270d91.png">

The difference is that one leaf has several internal nodes and the other one doesn't. This is more obvious when you show the node circles:

<img width="996" alt="Screen Shot 2021-02-02 at 5 10 48 PM" src="https://user-images.githubusercontent.com/375307/106683580-9f1f2a00-6579-11eb-9e25-e91e420a2138.png">

fedarko:
As a compromise, could we add a third option besides "show all node circles" and "don't show any node circles" where just internal nodes with one child have circles drawn? This would address the confusion without impacting performance or aesthetics as much as drawing all node circles by default. (This is already the default in existing tools like iTOL -- this is controlled by the `Internal node symbols` option there.)

One of the problems I remember with defaulting to drawing node circles for all nodes (see #349) is that they could be kind of visually overwhelming, especially when you zoom out, since the "point" drawing method Empress uses currently doesn't adjust the size of node circles with zoom level (although changing this would be possible). Since we will be updating the default setting here (and therefore will be determining the most common representation users see when using Empress), I think it would be good to try to make this default a bit more pleasant for users.

-----

Newick string detailing a test tree for this:

```
(((c:3, d:3)b:2)a:1)root:1;
```

In iTOL:

![image](https://user-images.githubusercontent.com/4177727/106691680-c8938200-6588-11eb-8971-4f87d51b000b.png)

ElDeveloper:
> I think it would be good to try to make this default a bit more pleasant for users.

Yes, I like this suggestion the most. "Only show internal node circles" or something along those lines. This should be the default value. And we probably make this a dropdown menu rather than the 3 checkboxes that it would otherwise result in?

PR #486, opened 2021-02-16
Node default

kwcantrell:
This PR address #479. I added new node circle option "Show internal node circles" and made it the default option.

### DIFF
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -262,8 +262,12 @@ define([
         /**
          * @type{Bool}
          * Whether or not to draw node circles.
+         * Values will range between 0 and 2. 
+         * 0 - Show only internal node circles
+         * 1 - Show all node circles
+         * 2 - Do not show node circles
          */
-        this.drawNodeCircles = false;
+        this.drawNodeCircles = 0;
 
         /**
          * @type{Bool}
@@ -512,12 +516,7 @@ define([
      */
     Empress.prototype.drawTree = function () {
         this._drawer.loadTreeColorBuff(this.getTreeColor());
-        if (this.drawNodeCircles) {
-            this._drawer.loadNodeBuff(this.getNodeCoords());
-        } else {
-            // Clear the node circle buffer to save some memory / space
-            this._drawer.loadNodeBuff([]);
-        }
+        this._drawer.loadNodeBuff(this.getNodeCoords());
         this._drawer.loadCladeBuff(this._collapsedCladeBuffer);
         this._drawer.draw();
     };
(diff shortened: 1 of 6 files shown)


empress/support_files/js/empress.js
```diff
@@ -854,9 +853,31 @@ define([
...
         var coords = [];
+        var scope = this;
+        var visible = function (node) {
+            return scope.getNodeInfo(node, "visible");
+        };
+        var comp;
+        if (this.drawNodeCircles === 0) {
+            comp = function (node) {
+                return (
+                    visible(node) && !tree.isleaf(tree.postorderselect(node))
```
ElDeveloper, line 864: This is pseudo-code but I think this would help further reduce the number of circles drawn on screen:

```suggestion
                    visible(node) && !tree.isleaf(tree.postorderselect(node)) && tree.children(...).length === 1```

**Review by ElDeveloper — CHANGES_REQUESTED** [review]
@kwcantrell, thanks so much. This looks good but I think we can reduce the number of circles drawn. Instead of drawing all internal node circles, just show node circles for internal nodes that have only 1 descendant. This will help avoid issues where a branch looks "long" but is really a series of nodes one after the other. It will also reduce overhead by default drawing even fewer nodes.

See the example I posted [here](https://github.com/biocore/empress/issues/479) you'll see there's a few branches with a lot of internal nodes that have only one descendant.

File: empress/support_files/js/empress.js (hunk 1 above)
Comment (kwcantrell) [reply empress/support_files/js/empress.js:864]: I added a forth option (show internal nodes with one descendant) and made it the default. Here is an example.
Screenshot from 2021-02-17 15-54-45

Comment (ElDeveloper) [reply empress/support_files/js/empress.js:864]: Excellent, that looks exactly like what we need! Just for completeness how does that clade in the tree look when you show all node circles?

Comment (kwcantrell) [reply empress/support_files/js/empress.js:864]: Here is the clade. The node highlighted in the above clade is located on the bottom branch.

Screenshot from 2021-02-17 17-54-22

Comment (ElDeveloper) [reply empress/support_files/js/empress.js:864]: Perfect, this looks great! Whenever you get a chance feel free to push the code and I'll be happy to re-review the code.

Comment (kwcantrell) [reply empress/support_files/js/empress.js:865]: my bad. I had commited it but forgot to push it.

File: tests/test-empress.js [hunk 2]
```diff
@@ -105,6 +107,43 @@ require([
             deepEqual(empressUnrootCoords, unrootCoords);
         });
 
+        test("Test getNodeCoords only internal node circles", function () {
+            // have empress draw only internal node circles
+            this.empress.drawNodeCircles = 0;
```
ElDeveloper, line 112: Can you add a test for the case where this value is 3 i.e. the default?

empress/support_files/js/empress.js
```diff
@@ -2597,12 +2628,12 @@ define([
     /**
      * Display the tree nodes.
      *
-     * @param{Boolean} showTreeNodes If true, then Empress will draw circles at
-     *                               each node's position.
+     * @param{String} showTreeNodes 0 - draw only internal nodes circles,
+     *                              1 - draw all node circles,
+     *                              2 - hide all node circles
```
ElDeveloper, line 2633: Can you add the missing value here?

empress/support_files/js/empress.js
```diff
@@ -854,9 +854,40 @@ define([
     Empress.prototype.getNodeCoords = function () {
         var tree = this._tree;
         var coords = [];
+        var scope = this;
+        var visible = function (node) {
+            return scope.getNodeInfo(node, "visible");
+        };
+        var comp;
+        if (this.drawNodeCircles === 0) {
```
ElDeveloper, line 862: Can you add a comment before each `if` describing what case this is?

ElDeveloper (approved):
The result is perfect! I tried it out in a tree and it looks great. Just a couple of comments regarding the tests and comments.

tests/test-empress.js
kwcantrell: test added

empress/support_files/js/empress.js
kwcantrell: added missing value

empress/support_files/js/empress.js
kwcantrell: add comment to each if case

kwcantrell:
thanks @ElDeveloper! The PR should be good to merge in now.

ElDeveloper (approved):
This looks great, thanks so much @kwcantrell!

PR #495, opened 2021-03-18
Fix bug with running standalone version; account for NumPy dependency in standalone install instructions; a few small changes w/r/t node circle options

fedarko:
There's a line in `_plot_utils.py` that imports `q2templates`, a package which is (usually) only available in QIIME 2 conda environments. This was breaking the standalone script when ran outside of a QIIME 2 environment.

To fix this, we can just move the import down so that it's only done if EMPress is being run through QIIME 2. This fixes the problem.

A very warm thank you to @ammaraziz for pointing this out!

Edit - I also made a few changes in this PR's later commits that make the UI for changing the node circle behavior a bit nicer. Should be limited to just cosmetic changes to the HTML.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,14 @@ Empress is available as either a standalone program or a QIIME 2 plugin. The sta
 
 ## Standalone Version
 
-EMPress is available through [PyPI](https://PyPI.org/project/empress/). Run the following command to install Empress:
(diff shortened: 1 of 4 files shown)


ElDeveloper:
Thanks @fedarko !

PR #499, opened 2021-03-18
Creates new GitHub Action for standalone testing

### LINKED ISSUE #496: Explicitly test standalone installation / tests in the CI
One of the reasons that the bug in #495 slipped through, I think, is that the CI runs all of its tests within a QIIME 2 conda environment -- so if there are installation errors made due to installing Empress into an environment that _doesn't_ have all the QIIME 2 stuff pre-loaded, these errors won't be caught on the CI.

As a quick solution, one way to fix this would be adding a chunk of code (a build? a step? I'm not sure yet...) to the GitHub actions script that creates a new "vanilla" conda environment with just Python installed, and then tests that following the installation instructions in the README (see #495 -- need to install Cython/NumPy first) installs Empress properly. To test this environment, we could just run the couple of standalone-specific tests Empress has set up -- nosetest has ways of [only running a handful of tests](https://stackoverflow.com/questions/11286688/how-do-i-specify-a-single-test-in-a-file-with-nosetests).

This would be really useful to have in -- since changes within the Python ecosystem at large can be more volatile than those within a QIIME 2 environment, this sort of setup would help keep us safe from Weird Stuff Happening (tm).

gibsramen:
Closes #496

New workflow "Empress Standalone CI" that creates a non-QIIME2 environment and runs tests that do not require and QIIME2 dependencies. `test_cli` updated to not use the Artifact API.

### DIFF
--- a/.github/workflows/standalone.yml
+++ b/.github/workflows/standalone.yml
@@ -56,10 +56,10 @@ jobs:
       # tests that don't import qiime2 dependencies
       - name: Run Python tests
         shell: bash -l {0}
-        run: nosetests \
-          tests/python/test_cli.py \
-          tests/python/test_compression_utils.py \
-          tests/python/test_core.py \
-          tests/python/test_taxonomy_utils.py \
-          tests/python/test_tools.py \
-          tests/python/test_tree.py
+        run: echo $(ls) && nosetests \
+          ./tests/python/test_cli.py \
+          ./tests/python/test_compression_utils.py \
+          ./tests/python/test_core.py \
+          ./tests/python/test_taxonomy_utils.py \
+          ./tests/python/test_tools.py \
+          ./tests/python/test_tree.py
--- a/.github/workflows/standalone.yml
+++ b/.github/workflows/standalone.yml
@@ -56,10 +56,11 @@ jobs:
       # tests that don't import qiime2 dependencies
       - name: Run Python tests
         shell: bash -l {0}
-        run: echo $(ls) && nosetests \
-          ./tests/python/test_cli.py \
-          ./tests/python/test_compression_utils.py \
-          ./tests/python/test_core.py \
-          ./tests/python/test_taxonomy_utils.py \
-          ./tests/python/test_tools.py \
+        run: >
+          nosetests
+          ./tests/python/test_cli.py
+          ./tests/python/test_compression_utils.py
+          ./tests/python/test_core.py
+          ./tests/python/test_taxonomy_utils.py
+          ./tests/python/test_tools.py
           ./tests/python/test_tree.py
--- a/.github/workflows/standalone.yml
+++ b/.github/workflows/standalone.yml
@@ -1,0 +1,65 @@
+name: Empress Standalone CI
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on pull request and push events, only on the master
+  # branch
+  pull_request:
+    branches: [ master ]
+  push:
+    branches: [ master ]
+
+jobs:
+  # name of job
+  build:
+    # The type of runner that the job will run on (available options are window/macOS/linux)
+    runs-on: ubuntu-latest
+    # we can add more versions of node.js in the future
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    # used in McHelper (similar to TRAVIS_PULL_REQUEST variable)
+    env:
+      BRANCH_NUMBER: ${{ github.event.number }}
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # first grab branch from github
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+
+      - name: Set up Node.js enviroment
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: empress
+          python-version: 3.6
+
+      - name: Install conda packages
+        shell: bash -l {0}
+        run: conda install flake8 nose
+
+      - name: Install cython & numpy
+        shell: bash -l {0}
+        run: pip install cython "numpy >= 1.12.0"
+
+      - name: Install EMPress
+        shell: bash -l {0}
+        run: pip install -e .[all]
+
+      # tests that don't import qiime2 dependencies
+      - name: Run Python tests
+        shell: bash -l {0}
+        run: nosetests \
+          tests/python/test_cli.py \
+          tests/python/test_compression_utils.py \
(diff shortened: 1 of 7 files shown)


tests/python/util.py
```diff
@@ -0,0 +1,73 @@
+import os
+import glob
+import tempfile
+from zipfile import ZipFile
+
+PREFIX_DIR = os.path.join("docs", "moving-pictures")
+
+
+def extracted_artifact_path(dir_name, artifact_loc, file_name):
```
fedarko, line 9: Two things:

1. Could you add a docstring for this function? Doesn't have to be anything fancy or long, just at least a brief description of what this does and what the params do.

2. Unless I'm misinterpreting things, it might make sense to rename this function to `extract_q2_artifact_to_path()` or something more direct than the current name -- the use of `extracted_` makes it seem like this function is just finding a path that already exists, when from what I can tell it's actually creating a path.

tests/python/util.py
```diff
@@ -0,0 +1,73 @@
...
+                                              "feature-table.biom")
+            table = biom.load_table(tbl_loc)
+            pcoa_loc = extracted_artifact_path(_tmp, q2_pcoa_loc,
+                                               "ordination.txt")
+            pcoa = OrdinationResults.read(pcoa_loc)
+            tax_loc = extracted_artifact_path(_tmp, q2_tax_loc,
+                                              "taxonomy.tsv")
+            fmd = pd.read_csv(tax_loc, sep="\t", index_col=0)
+            md = pd.read_csv(md_loc, sep="\t", index_col=0, skiprows=[1])
+        pass
```
fedarko, line 72:
```suggestion
```

tests/python/util.py
```diff
@@ -0,0 +1,73 @@
...
+            Phylogenetic tree.
+        table: Artifact with semantic type FeatureTable[Frequency]
+            Feature table.
+        md: Metadata
+            Sample metadata.
+        fmd: Metadata
+            Feature metadata. (Although this is stored in the repository as a
+            FeatureData[Taxonomy] artifact, we transform it to Metadata.)
+        pcoa: Artifact with semantic type PCoAResults
+            Ordination.
```
fedarko, line 37:
```suggestion

    Parameters
    ----------
    use_artifact_api: bool, optional (default True)
        If True, this will load the artifacts using the QIIME 2 Artifact API,
        and the returned objects will have types corresponding to the first
        listed types (before the | characters) shown below.
        If False, this will instead load the artifacts without using QIIME 2's
        APIs; in this case, the returned objects will have types corresponding
        to the second listed types (after the | characters) shown below.
    
    Returns
    -------
    (tree, table, md, fmd, ordination)
        tree: qiime2.Artifact | skbio.tree.TreeNode
            Phylogenetic tree.
        table: qiime2.Artifact | biom.Table
            Feature table.
        md: qiime2.Metadata | pandas.DataFrame
            Sample metadata.
        fmd: qiime2.Metadata | pandas.DataFrame
            Feature metadata. (Although this is stored in the repository as a
            FeatureData[Taxonomy] artifact, we transform it to Metadata if
            use_artifact_api is True.)
        pcoa: qiime2.Artifact | skbio.OrdinationResults
            Ordination.
```

There may be a more official way to structure this docstring, but I think this makes it pretty clear

tests/python/test_cli.py
```diff
@@ -55,12 +34,21 @@ def setUpClass(cls):
...
+        # glob used because Artifacts unzip to UUIDs
+        cls.tree_loc = extracted_artifact_path("tree", q2_tree_loc, "tree.nwk")
+        cls.table_loc = extracted_artifact_path("tbl", q2_table_loc,
+                                                "feature-table.biom")
+        cls.fm_loc = extracted_artifact_path("fm", q2_fm_loc, "taxonomy.tsv")
+        cls.pcoa_loc = extracted_artifact_path("pcoa", q2_pcoa_loc,
+                                               "ordination.txt")
+        # need to re-save sample metadata to remove q2:types row
+        cls.sm_loc = "tmp_sample_metadata.tsv"
+        pd.read_csv(q2_sm_loc, sep="\t", index_col=0, skiprows=[1]).to_csv(
```
fedarko, line 47: Great point, thanks for bringing this up.

tests/python/test_cli.py
```diff
@@ -55,12 +34,21 @@ def setUpClass(cls):
...
         cls.iso_fs.__enter__()
 
-        nwk_tree.write(cls.tree_loc)
-        with biom.util.biom_open(cls.table_loc, "w") as f:
-            biom_tbl.to_hdf5(f, "test")
-        sm.to_csv(cls.sm_loc, index=True, sep="\t")
-        fm.to_csv(cls.fm_loc, index=True, sep="\t")
-        pcoa.write(cls.pcoa_loc)
+        # extract Artifacts to temporary filesystem
+        # glob used because Artifacts unzip to UUIDs
```
fedarko, line 38:
```suggestion
```

(I think this comment should still be kept around in the codebase, but it should be moved down to the `util.extract...()` function's docstring)

fedarko (changes requested):
Thanks so much @gibsramen! This is super helpful.

.github/workflows/standalone.yml
```diff
@@ -0,0 +1,66 @@
+name: Empress Standalone CI
```
fedarko, line 1: ... Oh, one more comment I thought of (sorry for spamming you with a million): it would be nice to add a badge to the README for this workflow, similarly to the other workflow. This'd be useful, since I imagine that this build failing (but the other not) could actually happen in practice.

emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F499%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F499%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F499%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F499%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F499%2Fplain.qzv)

fedarko (approved):
Everything looks good to me, thanks so much @gibsramen!!!

PR #502, opened 2021-03-19
BLD: Small tweaks to GH Actions workflows

fedarko:
Nothing big, just some small fixes that should make development easier.

- Remove Node.js matrix for main CI, since the version number is hard-coded currently; this should remove the need to click through multiple pages when accessing the main CI (edit: actually maybe you still need to click through things twice, hm. at least it should simplify the code.)
  - In the future we could add back this matrix and use it to try multiple Node.js versions -- however I don't think it's an urgent priority, since Node.js is only used on the development side (so being exhaustive about versions working probably isn't too important, at least compared to Python versions in the standalone CI)
  - Another concern here is that trying multiple Node.js versions should probably be a separate process from installing/testing the Python code; I think the most efficient way to set this up this would be something like adding a different workflow for JS testing, then adding _another_ workflow for uploading to McHelper that is dependent on both the "JS testing" and "Python testing" workflows passing [but IDK how to do that right now]

- Make some documentation consistent btwn YAML files (e.g. Empress -> EMPress)

- Remove the "Empress" from the workflow names, renaming them to just "Main CI" and "Standalone CI" -- I think this makes the README badges/etc. easier to interpret

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,3 +1,7 @@
+# Tests EMPress' Python and JS code within a QIIME 2 conda environment.
+# Also lints and stylechecks the code (both Python and JS).
+# When tests pass, this runs "make docs" to generate various test QIIME 2
+# visualizations (.qzv files) and uploads them to McHelper.
 name: Main CI
 
 # Controls when the action will run. 
@@ -42,7 +46,7 @@ jobs:
       - name: Install flake8
         run: conda run -n base pip install flake8
 
-      - name: Run linters
+      - name: Run linting and stylechecking
         run: |
           conda run -n base make stylecheck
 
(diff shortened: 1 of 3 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F502%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F502%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F502%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F502%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F502%2Fplain.qzv)

ElDeveloper (approved):

ElDeveloper:
Thanks @fedarko !

PR #504, opened 2021-03-22
[WIP] Barplot clicking

fedarko:
This isn't done yet -- there are still some major issues that need to be fixed (see #308) before merging this in. However, I wanted to set up this PR so that this prototype can be easily tested out, at least.

cc @ElDeveloper

Issue #322, opened 2020-08-11
Support drawing variable-length stacked sample metadata barplots


fedarko:
Where, rather than having all tips' bars be of equal length and coloring them based on proportions, bars' lengths are scaled based on the number of samples total that contain a tip.

This would probably look similar to the bar plots in Fig. 2A of [this paper](https://www.biorxiv.org/content/10.1101/804443v2.full.pdf), with the difference that those graphs use relative abundance where this would just be based on sample presence count. (We could totally scale bars by actual abundance data, also, but that'd require a lot of restructuring to do.)

Brought up by @kwcantrell in https://github.com/biocore/empress/pull/313#discussion_r468606735.

PR #505, opened 2021-03-25
Shear UI

kwcantrell:
I still need to finish documenting and add a couple of tests but in the meantime if @ElDeveloper and @fedarko can check the .qzv for UI bugs (i.e. test different combinations of coloring/shearing) to see if you can find errors that I did not count for that would be great! This PR branches from tree-controller so #492 needs to be merged in before this PR.


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F505%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F505%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F505%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F505%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F505%2Fplain.qzv)

empress/support_files/js/legend.js
```diff
@@ -218,17 +218,8 @@ define(["jquery", "underscore", "util"], function ($, _, util) {
      * @param {Object} info Color key information. This should map unique
      *                      values (e.g. in sample or feature metadata) to
      *                      their assigned color, expressed in hex format.
-     *
-     * @throws {Error} If info has no keys. This check is done before anything
-     *                 else is done in this function.
      */
     Legend.prototype.addCategoricalKey = function (name, info) {
-        if (_.isEmpty(info)) {
```
ElDeveloper, line 226: Why is the error check being removed?

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,208 @@
+define(["underscore", "util", "TreeController"], function (
+    _,
+    util,
+    TreeController
+) {
+    function Layer(
```
ElDeveloper, line 6: Maybe Filter?
```suggestion
    function Filter(
```

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,208 @@
+define(["underscore", "util", "TreeController"], function (
```
ElDeveloper, line 1: Great to see this module is using MVC. Even though objects are not exported outside of this closure, I would still suggest that you namespace them, for example: ShearingLayer, ShearModel....

ElDeveloper (changes requested):
Thanks so much @kwcantrell, looks **GREAT** and it is useful. Here's a few high-level observations:

- I noticed some oddities with the barplots when you filter down to very few tips in the tree:

Notice the selection of only Bacteria and only Thermi.
Screen Shot 2021-03-29 at 11 28 34 AM

Seems like it's also an issue with the rectangular layout:
Screen Shot 2021-03-29 at 11 28 50 AM


- Do you think we should include a warning message that reminds users that hiding elements in the tree can potentially be misleading?
- Do you also think we should also include a warning about the ordination not being affected by any of the visibility changes?
- Can we have an easy way to deselect all from a layer? 
- The shearing panel shows up in the UI even when there's no feature metadata, see for example [`plain.qzv`](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F505%2Fplain.qzv).

kwcantrell:
Thanks @ElDeveloper! Those oddities you pointed out are going to be bugs. When implementing the shear I did not really test out the sample metadata for the barplots as I thought it would behave like feature metadata but I guess that is not the case. I'll need to look the code to figure out what is causing it.

ElDeveloper:
Thanks for looking into it!

empress/support_files/js/empress.js
```diff
@@ -2177,6 +2224,11 @@ define([
         var cm = colorer.getMapRGB();
         // colors for the legend
         var keyInfo = colorer.getMapHex();
+        for (var key in keyInfo) {
+            if (obs[key].length === 0) {
+                delete keyInfo[key];
+            }
+        }
```
fedarko, line 2231: Not sure I understand the purpose of this. If we don't care about sample metadata values that are no longer represented by tips in the tree, wouldn't it be more efficient for `getUniqueSampleMetadataInfo()` to just not contain these values in the first place? Or is the purpose of this to make sure that these values still "impact" the colors used in the legend? (e.g. I assume this is why the `right palm` sample metadata value is assigned a color of orange instead of red here, after shearing to just Level 1 = Archaea features.)

orange

empress/support_files/js/empress.js
```diff
@@ -3565,5 +3626,63 @@ define([
...
+            keepNames.push(name);
+        }
+
+        this._tree.shear(new Set(keepNames));
+        var nodeNames = this._tree.getAllNames();
+        // Don't include nodes with the name null (i.e. nodes without a
+        // specified name in the Newick file) in the auto-complete.
+        nodeNames = nodeNames.filter((n) => n !== null);
+        this._events.autocomplete(nodeNames);
+
```
fedarko, line 3669: Looks like the canvas should be updated (?) after shearing -- currently, shearing doesn't remove old collapsed clades until zooming/panning is done (notice how the red triangles here linger):

colasp

empress/support_files/js/empress.js
```diff
@@ -2142,6 +2176,18 @@ define([
         this.drawTree();
     };
 
+    /**
+     *
+     */
+    Empress.prototype.getUniqueSampleMetadataInfo = function (cat) {
```
fedarko, line 2182: I know you mentioned docs were a work in progress, but this function in particular could really use some details -- I am unsure why exactly this has to be used in place of `getObsBy` (I get that this removes tips that have been sheared from the tree, but it isn't clear where this change is useful "downstream" later on in `colorBySampleCat`).

fedarko (changes requested):
This is a really neat change, thanks! It looks like there are a decent amount of things that could stand to be improved before we throw this at users; let me know if you have any questions.

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,208 @@
...
+        _.each(this.fVals, function (val) {
+            var row = document.createElement("tr");
+
+            // add checkbox
+            var dataCheck = document.createElement("td");
+            var input = document.createElement("INPUT");
+            input.setAttribute("type", "checkbox");
+            input.checked = true;
+            input.onchange = function () {
+                chkBxClickFunction(!input.checked, scope.fCol, val);
```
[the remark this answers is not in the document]
kwcantrell: Personally I want to avoid using a "update" button for this feature. The main reason being it provides users with visual feedback when they remove something from the tree. Additionally, (as you mentioned) the "update" button makes the UI a bit clunky to use. 

Per @ElDeveloper suggestion, I plan on adding an additional checkbox that would allow users quickly un-select all checkboxs in a particular filter so that they can easily select a couple of values.

With that being said, if the consensus is to add an update button I will. @ElDeveloper what do you think?

tests/test-legend.js
```diff
@@ -156,12 +156,6 @@ require([
             deepEqual(legend._sortedCategories, ["hjkl"]);
             deepEqual(legend._category2color, colorInfo);
         });
-        test("addCategoricalKey (error: no categories)", function () {
-            var legend = new Legend(this.containerEle);
-            throws(function () {
-                legend.addCategoricalKey("oops", {});
-            }, /Can't create a categorical legend when there are no categories in the info/);
-        });
```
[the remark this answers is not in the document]
kwcantrell: It is possible to remove all feature metadata categories in cases where all tips where removed from the tree or the remaining tips do not have metadata associated with them.

empress/support_files/js/empress.js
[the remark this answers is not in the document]
kwcantrell: > The error you linked was caused by line 167 of canvas-events.js: for (var node = 1; node <= empress._tree.size; node++) {. I forgot to change it to use the new iterators of TreeController. So, I fixed the error by changing line 167 to for (var node of this._tree.postorderTraversal(includeRoot = true)) {

This was the cause of this issue.

empress/support_files/js/empress.js
kwcantrell: > Or is the purpose of this to make sure that these values still "impact" the colors used in the legend? (e.g. I assume this is why the right palm sample metadata value is assigned a color of orange instead of red here, after shearing to just Level 1 = Archaea features.)

Correct. I still want values to get assigned the same color. If `getUniqueSampleMetadataInfo()` removes these values then every time a shear takes place the colors will change.

empress/support_files/js/empress.js
kwcantrell: Thanks for catching this!

empress/support_files/js/empress.js
kwcantrell: I agree documentation is definitely needed!

Reason this is used in place of `getObsBy` is that the `BIOMTable` object does not access to the tree so it is unable to remove the nodes and I don't want this to be "downstream" in `colorBySampleCat` so we can avoid having to implement this in future functions if they require sample metadata.

empress/support_files/js/shearer.js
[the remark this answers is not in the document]
ElDeveloper: I would prefer if we could move away from the `Update` button pattern whenever possible. The value of having feedback with every action shouldn't be discounted. In general we should strive to make the code faster such that we don't need an `Update` button to cope with bottlenecks. @kwcantrell can you try this feature in the EMP dataset and see if it is slow to hide/show groups in the tree?

IIRC, tree coloring and barplot drawing are time-consuming and each modifying parameter (colorscheme, variable, type interpretation, etc) has significant overhead that is best avoided if we perform several operations simultaneously.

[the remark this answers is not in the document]
kwcantrell: I'll test it out. Shearing its self is a linear alg so I dont expect it to be that slow. With that being said, it might be slow if the tree is already colored with feature/sample metadata. But I guess we will see.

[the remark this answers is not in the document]
fedarko: Thank you both for your detailed insight! Agreed that ideally things would be fast enough so that there'd be no need for `Update` buttons; if shearing is fast enough to be decently responsive on the EMP dataset (less than a few seconds per checking / unchecking a feature metadata field), I'm cool with not having an `Update` button. However, if a single shearing operation takes longer than that, my vote is to bite the bullet and include such a button here (at least for the time being).

The idea to have a "select all / unselect all" checkbox sounds dope -- that would work really well with this.

(The "worst-case scenario" situation I'm imagining here is a user new to EMPress clicking on a feature metadata field in the shearing options and then seeing the application freeze in place for a few seconds, which would not be ideal. Deferring the work to these Update buttons reduces the amount of places where the application can get stuck in place doing something, I guess? Although for small datasets this really isn't a problem.)

empress/support_files/js/empress.js
```diff
@@ -391,21 +391,32 @@ define([
...
-                // option for the Layout functions since the layout function only
-                // needs to know lengths in order to layout a tree, it doesn't
-                // really need encapsulate all of the logic for determining
-                // what lengths it should lay out.
-                this.leafSorting,
-                undefined,
-                lengthGetter,
-                checkLengthsChange
-            );
+            // tree is just root
```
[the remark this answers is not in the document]
fedarko: Gotcha, thanks! This makes sense to me. In this case, would it be possible to hide the single node circle shown here? Since I think shearing all tips would imply that there are no nodes left in the tree, we could just temporarily hide the canvas, I guess (since at that point there is nothing meaningful left to show).

I really like the idea of displaying a message alerting users about shearing (copying text from @ElDeveloper above: "Do you think we should include a warning message that reminds users that hiding elements in the tree can potentially be misleading?"). You may have already started looking into this, but it might be nice to format the warning similarly to how EMPeror's warning message is formatted, with a similar box in a corner of the screen saying `x / y nodes visible; warning: shearing the tree can be misleading` or something like that (see top-left corner in the screenshot below).

emperor-hidden

empress/support_files/js/empress.js
fedarko: Makes sense, thanks! In this case I'd suggest adding a brief comment explaining this (or just linking to this discussion thread lol), to make it clearer to people reading the code.

empress/support_files/css/empress.css
```diff
@@ -585,6 +585,13 @@ p.side-header button:hover,
     white-space: nowrap;
 }
 
+.shear-layer-legend div {
```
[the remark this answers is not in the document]
kwcantrell: When an overflow occurs I wanted the title of the filter to always be in view. For example if you select "Level 7" for the shear filter, you can scroll and always see "Level 7". This occurs because `.shear-layey-legend` only applies to `div` elements within the `shear-legends` container.  If  I used `barplot-layer-legend` then the title of the filter would also scroll and if I added div to `barplot-layer-legend` then it would break the barplot legend.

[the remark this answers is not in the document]
kwcantrell: There's probably a better solution but that is what was easiest for me.

empress/support_files/js/empress.js
```diff
@@ -3565,5 +3626,63 @@ define([
...
+            _.each(values, function (val) {
+                var obs = uniqueValueToFeatures[val];
+                for (var node of obs) {
+                    removeNodes.add(node);
+                }
+            });
+        });
+
+        // remove removeNodes
+        var allNodes = Array.from(Array(this._tree.size + 1).keys());
```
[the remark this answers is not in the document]
kwcantrell: Good idea. :smile:

empress/support_files/js/legend.js
kwcantrell: This is removed to handle cases where all tips are removed.

For example:

clicking on `k__Archaea` leads to:

empress/support_files/js/empress.js
```diff
@@ -391,21 +391,32 @@ define([
...
-                // what lengths it should lay out.
-                this.leafSorting,
-                undefined,
-                lengthGetter,
-                checkLengthsChange
-            );
+            // tree is just root
+            if (this._tree.currentSize == 1) {
+                data = {
+                    xCoord: [null, 0],
```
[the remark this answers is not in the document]
kwcantrell: Instead of setting the value of each component, I create the following helper function:
```
var dataForOnlyRoot = function(coordKeys) {
    var rootCoordData = {};
    _.each(coordKeys, function(key) {
        rootCoordData[key] = [null, 0];
    });
    return rootCoordData;
};
```

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,208 @@
...
+            };
+            dataCheck.appendChild(input);
+            row.appendChild(dataCheck);
+
+            // add checkbox label
+            var dataLabel = document.createElement("td");
+            dataLabel.innerText = val;
```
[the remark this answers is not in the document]
fedarko: I just tried out the latest code on the PR, and it doesn't look like the click event is added yet -- is this code that hasn't been pushed yet?

Hm, I didn't read things carefully enough to realize that the label element was a `<td>` instead of a `<label>`. Besides the clicking thing, an advantage of using "real" labels is that hovering over a label also (at least in EMPress' CSS) highlights the corresponding checkbox (see GIF [here](https://user-images.githubusercontent.com/4177727/113232785-b36a6500-9252-11eb-8f09-78f8c224c597.gif)).

I took some time and modified the `shearer.js` code to place labels within `<td>`, rather than just using `<td>` -- this enables use of the `for` attribute, and avoids the need for custom click events. I also added a basic way of assigning unique IDs to each checkbox. Hopefully this is helpful!

**[shearer.js](https://github.com/biocore/empress/files/6240614/shearer.js.txt)**

(I realize that including lots of checkbox inputs with `class="empress-input"` results in the table having weirdly thick rows, because the `empress-input` class sets a strange fixed height -- so the attached code doesn't use the `empress-input` class. It still looks reasonable enough tho -- see GIF.)

shearing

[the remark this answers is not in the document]
kwcantrell: Sorry I have not pushed the code. I'll push what I have in a sec.

empress/support_files/css/empress.css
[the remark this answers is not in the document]
fedarko: A simple-ish solution might be creating a title element for each shear layer (for example, the `Layer N` text in barplot layers), and then not having a title at all of the stuff in the shear layer legend. Since each feature metadata column can correspond to at most one shear layer, the title we assign a shear layer could be `Shear Layer: Level 4` or something like that?

image

empress/support_files/js/shearer.js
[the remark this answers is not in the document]
kwcantrell: The current code should be the same as when you first looked at the PR

kwcantrell:
@fedarko Sorry for the slight confusion. I have now pushed the most current code I have. Improvements:

1) Sample barplots should now function properly. (Although I am still cleaning up/improving it)
2) Everything I have marked as `resolved` is complete. ( Again still working on cleaning things up/documentaion)
3) Still need documentation

empress/support_files/js/empress.js
```diff
@@ -3565,5 +3626,63 @@ define([
...
+            [...allNodes].filter((x) => !removeNodes.has(x))
+        );
+
+        var keepNames = [];
+        for (var node of keepNodes) {
+            var name = this._tree.name(this._tree.postorderselect(node));
+            keepNames.push(name);
+        }
+
+        this._tree.shear(new Set(keepNames));
```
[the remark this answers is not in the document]
ElDeveloper: What I meant is that some of the shearing functionality in `empress.js` can be moved to `tree-controller.js`. Namely the part where you figure out which tip identifiers to shear. My suggestion is to take the data in `shearMap` (the argument that's currently being passed to `Empress.shear`) and pass that to `TreeController` for example `TreeController.shearToMap` (or `shearWithMap` up to you). That way additional checks like verifying that names aren't null, etc can all be done in `TreeController` rather than in `Empress`.

If this doesn't make sense, let me know and we can hop on a quick call.

kwcantrell:
@fedarko and @ElDeveloper I have finished addressing your comments/documenting the code. I also profiled the shear operation with the EMP tree. It takes about 1.5-2 seconds to shear the tree plus the time it takes to recolor the tree/create the barplots if they were there before the shear. So, I added a warning message letting users know if shearing takes a while to disable coloring/barplots until they finish shearing the tree. #371 will also speed up help speed up this process.

Also, I cache parent nodes in bp tree which helps speed up the and also improved the performance of _projectObservations() which both help improve the performance of the coloring methods.

empress/support_files/js/bp-tree.js
```diff
@@ -984,41 +1014,58 @@ define(["ByteArray", "underscore"], function (ByteArray, _) {
      *                  convert the original postorder positions to the sheared
      *                  tree postorder positions ("newToOld") and vice-versa ("oldToNew").
      */
-    BPTree.prototype.shear = function (keepTips) {
+    BPTree.prototype.shear = function (removeTips) {
```
ElDeveloper, line 1017: Can you update the rest of the docstring there's still some references to `keepTips`.

empress/support_files/js/shearer.js
ElDeveloper: Thanks @kwcantrell

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,373 @@
...
+     * @param{Object} obs The object to register.
+     */
+    ShearModel.prototype.registerObserver = function (obs) {
+        this.observers.push(obs);
+    };
+
+    /**
+     * Removes a shear layer from a ShearModel
+     * @param{ShearModel} model The ShearModel to use
+     * @param{String} layer The name of layer to remove.
```
ElDeveloper, line 226: I absolutely missed this before but it looks like there's several methods here that are class methods. Do you have an intuition for why this is the case?

empress/support_files/templates/side-panel.html
```diff
@@ -257,6 +257,35 @@
...
 </div>
 
+<!-- Shear options -->
+<button id="shearOpenButton" class="side-header collapsible needs-feature-metadata">
+    Shear Tree</button>
+<div class="side-content control hidden" id="shear-div">
+  <p class="side-panel-notes">
+    Shear the tree to tips with certain feature metadata values.<p></p>
+    WARNING: shearing the tree can take awhile if Feature/Sample Metadata Coloring and/or
+      Barplots are active thus if you are experiencing a delay, we recommend to displable these
```
ElDeveloper, line 267:
```suggestion
      Barplots are active thus if you are experiencing a delay, we recommend to disable these
```

empress/support_files/templates/side-panel.html
```diff
@@ -257,6 +257,35 @@
...
+      feature will shearing the tree.
```
ElDeveloper, line 268:
```suggestion
      features while shearing the tree.
```

tests/test-bp-tree.js
```diff
@@ -1037,15 +1037,23 @@ require(["jquery", "ByteArray", "BPTree"], function ($, ByteArray, BPTree) {
...
             deepEqual(result.tree.lengths_, [null, 11]);
             deepEqual(result.shearedToFull, shearedToFull);
             deepEqual(result.fullToSheared, fullToSheared);
+
+            remove = new Set([3, 6, 7, 8]);
+            result = preShearBPTree.shear(remove);
+            deepEqual(
+                result.tree.names_,
+                [null, "3", "4", "2", "r"]
+                // [null, "3", "4", "2", "7", "10", "11", "9", "8", "r"]
```
ElDeveloper, line 1055:
```suggestion
```

tests/utilities-for-testing.js
```diff
@@ -23,7 +23,7 @@ define(["Empress", "BPTree", "BiomTable"], function (
         // tree comes from the following newick string
         // ((1,(2,3)4)5,6)7;
         var tree = new BPTree(
-            new Uint8Array([1, 1, 1, 0, 1, 1, 0, 1, 0, 0, 0, 1, 0, 0]),
```
ElDeveloper, line 26: Seems a bit odd, any reason why this is needs to be a JavaScript array?

ElDeveloper (approved):
@kwcantrell I've added a few comments, nothing major. Overall this is working very smoothly.  I particularly like the ability to use this in conjunction with the animations.

I noticed one glitch, I wanted to check if others are seeing this. When I unselect all and then select all, the checkboxes in other panels are misaligned for a second. Looks like a browser glitch but I don't know if it's specific to Safari or if others can reproduce this:

glitch

The second thing I noticed is that when shearing the node info window should be hidden if the node in focus is being hidden.

edge-case

Overall, this is a wonderful new feature and is also rather snappy even in the EMP dataset.

tests/utilities-for-testing.js
kwcantrell: I switched it to a js arrray because BPTree no longer uses Uint8Array (we switched when we introduced the tree compression on the python side). Also _.clone() for whatever reason does not work properly with _.clone() which is used in tree shearing

ElDeveloper:
Thanks for clarifying!

empress/support_files/js/shearer.js
kwcantrell: I wanted to make ShearLayer very flexible so I up its constructor to accept general functions that will intern be called when users do things like unselect a checkbox. However javascript handles `this` in an odd way (https://stackoverflow.com/questions/20279484/how-to-access-the-correct-this-inside-a-callback/20279485#20279485). Essentially, `this` is not binded to anything rather its interpreted during execution. So if I passed an object's function `this` would not reference the object but rather ShearLayer so it would throw an error. Thus I made those function Class function and passed in the model as a parameter ([example](https://github.com/biocore/empress/blob/3b6d854953c33a1d927d656835f403dde64ac5b0/empress/support_files/js/shearer.js#L165-L167))

ElDeveloper: I understand, this is such an annoying "feature" of the language. Thanks for the explanation.

ElDeveloper:
Per @jpshaffer's recommendation, @kwcantrell can you add a note in the Shearing panel to warn about this being potentially misleading?

empress/support_files/templates/empress-template.html
```diff
@@ -209,6 +215,16 @@ <h3 id="menu-sm-header">Sample Presence Information</h3>
             $(".needs-feature-metadata").addClass("hidden");
         }
 
+        var statsButton = document.getElementById("stats-btn");
```
fedarko, line 218: I think I understand the logic of this block -- this is clever. I think we should add a comment here briefly explaining what is going on, however, so that readers are not lost.

empress/support_files/templates/empress-template.html
```diff
@@ -209,6 +215,16 @@ <h3 id="menu-sm-header">Sample Presence Information</h3>
...
+        statsButton.shearUpdate = () => {
+          statsButton.classList.remove("unpopulated");
+          if (!statsButton.classList.contains("unpopulated")) {
+            sPanel.populateTreeStats();
+          }
```
fedarko, line 223:
```suggestion
          sPanel.populateTreeStats();
          statsButton.classList.remove("unpopulated");
```

I don't think there is ever a case where the conditional in the `if` statement isn't true, right? Since we remove the `unpopulated` class immediately beforehand.

Also, it may be worth taking these two lines of code and making them into a function since they're reused in this file... but not a big deal.

empress/support_files/templates/empress-template.html
```diff
@@ -209,6 +215,16 @@ <h3 id="menu-sm-header">Sample Presence Information</h3>
...
+        };
+        shearer.registerObserver(statsButton);
```
fedarko, line 225: I didn't actually realize that the shearing code updated the tree statistics panel -- thanks for the thorough work.

I noticed from some testing that the behavior of the tree stats seems strange when everything in the tree is sheared: notice how the node lengths are all messed up here, since the only node left is the root (and we ignore the root's length):

test

I think one good way to handle this would be having each of these three length values say `N/A` or `N/A: only node left is root` or something like that. I think this check could be done in `BPTree.getLengthStats()`, where if `this.lengths_.length` is 2 (I think? since it's 1-indexed) then it just returns `{ min: "N/A", max: "N/A", avg: "N/A" };` or something.

That all said, if you feel this is too much of a hassle for right now, I suggest we just comment out this part of the code for now and add a warning that the tree stats aren't updated during shearing. We could add this functionality back in at a later date in this case.

empress/support_files/templates/side-panel.html
```diff
@@ -257,6 +257,38 @@
...
+  <div id="shear-filter-container">
+    <hr />
+  </div>
+  <div id="shear-legends">
+  </div>
+</div>
+
 <!-- animate Options -->
 <button id="animationOpenButton" class="side-header collapsible hidden">
     Animation</button>
```
fedarko, line 294: Another instance of "github isn't letting me comment on this line but it really should be changed in this PR". In the tree stats section, there is a paragraph as follows:

```html
  <p class="needs-community-data side-panel-notes">
    Also note that these statistics only pertain to the tree shown here; if the
    tree was sheared to a table, for example, the node counts shown here will
    be lower than for the full tree.
  </p>
```

I propose we update it to:

```html
  <p class="needs-feature-metadata side-panel-notes">
    These statistics will be automatically updated as the tree is sheared
    using the "Shear Tree" functionality.
  </p>
  <p class="needs-metadata side-panel-notes">
    Note that these statistics only pertain to the tree shown here,
    and don't account for modifications to the tree done before this
    visualization. For example, if the
    tree was sheared to a table using Empress' command-line interface
    (this is the default behavior of the
    <span style="font-family: monospace;">shear-to-table</span>
    parameter of <span style="font-family: monospace;">community-plot</span>)
    then the node counts shown here will only reflect the
    resulting sheared tree.
  </p>
```

This avoids confusion between `--p-shear-to-table` and the dynamic shearing added in this PR.

fedarko (changes requested):
@kwcantrell this is looking really nice, thanks! I have not gone over everything in depth yet, but I have some initial comments based on the codebase/UI. I will try to make another pass over the codebase in some more detail soon, but I think these suggestions should be a good starting point. Let me know if you have any questions.

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,373 @@
...
+        var legendChkBoxs = document.createElement("div");
+        chkBoxLegendDiv.appendChild(legendChkBoxs);
+
+        // create checkboxes
+        var table = document.createElement("table");
+        legendChkBoxs.appendChild(table);
+        _.each(this.fVals, function (val) {
+            scope.values.push(val);
+            var row = document.createElement("tr");
+            var id = scope.fCol.replace(" ", "-") + "-" + val;
```
fedarko, line 77: I think this will cause problems. `Test A` and `Test-A` are two distinct metadata column names, but this code will assign their checkboxes identical IDs (both will end up with `Test-A-[val]`, so if they share any metadata values then this code will break).

Obviously that is ... pretty unlikely, but we should avoid this scenario if at all possible. To do this, we should define a unique number for each shearing layer and then use that in all of the IDs -- this would ensure IDs are not duplicated across shearing layers. The barplot panel / layer JS has [code that could be adapted for this](https://github.com/biocore/empress/blob/5b8a340bbff046f9d1d7f992248db87224b13c7a/empress/support_files/js/barplot-layer.js#L26-L42).

empress/support_files/js/shearer.js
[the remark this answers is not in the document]
fedarko: The current solution is basically fine, although I still think it would be good in the future to make `dataLabel` actually a label so we can use the `for` attribute (which as discussed has additional benefits, e.g. highlighting the corresponding checkbox on hover -- this would also make the code cleaner since it'd remove the need for a bunch of `onclick` events).

I'm ok if you'd prefer to keep this as is for this PR, though. Will leave the choice up to you. (Although IMO we should open an issue if we don't address this now -- fixing this will make the code easier to understand.)

empress/support_files/templates/side-panel.html
```diff
@@ -257,6 +257,38 @@
...
+      <label for="shear-feature-select">Shear by...</label>
+      <label class="select-container">
+        <select id="shear-feature-select"></select>
+      </label>
+    </p>
+    <p>
+      <label for="shear-add-btn">Add shear filter</label>
+      <button id="shear-add-btn">+</button>
+    </p>
+  </div>
```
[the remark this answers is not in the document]
kwcantrell: Rather than swapping the elements around, I add each new layer to the front of the div container. I changed line 31 on shearer.js to `this.container.insertBefore(this.layerDiv, this.container);`

Here is an example of the effect:
add "Level 1"
Screenshot from 2021-05-03 15-42-33
add "Level 2"
Screenshot from 2021-05-03 15-43-08

I think this is a more natural way of handling this as the newest layer will appear on top.

tests/test-bp-tree.js
```diff
@@ -928,7 +928,7 @@ require(["jquery", "ByteArray", "BPTree"], function ($, ByteArray, BPTree) {
                 null
             );
 
-            var keep = new Set(["4", "6", "7", "10", "11"]);
+            var remove = new Set([1]);
```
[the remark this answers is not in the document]
kwcantrell: I'll add a comment.

This modification was implemented to better improve the performance of the shear operation.

empress/support_files/templates/side-panel.html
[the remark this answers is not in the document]
kwcantrell: Instead of swapping the positions of the UI elements, I add the new shear layers to the top.

Here is adding `Level 1`
Screenshot from 2021-05-04 11-05-39

and adding `Level 2`
Screenshot from 2021-05-04 11-05-55

empress/support_files/js/empress.js
```diff
@@ -914,7 +946,6 @@ define([
         } else {
             throw new Error("getNodeCoords() drawNodeCircles is out of range");
         }
```
fedarko, line 948: The relevant block of code is a bit up from this, but I have a maybe weird question: why is a node circle drawn for the root when all nodes are sheared? It's possible to hide this circle when selecting the `Hide all node circles` option, but the circle is still drawn when `Show node circles for internal nodes with only one child`, the default, is used.

It isn't a big deal or anything -- but in the future we may want to use a special symbol for "just the root", to avoid confusion. Since, going by the way the tree statistics are updated in the "just sheared to root" case, the root counts as a tip -- but this would mean that by default it should not have a node circle drawn.

empress/support_files/templates/side-panel.html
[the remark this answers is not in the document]
fedarko: Fair! I think I would still prefer the swapped UI element ordering -- if nothing else, for consistency with the barplot UI -- but the current behavior is fine for this next release. We can adjust things as needed in the future.

empress/support_files/js/empress.js
```diff
@@ -2225,6 +2256,15 @@ define([
         // colors for the legend
         var keyInfo = colorer.getMapHex();
 
+        // if the tree has been sheared then categories in obs maybe empty.
```
fedarko, line 2259: I have two minor comments/questions about the shearing-all-tips-off case that I wanted to bring up -- both involve sample metadata coloring, so I'm raising them here.

**First minor question about legends**: the behavior of legends differs between feature and sample metadata tree coloring. With feature metadata coloring, an empty legend with no categories is shown; with sample metadata coloring, no legend at all is shown.

gif1

Do you think these cases should be consistent? This is not important at all, but I wanted to bring it to your attention in case this is actually a symptom of a more serious problem with the code (p sure not but wanted to check).

empress/support_files/js/empress.js
```diff
@@ -2225,6 +2256,15 @@ define([
...
+        // getObsBy() does not filter out those categories so that the same
```
fedarko, line 2260: **Second minor question about legends:** this behavior changes when sample metadata coloring has already been active on the tree, and then it is sheared. In this case, the "partial" legend (e.g. just the right palm body-site fm val) is shown, even though the tree is empty.

Additionally, this legend does not go away until the tree is restored. This can look sort of jarring when the user tries to select other sample metadata fields to color with, since the legend remains not updated:

gif2

Would it be possible to hide the main legend when sample metadata coloring is active and the tree is just sheared to the root? This might solve this problem, although idk if there is a more elegant solution.

This is also not a very important issue, but I wanted to bring it up.

fedarko (commented):
Thanks so much @kwcantrell! This is really close. Please let me know if you have any questions.

empress/support_files/js/biom-table.js
```diff
@@ -591,8 +604,25 @@ define(["underscore", "util"], function (_, util) {
...
         return fID2Freqs;
     };
 
+    /**
+     * Set which features to ignore. Features in this set will not be
+     * considered in funcitons such as getObsBy() or getFrequencyMap()
+     *
+     * @param {Set} nodes A set of feature ids to ignore
+     */
+    BIOMTable.prototype.setIngnoreNodes = function (nodes) {
```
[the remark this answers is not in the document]
kwcantrell: lol, good catch.

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,418 @@
...
+
+            scope.inputs.push(input);
+            dataCheck.appendChild(input);
+            row.appendChild(dataCheck);
+
+            // add checkbox label
+            var dataLabel = document.createElement("label");
+            dataLabel.setAttribute("for", input.id);
+            dataLabel.innerText = val;
+            row.appendChild(dataLabel);
```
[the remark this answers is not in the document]
kwcantrell: Thanks!

empress/support_files/js/bp-tree.js
kwcantrell: Thanks, I updated the docs for this function.

empress/support_files/js/empress.js
kwcantrell: good catch!

empress/support_files/js/empress.js
kwcantrell: Good point, it should be consistent!

empress/support_files/js/empress.js
kwcantrell: I kept the node circle, when its just the root, so that users can visually see that only the root is left. When I remove the node circle, all you see is a white screen. This can be a problem because canvas events are still active and thus the user can drag the tree off screen without realizing and when they unshear part of the tree it still be off screen. With that being said, we can definitely look into use a different symbol for the in the future.

W.R.T the tree statistics, I totally agree that counting the root at a tip is confusing. This is a result of how a bp tree counts tips (i.e. an open paren followed by a closed paren) and is not necessarily a bug since it would technically be a leaf node. :thinking: We should probably add some additional logic to account for this case. What do you think would make the most sense: just showing 'N/A' for all categories in tree statistics or counting it as an internal node?

empress/support_files/js/shearer.js
```diff
@@ -0,0 +1,418 @@
...
+            var opt = document.createElement("option");
+            opt.innerText = col;
+            opt.value = col;
+            scope.shearSelect.appendChild(opt);
+        });
+
+        this.addLayerButton.onclick = function () {
+            scope.controller.addLayer(scope.shearSelect.value);
+            scope.shearSelect[scope.shearSelect.selectedIndex].remove();
+        };
```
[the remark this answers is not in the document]
kwcantrell: agree, the 'shear by...' Select menu and 'Add shear filter' button are now hidden when the select menu is empty.

empress/support_files/js/empress.js
```diff
@@ -2464,11 +2510,21 @@ define([
...
         var keyInfo = colorer.getMapHex();
 
+        // if the tree has been sheared then categories in obs maybe empty.
+        // getUniqueFeatureMetadataInfo() does not filter out those categories
+        // so that the same color can be assigned to each value in obs.
+        for (var key in keyInfo) {
+            if (uniqueValueToFeatures[key].length === 0) {
+                delete keyInfo[key];
+            }
+        }
```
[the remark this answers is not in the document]
kwcantrell: abstracted this block to 'removeEmptyArrayKeys' in util.js

kwcantrell:
Thanks @fedarko for all you great feedback! If there are no more additional comments, this PR should be good to merge (although it looks like McHelper is still failing :cry:)

empress/support_files/js/shearer.js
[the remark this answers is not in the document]
fedarko: Looks great, thanks!

empress/support_files/js/empress.js
fedarko: I think counting the root as a "leaf" in this case makes sense. I guess there isn't an obvious correct way to handle this case...

For now, I think probably it'd make sense to just add an extra sentence or two to the tree statistics and node circle parts of the menu clarifying what happens in the just-root-left case. I'll go ahead and add suggestions later today.

fedarko (changes requested):
Ok, ~four things left:

1. @kwcantrell Could you add the following text to replace the notes under the node circle section of side-panel.html? Should go within the same `<p>`.

```
    Controls behavior of when to draw circles at nodes' positions.
    Note that, in the rare case when the tree is sheared to only
    contain the root node, a circle will nonetheless always be
    drawn for it to serve as a visual marker (unless "Hide all node
    circles" is selected).
```

2. I also added text about this to the tree stats -- this should be a suggestion in the code

3. Added a suggestion that fixes the docs for the new util function added

4. While doing some testing I noticed a visual bug that happens on collapsing then un-collapsing the tree in the circular / rectangular layouts. It looks like this removes some of the internal node stuff (vertical lines in the rectangular layout / arcs in the circular layout), but not all of them. This seems to impact both sample and feature metadata coloring. See this GIF:

seharthing

It looks like some of the colors from feature metadata coloring are wrong, too? From what I can tell collapsing seems to mess with these (this is around the location of `g__Acanthamoeba`, `s__palestinensis` on the moving pictures tree):

hm

It doesn't seem like these problems are in the master branch -- I think this is something caused by this PR's changes. Not sure what. Sorry, but we should definitely fix this before merging this.

fedarko:
This PR also works very well with the EMPeror integration -- I'm really impressed!

The only thing I really notice offhand is that coloring (e.g. while double-clicking a sample metadata category, or after selecting a group of samples) isn't applied to features that are only revealed midway through the coloring operation. See the `Bacteria` group of features below: these features remain uncolored, although most of the tree should be colored orange --

shearoingoinf

If fixing this is too much of a hassle we can probs add a disclaimer or something, or maybe disable shearing during these sorts of "sensitive" operations (maybe also disabling shearing during animations would be a good idea, just to be on the safe side for now). But up to you.

wasade:
This is super coool!!

kwcantrell:
@fedarko I am glad you caught the visual bug (4). I had changed the way `empress.resetTree()` worked so that it would only call `getTreeCoords` if the tree was collapsed (to save a small amount of time). I guess the way we cache the collapsed clades it wound up causing some odd visual bugs.

I also registered a shear observer for the emperor callbacks so that they get updated when a user shears the tree. The observers will also be unregistered after the 4 second timer expires. 

w.r.t. the animations, its get a bit hairy to properly deal with the shear functionality during an animation so instead I added a warning message to the animation panel.

empress/support_files/js/emperor-callbacks.js
```diff
@@ -64,15 +64,12 @@ plotView.on("click", function (name, object) {
 });
 
 plotView.on("select", function (samples, view) {
+    // remove any strangling observers
```
fedarko, line 67:
```suggestion
    // remove any ongoing observers
```
... I don't think "strangling" is the correct word for this, but please correct me if i'm wrong!

empress/support_files/js/emperor-callbacks.js
```diff
@@ -83,9 +80,23 @@ plotView.on("select", function (samples, view) {
...
+
+    var colorEmpress = () => {
+        // if there's any coloring setup remove it, and re-enable the update button
+        sPanel.sUpdateBtn.classList.remove("hidden");
+        sPanel.fUpdateBtn.classList.remove("hidden");
+        empress.clearLegend();
+        empress.resetTree();
+        empress.colorSampleGroups(namesOnly);
+    };
+    colorEmpress();
```
fedarko, line 92: Looks like this code is duplicated within this file -- would it be possible to turn this into a function that takes the `colorSampleGroups` parameter as its parameter, so that this code could only be defined once but used twice?

empress/support_files/js/emperor-callbacks.js
```diff
@@ -164,6 +176,9 @@ ec.controllers.animations.addEventListener("animation-ended", function (
 ec.controllers.color.addEventListener("value-double-clicked", function (
     payload
 ) {
+    // remove any strangling observers
```
fedarko, line 179:
```suggestion
    // remove any ongoing observers
```

empress/support_files/templates/side-panel.html
```diff
@@ -300,6 +300,10 @@
     (specified by the trajectory) at different timepoints throughout a
     gradient.
   </p>
+  <p class="side-panel-notes">
+    <b>Warning:</b> Do not shear the tree while an animation is running. Please
+    stop and restart if the tree was sheared during an animation.
+  </p>
```
fedarko, line 306: Thanks for adding this -- looks good!

If we are going to completely say that the user shouldn't do shearing during an animation, I think it might make sense to just explicitly prevent this -- I assume this warning would also apply to synchronized animations that Emperor starts?

Maybe we could have the `Animator` class update the `Empress` class somehow (so that the `Empress` class always has knowledge of if an animation is ongoing), and then `ShearModel` could use that knowledge to only apply shearing changes when an animation isn't ongoing. Or, IDK, you would know best how to do that :)

...We don't have to do this before merging, tho -- I recognize that this PR has been waiting for a while, so I'm ok if we want to get this in and then take care of _preventing_ this case later.

fedarko (commented):
A few updates based on the changes made. Sorry to be a pain and thanks so much Kalen!

empress/support_files/js/emperor-callbacks.js
kwcantrell: Naming things is the bane of my existence lol. I actually  meant `dangling observer` :laughing: My logic was as follows:
Every time samples are selected or values are clicked on in emperor the callback functions are called. Thus, if a value is clicked on before the timeout event is called, the new observer will supersede the old observer and the old observer can be thought of as being de-referenced  hence a dangling observer.

However, I like `ongoing` much better since the old observer is not actually de-referenced and will still be called unless explicitly unregistered.

empress/support_files/templates/side-panel.html
kwcantrell: I added a bunch of code to disable **sample/feature metadata coloring, barplots, and shear** while an animation is active. This is how the UI looks while an animation is active:
Screenshot from 2021-06-10 23-48-23

Additionally,  there was another case where the user started the animation using the emperor interface. In that case, I also disable the `Animation` tab within empress.
Screenshot from 2021-06-10 23-51-13

kwcantrell:
@fedarko I went ahead and added a bunch of code to disable **sample/feature metadata coloring, barplots, and shear tree** while an animation is active. Since I added some new code, I am sure you can continue to be "pain" :laughing: (please continue to be a "pain" as that will ultimately lead to a better feature).  This is how the UI looks while an animation is active:
Screenshot from 2021-06-10 23-48-23

Additionally,  there was another case where the user started the animation using the emperor interface. In that case, I also disable the `Animation` tab within empress.
Screenshot from 2021-06-10 23-51-13

empress/support_files/js/enable-disable-animation-tab.js
```diff
@@ -0,0 +1,50 @@
...
+     */
+    function EnableDisableAnimationTab(tab) {
+        // call EnableDisableTab constructor
+        EnableDisableTab.call(this, tab);
+
+        // add disable text message
+        this.disableContainer.innerHTML =
+            '<p class="side-panel-notes">' +
+            '<span style="font-weight: bold;">This tab</span> ' +
+            "has been disabled while the animation is active from within " +
```
fedarko, line 25:
```suggestion
            "has been disabled while an animation is active within " +
```

empress/support_files/js/enable-disable-animation-tab.js
```diff
@@ -0,0 +1,50 @@
...
+        // add instructions to re-enable the tab
+        this.disableContainer.innerHTML +=
+            '<p class="side-panel-notes">' +
+            '<span style="font-weight: bold;">To re-enable</span> ' +
+            "this tab please click on the " +
+            '<span style="font-style: italic;">Reset the animation</span>' +
+            " button located within the " +
+            '<span style="font-style: italic;">Animations</span> ' +
+            "tab of Emperor." +
+            "</p>";
```
fedarko, line 38:
```suggestion
            '<p class="side-panel-notes">' +
            '<span style="font-weight: bold;">To re-enable this tab,</span> ' +
            'please click on the "Reset the animation" button located within ' +
            'the "Animations" tab of Emperor.' +
            "</p>";
```

fedarko (commented):
Thank you so much, @kwcantrell! I'm really excited by these changes -- this is very comprehensive. I think this PR will actually also close https://github.com/biocore/empress/issues/195, an issue that it turns out we opened exactly one year ago (!!). So that's neat.

All I have at this point are minor suggestions. One more round and I think we can merge this in!!!

It isn't necessary by any means, but I also think we can reduce the size of the warning messages for the disabled tabs a bit -- we can do this by adding another flag variable, `is_empire_plot`, that is passed from `core.py` to `empress-template.html` to the `EnableDisableSidePanelTab` objects. This flag variable will control whether or not the note about Emperor animations will be shown in e.g. the sample/feature metadata coloring tabs while they're disabled (so, if there isn't an Empire plot, then we can just show a smaller warning). If you're ok with it, I propose the following changes to three files within the code:

[3-modified-files.zip](https://github.com/biocore/empress/files/6697794/3-modified-files.zip)

I've tested this on a checked-out version of this branch, and I think this should work. The differences between these three files and the repo can be assessed by copying them in and using `git status` (note that my other suggestions in this review should be applied on top of these changed files).

Thanks again, and let me know if I can be of further help.

empress/support_files/templates/empress-template.html
```diff
@@ -123,19 +128,62 @@ <h3 id="menu-sm-header">Sample Presence Information</h3>
...
+                },
+                {
+                    tabName: "Feature Metadata Coloring",
+                    container: document.getElementById(
+                        "feature-metadata-coloring-div")
+                },
+                {
+                    tabName: "Barplots",
+                    container: document.getElementById("barplot-div")
+                },
```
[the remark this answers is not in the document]
kwcantrell: good point. I removed the barplots tab.

kwcantrell:
Thanks for the many reviews @fedarko! :laughing: This feature went through a lot of iterations and I think the final product is way more polished as a result.

fedarko:
Thanks so much @kwcantrell (and @ElDeveloper)! It's really great to finally have this in :)

At long last, I guess we can push out a release early next week? :eyes: :eyes: :eyes:

kwcantrell:
Its definitely time for a release. Although, if we can merge #484 , #521 (updated to include #484) and the new adjust color alpha feature first, then I think we will truly have the "next gen" release! :astonished: 

I believe #484 and #521 just have a few issues left to address and the adjust color alpha is ready for review (I was just waiting for #505 to be merge before I opened the PR as it abstracts/builds on #505).

However, I am totally fine with cutting a release next week!

fedarko:
The idea of Next-Generation EMPress :tm: certainly sounds enticing ;) However, I'd vote we cut a release ASAP, just because the review process can sometimes take a while (... as I guess was demonstrated by me on this PR :sweat_smile: ), and it's been some time since our last release.

We could always cut one release next week, and then cut another release soon afterwards with #484 / #521 / alpha adjustment / etc. I have a few features in prep that I hope to get ready soon as well (#322, #371, #504).

kwcantrell:
Okay, that sounds like a plan then! Lets to cut a release next week!

PR #512, opened 2021-05-01
Automatically resize main legend when changing tree coloring

fedarko:
The main legend in the application is resizable, and its size will automatically change as the tree is colored by different categories. However, this behavior currently stops working after the user manually resizes the legend at all:

leg

We should keep the ability for users to manually resize the legend, IMO, since the legend can get really large and we don't want it to obstruct the user's view.

It looks like the reason for this is that, when the user manually resizes the legend, that the `width` / `height` attributes of the legend go from being unspecified to being fixed: (note the `element.style` text in the dev console on the right)

wh

This PR fixes this, by resetting the width/height of the main legend to their defaults right before we adjust tree coloring:

Peek 2021-04-30 20-53

It's worth noting that this PR will need to be updated once #484 is merged in so that `Empress.resizeLegend()` is called before creating a continuous legend.

### DIFF
--- a/empress/support_files/css/empress.css
+++ b/empress/support_files/css/empress.css
@@ -594,6 +594,12 @@ p.side-header button:hover,
 
 /* The legend is resizable, thanks to resize: both; and overflow: auto.
  * See https://stackoverflow.com/a/61976603/10730311.
+ *
+ * The 0.1px bottom padding prevents an unnecessary vertical scrollbar
+ * that kept showing up when setting the legend's width/height back to
+ * their defaults ("") in empress.js. It seems like having a tiny bit
+ * of extra vertical space is needed to get rid of this. It's barely
+ * visible to the user so it shouldn't be a problem.
  */
 #legend-main {
     margin: 20px;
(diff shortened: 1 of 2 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F512%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F512%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F512%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F512%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F512%2Fplain.qzv)

kwcantrell:
Thanks @fedarko, this looks good! I will go ahead and merge this and update #484.

Issue #513, opened 2021-05-01
Replace high-level <p> tags with <div>


fedarko:
One weird thing I noticed in the side panel HTML is that there are hidden `<p>` tags that take up some space at the bottom of the sample / feature metadata coloring panels:

![huh](https://user-images.githubusercontent.com/4177727/116770945-5ce07a00-a9fc-11eb-859d-2821f132f1bc.png)

While trying to figure out the cause of this I ran the side panel HTML through prettier (currently prettier just works on the JS/CSS, but it supports HTML). prettier gave the following error:

```
[error] empress/support_files/templates/side-panel.html: SyntaxError: Unexpected closing tag "p". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags (62:3)
[error]   60 |       </p>
[error]   61 |     </div>
[error] > 62 |   </p>
[error]      |   ^
[error]   63 | </div>
[error]   64 | 
[error]   65 | <!-- Feature Metadata Coloring Options -->
```

It looks like the reason for this (https://github.com/prettier/prettier/issues/5833) is that putting other `<p>` tags, or `<div>` tags, inside of `<p>` is not allowed per the HTML spec. I suspect this is causing small formatting problems in some browsers, likely including the one I mentioned earlier. (Even if not, we should try to have valid HTML.)

I think the way to address this is change the top-level `<p>` tags to `<div>`, make sure they're styled as they were before (or close enough), and then re-run the HTML through prettier. I think this should fix things?

PR #514, opened 2021-05-01
Small side panel formatting fix: hide button <p> containers instead of buttons; fix prev/next frame animation buttons

fedarko:
This doesn't entirely fix the problems brought up in #513, but it is a start -- it removes some unnecessary whitespace in the sample/feature metadata coloring and animation panels.

Hiding the button(s) inside a `<p>` tag still leaves some extra whitespace, since empty `<p>` tags still take up some space. Hiding the `<p>` tag directly (instead of the button(s)) fixes this, so now for example the default state of the animation panel looks a bit nicer:

| previous | now |
| -- | -- |
| image | image |

This PR also adjusts the animation panel so that:

- disabled buttons (prev frame / next frame, when on the first/last frame) are now given a special style to indicate they're disabled
- the next frame button is correctly disabled when on the last frame. It was off by 1 before, apparently.

| prev frame button disabled | next frame button disabled |
| -- | -- |
| image | image |


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F514%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F514%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F514%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F514%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F514%2Fplain.qzv)

ElDeveloper:
@fedarko can you pull from upstream to see the mchelper build?

empress/support_files/css/empress.css
```diff
@@ -501,6 +501,11 @@ p.side-header button:hover,
     background-color: dimgray;
 }
 
+.control p button:disabled {
+    background-color: #ccc;
+    cursor: not-allowed;
+}
+
```
ElDeveloper, line 508:
```suggestion
```

empress/support_files/css/empress.css
```diff
@@ -501,6 +501,10 @@ p.side-header button:hover,
     background-color: dimgray;
 }
 
+.control p button:disabled {
+    background-color: #ccc;
+    cursor: not-allowed;
+}
```
ElDeveloper, line 507:
```suggestion
}

```

ElDeveloper:
I added/removed a suggestion so that triggered a new build.

empress/support_files/css/empress.css
```diff
@@ -501,6 +501,11 @@ p.side-header button:hover,
     background-color: dimgray;
 }
 
+.control p button:disabled {
```
ElDeveloper, line 504: This is slightly unrelated to this PR but would you be open to changing the name of this class to be `empress-control` or something that namespaces the class a bit better? Seems to me that `control` is too generic and prone to a collision with other CSS files.

ElDeveloper (approved):
All looks good other than one minor comment that if you are comfortable can be addressed here or in a separate PR.

empress/support_files/css/empress.css
fedarko: I agree that this name should be made more specific -- although I vote we move that to another PR. Making that change will require changing a lot of lines (see below), and I'd like to follow that up with a decent amount of manual testing to make sure it doesn't break something (just find-and-replacing `control` will cause problems, so the changes will probably have to be done manually...)

```
$ grep -ri " control " empress/support_files/templates/*
empress/support_files/templates/empress-template.html:        <!-- show control panel button -->
empress/support_files/templates/empress-template.html:        <button id="show-ctrl" title="Show control panel" class="hidden unselectable-text">&#9699;</button>
empress/support_files/templates/side-panel.html:    <button id="hide-ctrl" title="Hide control panel"
empress/support_files/templates/side-panel.html:<div class="side-content control hidden" id="sample-metadata-coloring-div">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden" id="feature-metadata-coloring-div">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden" id="barplot-div">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden" id="shear-div">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden" id="animation-div">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden" id="export-div">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden">
empress/support_files/templates/side-panel.html:<div class="side-content control hidden">
```
```
$ grep -ri "\.control" empress/support_files/css/empress.css
.control {
.control p {
.control p label:first-of-type {
.control p button {
.control p button:hover {
.control p button:disabled {
.control p button:only-child {
.control div.animate-btns {
 * Also, the reason for display: block is because the .control p CSS
 ```

ElDeveloper:
Sounds fair. Thanks @fedarko !

PR #516, opened 2021-05-21
throttled onwheel event

kwcantrell:
I don't know if this an issue was created for this but I just tried using empress on a mac laptop and the zoom functionality was incredibly sensitive, which caused the tree to zoom in/out incredibly fast. This PR fixes that issue by throttling the .onwheel event to fire at most 1 every 50 milliseconds.

### DIFF
--- a/empress/support_files/js/canvas-events.js
+++ b/empress/support_files/js/canvas-events.js
@@ -220,7 +220,7 @@ define(["underscore", "glMatrix", "SelectedNodeMenu"], function (
 
         canvas.onmousedown = mouseDown;
         canvas.onclick = mouseClick;
-        canvas.onwheel = _.throttle(zoomTree, 50);
+        canvas.onwheel = _.throttle(zoomTree, 10);
         canvas.ondblclick = doubleClick;
     };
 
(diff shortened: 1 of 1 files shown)


kwcantrell:
McHelper is failing to connect (see #517) which is causing the build to fail.

fedarko:
Nice catch, thank you!

Not to split hairs, but would a slightly lower throttle value be ok? On an Ubuntu laptop, 50 ms seems too jerky for me (compared to the previous settings) -- 10 ms feels better. (I captured two GIFs comparing the settings, but I'm not sure these will be useful due to frame-rate stuff...)

| 50 ms | 10 ms |
| --- | --- |
| t50 | t10 |

I guess we should ideally try to pick the lowest value that allows Mac users to control things well. That all being said, I suspect that there probably isn't a "perfect" value here, especially since lots of things like touchpad sensitivity are up to the user -- I'm fine choosing a setting that works reasonably well for both of us, and in the future we could make this configurable in the side panel or something if desired.

> McHelper is failing to connect (see #517) which is causing the build to fail.

Weird -- not sure why this is happening. In any case, this PR looks fine, so I'm happy to merge it in when we agree on a value.

kwcantrell:
agreed @fedarko, ideally there should be some 'sensitivity' adjuster. I lowered the throttle to 10. I tried lower numbers (as I felt my normal laptop worked perfect before the throttle) but it seemed a bit fast for mac. So, 10 seems to be the optimal number to make everyone happy.

fedarko:
Thanks @kwcantrell!

PR #519, opened 2021-05-27
Use toastr to support showing multiple toast messages at once, with a fancy UI

### LINKED ISSUE #286: Support showing multiple toast messages at once
Technically a bug, but IMO more of a feature request.

Some of the barplot code involves showing errors to the user if certain settings are invalid for a barplot layer (e.g. the user tries to do numeric scaling with a non-numeric field, or the user sets the maximum bar length to have a value less than the minimum bar length). I figured it'd be nice to show these errors to the user via `util.toastMsg()`.

However, there are two problems I've run into with this approach:

1. Only one error message can be shown at a time, since there is just a single `#toast` HTML element being used.

2. The use of `setTimeout()` to re-hide the `#toast` element means that, if multiple error messages are triggered within a short timespan, not all of them will be shown for an equal time -- the first error message's timeout will cause "later" error messages to be hidden early. (This is the same sort of problem as what causes #197.) See the end of this post for a rough "log" and GIF of what this sort of situation looks like.

It would be nice to develop the toast functionality a bit so that multiple messages can be shown simultaneously without bonking into each other. This _could_ involve making `util.toastMsg()` more complex, but I think __probably the easiest course of action for everyone would just be using an established third-party library for showing toasts__: one of the following should be more than sufficient --

- [Bootstrap's toast functionality](https://getbootstrap.com/docs/4.3/components/toasts/)
- [toastr](https://codeseven.github.io/toastr/demo.html)
- [jQuery toast plugin](https://kamranahmed.info/toast)

Thanks!

### Appendix: what timeouts bonking into each other looks like
```
0 seconds in: user triggers error A, a toast message appears with a duration of 5000 ms (aka 5 seconds)
1 seconds in: toast A error message is visible
2 seconds in: toast A error message is visible
3 seconds in: user triggers error B, a toast message appears with a duration of 5000 ms
4 seconds in: toast B error message is visible
5 seconds in: error A's timeout activates and the toast is hidden, even though error message B was only shown for ~2 seconds
```

toast

fedarko:
This closes #286. It adds the [toastr](https://codeseven.github.io/toastr/) JS/CSS dependency, which is called from EMPress' existing `util.toastMsg()` function.

It is now possible for many toast messages to be active at once, which make the general "error message" UI much more pleasant: (note that in this GIF I'm clicking on the buttons causing each error a bunch of times, normally the user will just see at most 1-2 toast messages at a time)

toast

It's also possible to show different "types" of toast messages, e.g. distinguishing errors from successes from warnings. Below is what an `Animation complete.` message, which we classify as `success`, looks like:

Peek 2021-05-26 21-09

This GIF also shows that it's possible to dismiss a toast by clicking on it, which is a nice feature to have.

Lastly (and unrelated to the above stuff), this PR changes the `Main CI` title to be `Main CI + McHelper`. This makes it clearer that the cause for this CI failing (as is currently shown on the README) is McHelper, not the actual CI (all the actual tests/checks pass) -- not a big deal or anything, but this hopefully gives users some confidence that our tests aren't entirely broken ;)


kwcantrell:
This is awesome @fedarko! Distinguishing "error" messages from "success" messages is a really nice touch.

Since this PR modifies the CI, do you think it would make sense to just put  McHelper in a separate action? That way 'Main CI' will still pass if McHelper fails.

fedarko:
Thanks @kwcantrell! I had a similar thought -- however, I didn't do this, since I thought it'd make sense to only upload stuff to McHelper when the Main CI passes (and we know that at least the tests are passing). This would avoid polluting McHelper with a bunch of broken QZVs.

...But thinking about it more, it doesn't have to be an all-or-nothing thing though -- it looks like it should be possible to configure GitHub actions to [only run the separate McHelper action when the Main CI action (and maybe the standalone CI too?) passes](https://stackoverflow.com/a/65698892). I think that might require a decent amount of messing around to get everything working right -- could we make that a separate issue for now?

kwcantrell:
We can move this to a different issue. In the meantime, how about we just comment out McHelper?

kwcantrell:
Thanks @fedarko!

PR #525, opened 2021-06-09
Improves error handling when "shear" but no metadata passed

ChrisKeefe:
When `tree_plot` is run with `shear-to-feature-metadata`, but no metadata is passed, the following error is produced.
image

This PR explicitly handles that case, improving the error message:
image

### DIFF
--- a/empress/core.py
+++ b/empress/core.py
@@ -136,6 +136,11 @@ class Empress():
             # self._validate_and_match_data()
             self.features = feature_metadata.copy()
         else:
+            if shear_to_feature_metadata:
+                raise ValueError(
+                    "Feature metadata must be provided in order to shear "
+                    "to feature metadata."
+                )
             self.features = None
 
         self.ordination = ordination
@@ -224,12 +229,7 @@ class Empress():
 
         else:
             if shear_to_feature_metadata:
-                try:
-                    features = set(self.features.index)
-                except AttributeError:
-                    raise ValueError(
-                        "Feature metadata must be provided in order to shear "
-                        "to feature metadata.")
+                features = set(self.features.index)
                 all_tips = set(bp_tree_tips(self.tree))
                 # check that feature metadata contains at least 1 tip
                 if not features.intersection(all_tips):
--- a/empress/core.py
+++ b/empress/core.py
@@ -224,7 +224,12 @@ class Empress():
 
         else:
             if shear_to_feature_metadata:
-                features = set(self.features.index)
+                try:
+                    features = set(self.features.index)
+                except AttributeError:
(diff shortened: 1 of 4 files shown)


ChrisKeefe:
Fixed the linter error I introduced, but left the other two alone to keep the diff minimal. I can fix thes other two here if you like.
image

empress/core.py
```diff
@@ -224,7 +224,12 @@ def _validate_and_match_data(self, ignore_missing_samples,
 
         else:
             if shear_to_feature_metadata:
-                features = set(self.features.index)
+                try:
+                    features = set(self.features.index)
+                except AttributeError:
+                    raise ValueError(
+                        "Feature metadata must be provided in order to shear "
+                        "to feature metadata.")
```
fedarko, line 232: Thank you, this is great! This definitely fixes the problem. I think the code might read a bit more clearly if we trigger this error earlier -- I propose we move this error check up to right around line 139 in `core.py`, so that we can bail out as soon as we see that the feature metadata is `None`. I think the updated code will read something like: 

```python
        if feature_metadata is not None:
            # this will be transformed into self.tip_md and self.int_md in
            # self._validate_and_match_data()
            self.features = feature_metadata.copy()
        else:
            if shear_to_feature_metadata:
                raise ValueError(
                    "Feature metadata must be provided in order to shear "
                    "to feature metadata."
                )
            self.features = None
```

... and this way we won't need to use the try/except block to catch this error.

Please let me know what you think, though!

fedarko (commented):
Thank you @ChrisKeefe! I have one small comment -- I think we could catch this error a bit earlier, but in any case this is really useful to have fixed.

I am not sure about those linting errors -- from what I can tell, your code passes all of the linting stuff on my system (running `make stylecheck` in the root of the repository), and it seems now to be passing in our CI. Maybe our flake8 versions are slightly different? (Looks like flake8 v3.5.0 and v3.9.2 both accept this PR's code without a problem now.) Not a big deal, though!

Thanks again! :deciduous_tree:

empress/core.py
ChrisKeefe: :laughing: I was just running flake8 directly from within `empress/`. Defaults ftw!
Makefile updated, whitespace linted, and error check moved up. Thanks for the guidance.

kwcantrell:
Thanks @ChrisKeefe for catching/fixing the `shear-to-feature-metadata` bug as well as discovering/fixing the lint issue(s)! This a two-for-one PR!  :100::1st_place_medal: :100: 

Since you are a first time contributor, I will go ahead an approve the workflows so that our CI can run and after that I this should be good to merge into master.

PR #530, opened 2021-07-08
added flag to make-dev-page so that only tree plot is created

kwcantrell:
This PR adds a `--no-empire` flag to the `make-dev-page` script which will result in a tree plot being created instead of an empire plot.

PR #537, opened 2021-07-23
🐧🚑 add mchelper back to GitHub actions

### LINKED ISSUE #517: McHelper is failing to connect

ElDeveloper:
Looks like the fileserver was down. I was able to restart it and things
seem to be working locally.

Fixes #517


ElDeveloper:
Not sure what is going on but I can't see the `Main CI + Helper` action being run anywhere. Am I missing something?

fedarko:
That is weird -- it is missing from my perspective, also. #536 and #530 both show "5 checks" (the 4 standalone CI runs for different versions of Python, plus `Main CI + McHelper`) instead of 4 checks... and the master branch also is running 5 checks.

I'll wait to make sure the CI completes on the master branch, and then I think I'll just merge this in -- maybe GH Actions is just being strange right now.

PR #545, opened 2021-08-12
sorted all menus

kwcantrell:
This sorts all drop down menus and the node select menu. This should hopefully make it much easier to find things :smile:

### DIFF
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -3793,10 +3793,10 @@ define([
     };
 
     /**
-     * Returns the col variable value for the node.
+     * Returns the feature metadata value at a f. m. column for a node.
      *
      * @param {Number} node Postorder position of a node in the tree.
-     * @param {String} col The name of the feature metadata column.
+     * @param {String} col Name of a feature metadata column.
      *
      * @return {String} The col variable value for the node or undefined if
      *                  no value exists.
@@ -3813,11 +3813,11 @@ define([
     };
 
     /**
-     * Checks to see if node has feature metadata.
+     * Checks to see if a node has feature metadata.
      *
      * @param {Number} node Postorder position of a node in the tree.
      *
-     * @return true if node has feature metadata else false.
+     * @return true if node has feature metadata, false otherwise.
      */
     Empress.prototype.hasFeatureMetadata = function (node) {
         return _.has(this._tipMetadata, node) || _.has(this._intMetadata, node);
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -2712,7 +2712,7 @@ define([
     };
 
     /**
-     * Returns a list of sample categories.
+     * Returns a sorted list of sample categories.
      *
      * If this.isCommunityPlot is false (no table / sample metadata were
      * provided), this just returns [].
@@ -2721,7 +2721,7 @@ define([
      */
     Empress.prototype.getSampleCategories = function () {
         if (this.isCommunityPlot) {
-            return this._biom.getSampleCategories();
+            return util.naturalSort(this._biom.getSampleCategories());
         } else {
             return [];
         }
@@ -2849,12 +2849,12 @@ define([
     };
 
     /**
-     * Returns an array of feature metadata column names.
+     * Returns a sorted list of feature metadata column names.
      *
      * @return {Array}
      */
     Empress.prototype.getFeatureMetadataCategories = function () {
-        return this._featureMetadataColumns;
+        return util.naturalSort(this._featureMetadataColumns);
     };
 
     /**
@@ -3790,6 +3790,40 @@ define([
         this.getLayoutInfo();
 
         this.redrawBarPlotsToMatchLayout();
+    };
+
+    /**
+     * Returns the col variable value for the node.
+     *
+     * @param {Number} node Postorder position of a node in the tree.
+     * @param {String} col The name of the feature metadata column.
+     *
+     * @return {String} The col variable value for the node or undefined if
+     *                  no value exists.
+     */
+    Empress.prototype.getNodeFeatureMetadataValue = function (node, col) {
+        var colIndx = _.indexOf(this._featureMetadataColumns, col);
+        if (_.has(this._tipMetadata, node)) {
+            return this._tipMetadata[node][colIndx];
+        }
+        if (_.has(this._intMetadata, node)) {
+            return this._intMetadata[node][colIndx];
+        }
+        return undefined;
+    };
+
+    /**
+     * Checks to see if node has feature metadata.
+     *
+     * @param {Number} node Postorder position of a node in the tree.
+     *
+     * @return true if node has feature metadata else false.
+     */
+    Empress.prototype.hasFeatureMetadata = function (node) {
+        if (_.has(this._tipMetadata, node) || _.has(this._intMetadata, node)) {
+            return true;
+        }
+        return false;
     };
 
     return Empress;
--- a/empress/support_files/js/empress.js
+++ b/empress/support_files/js/empress.js
@@ -3820,10 +3820,7 @@ define([
      * @return true if node has feature metadata else false.
      */
     Empress.prototype.hasFeatureMetadata = function (node) {
-        if (_.has(this._tipMetadata, node) || _.has(this._intMetadata, node)) {
-            return true;
-        }
-        return false;
+        return _.has(this._tipMetadata, node) || _.has(this._intMetadata, node);
     };
 
     return Empress;
--- a/empress/support_files/js/select-node-menu.js
+++ b/empress/support_files/js/select-node-menu.js
@@ -288,23 +288,17 @@ define(["underscore", "util"], function (_, util) {
     ) {
         if (this.hasFeatureMetadata) {
             this.fmTable.innerHTML = "";
-            var mdObj;
-            if (tipOrInt === "tip") {
-                mdObj = this.empress._tipMetadata;
-            } else if (tipOrInt === "int") {
-                mdObj = this.empress._intMetadata;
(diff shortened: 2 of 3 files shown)


emperor-helper:
The following artifacts were built for this PR: [empire-biplot.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F545%2Fempire-biplot.qzv), [empire.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F545%2Fempire.qzv), [empress-tree.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F545%2Fempress-tree.qzv), [just-fm.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F545%2Fjust-fm.qzv), [plain.qzv](https://view.qiime2.org/visualization/?type=html&src=https%3A%2F%2Fmchelper.ucsd.edu%2Fdownloads%2Fempress%2F545%2Fplain.qzv)

empress/support_files/js/select-node-menu.js
```diff
@@ -288,23 +288,17 @@ define(["underscore", "util"], function (_, util) {
...
-            if (tipOrInt === "tip") {
-                mdObj = this.empress._tipMetadata;
-            } else if (tipOrInt === "int") {
-                mdObj = this.empress._intMetadata;
-            } else {
-                throw new Error("Invalid tipOrInt value: " + tipOrInt);
-            }
-            if (_.has(mdObj, nodeName)) {
+            if (this.empress.hasFeatureMetadata(nodeName)) {
                 var headerRow = this.fmTable.insertRow(-1);
```
fedarko, line 292: It doesn't look like the modified code considers `tipOrInt` any more. Is this due to `nodeName` no longer being the node name but actually just being the postorder position of a node in the tree (in which case there is no longer any ambiguity, since each position should be unique relative to all other nodes)? If so, we should update this function / its documentation, and probably other stuff in this module that still assumes nodes are referred to by their names. (That change is not really in scope of this PR, though -- I'm fine with delegating that to another issue, since I imagine doing it right will be a huge pain.)

tests/test-barplots.js
```diff
@@ -6,14 +6,16 @@ require([
...
 ], function (
     $,
     _,
     spectrum,
     Empress,
     BarplotLayer,
     UtilitiesForTesting,
-    Colorer
+    Colorer,
+    util
```
fedarko: Doesn't look like `util` is used anywhere in this file (I don't think the use of `deepEqual` later on in this file's changes depends on `util` being included, since IIRC `deepEqual` comes from QUnit). Could we remove it?

empress/support_files/js/empress.js
```diff
@@ -3792,5 +3792,39 @@ define([
...
+     *
+     * @param {Number} node Postorder position of a node in the tree.
+     *
+     * @return true if node has feature metadata else false.
+     */
+    Empress.prototype.hasFeatureMetadata = function (node) {
+        if (_.has(this._tipMetadata, node) || _.has(this._intMetadata, node)) {
+            return true;
+        }
+        return false;
```
fedarko:
```suggestion
        return _.has(this._tipMetadata, node) || _.has(this._intMetadata, node);
```

The current way this function is written is perfectly fine -- this suggestion just shortens it. Feel free to ignore this suggestion if you'd like :)

empress/support_files/js/empress.js
```diff
@@ -3792,5 +3792,39 @@ define([
...
+            return this._tipMetadata[node][colIndx];
+        }
+        if (_.has(this._intMetadata, node)) {
+            return this._intMetadata[node][colIndx];
+        }
+        return undefined;
+    };
+
+    /**
+     * Checks to see if node has feature metadata.
```
fedarko:
```suggestion
     * Checks to see if a node has feature metadata.
```

empress/support_files/js/empress.js
```diff
@@ -3792,5 +3792,39 @@ define([
...
+     *
+     * @param {Number} node Postorder position of a node in the tree.
+     *
+     * @return true if node has feature metadata else false.
```
fedarko:
```suggestion
     * @return true if node has feature metadata, false otherwise.
```

empress/support_files/js/empress.js
```diff
@@ -3792,5 +3792,39 @@ define([
         this.redrawBarPlotsToMatchLayout();
     };
 
+    /**
+     * Returns the col variable value for the node.
```
fedarko:
```suggestion
     * Returns the feature metadata value at a f. m. column for a node.
```

empress/support_files/js/empress.js
```diff
@@ -3792,5 +3792,39 @@ define([
...
+     *
+     * @param {Number} node Postorder position of a node in the tree.
+     * @param {String} col The name of the feature metadata column.
```
fedarko:
```suggestion
     * @param {String} col Name of a feature metadata column.
```

empress/support_files/js/empress.js
```diff
@@ -3792,5 +3792,39 @@ define([
...
+     */
+    Empress.prototype.getNodeFeatureMetadataValue = function (node, col) {
+        var colIndx = _.indexOf(this._featureMetadataColumns, col);
+        if (_.has(this._tipMetadata, node)) {
+            return this._tipMetadata[node][colIndx];
+        }
+        if (_.has(this._intMetadata, node)) {
+            return this._intMetadata[node][colIndx];
+        }
+        return undefined;
```
fedarko, line 3812: Wait, should this ever happen in practice? At least in the context of `makeFeatureMetadataTable()` I don't think it should, since we've already called `Empress.hasFeatureMetadata()`. I vote we throw an error here instead, unless I am missing something.

fedarko (commented):
Thanks @kwcantrell! Some of the abstractions done here from `select-node-menu.js` are really nice -- moving this sort of logic out of the node menu and into `empress` is a good idea.

That being said, I am not sure that the actual end goal of this PR, sorting all metadata columns, is always going to be beneficial. For sample metadata columns I can definitely see the benefit, but for feature metadata (esp. when the only feature metadata in use is QIIME 2 taxonomy feature metadata, which I think is a pretty common use case right now) not using the default ordering will make it so that `Confidence` is always the first listed value (as seen on the moving pictures dataset) --

image

... in my opinion this will almost always be inconvenient for users, since showing the taxonomy columns before the other feature metadata is usually what we want. Especially w/r/t taxonomy metadata, the confidence information describes the confidence of the taxonomic annotations, so I think it makes sense to list it after taxonomy rather than beforehand.

That being said, I understand that Emperor already sorts its feature metadata columns when listing them in the biplot coloring menu, so ... if we want to make this change anyway I can deal with it :)

es

kwcantrell:
The main motivation behind sorting everything came up while demonstrating EMPress to the health team during Tuesdays meeting. They have a lot of metadata columns and the current method made it fairly difficult to find things. I think for more general uses, this will be the case. It maybe nice to list confidence after taxonomy but I think that small convenience is outweighed by the possible frustration of having to search for every other category. I could be wrong though :smile:

fedarko:
Fair enough! Since there's a compelling reason for it, I don't mind this change.

> I think that small convenience is outweighed by the possible frustration of having to search for every other category.

Agreed.

We miiiight be able to get the best of both worlds by renaming `Confidence` to `Taxonomy Confidence` or something, which would automatically mean that the column would be listed after the taxonomy levels (the `Level 1` names, etc. are already made up by Empress, so this has precedent). But that fix will cause its own problems and is a discussion for another day :)

ElDeveloper (approved):

empress/support_files/js/select-node-menu.js
kwcantrell: Correct, `tipOrInt` no longer being used. `empress.hasFetaureMetadata` and `empress.getNodeFeatureMetadataValue` replace its functionality. I don't believe ambiguity is an issue since tip names must be unique thus if `nodeName` is in `tipMetadata` then by definition it cannot be in `intMetadata` and vice-versa. Right? 

Assuming I am not looking over anything, I'll remove `tipOrInt` from the doc string.

tests/test-barplots.js
kwcantrell: removed!

empress/support_files/js/empress.js
kwcantrell: Nice! :smile:

empress/support_files/js/empress.js
kwcantrell: We can throw an error. I choose to return undefined in case the calling function wants to do something special in situations where no metadata exists (which I guess can also be handled by a try catch block).

kwcantrell:
Looks like McHelper is failing again... :cry:

kwcantrell:
@fedarko I believe I addressed all your comments.

ElDeveloper:
Did you try rerunning the jobs? I just restarted them. It’s likely an IP related problem(?). 🐧 

Yoshiki Vázquez-Baeza

ElDeveloper:
Yep, artifacts should be updated now.

ElDeveloper:
@fedarko any chance you can review and merge if things look good to you?

fedarko:
@ElDeveloper yep, I will do my best to get to the remaining open pull requests ASAP.

ElDeveloper:
Thanks so much!

Yoshiki Vázquez-Baeza

Issue #553, opened 2021-12-08
Document dependence on binutils


fedarko:
See #552. The root cause of that issue was binutils not being installed before trying to install EMPress (installing `iow` involves running the `ar` command, which is included with binutils).

I suspect that binutils should be installed by default on _most_ Unix systems, so this problem will likely only be an issue for very minimal Linux distributions (e.g. Damn Small Linux, which -- [at least in 2007](http://damnsmalllinux.org/static/act-NW/f-14/t-18441.html) -- didn't include binutils) or for setups like Windows Subsystem for Linux (WSL) (WSL was used in #552).

This information should probably go alongside the information from #550 -- IMO the easiest way to do this would be adding a "Installation FAQ/Troubleshooting" subsection the end of the "Installation & Basic Usage" section of the README.

Issue #550, opened 2021-10-28
Document dependence on Xcode for macOS systems


fedarko:
See #549 -- it looks like installing `iow` requires that Xcode is already installed (at least for macOS systems). It'd be good to add a note about this somewhere in the README, maybe next to the installation instructions or in a new FAQ / troubleshooting section.

PR #555, opened 2022-04-20
Document a few dependencies more clearly in the README, and also work on fixing the build

fedarko:
This PR adds an "installation troubleshooting" section to the README; this addresses #550 and #553. It also adds two short places indicating where users can get support (opening an issue / via the Q2 forum).

**Update:** Multiple tests in the build are failing due to changes in the dependencies' behavior:

1. One of the failures is due to `biom-format` now explicitly raising errors when `Table.update_ids()` is called with a dict containing non-string values (as done in [this test](https://github.com/biocore/empress/blob/bf6f755c5187e543ff13c2c624b153fb97918b23/tests/python/test_tools.py#L143-L145)). This PR fixes this test. I've also submitted [a PR](https://github.com/biocore/biom-format/pull/879) to `biom-format` which clarifies this behavior.

2. Another issue (causing two failing tests) is due to our use of "linear" trees causing problems with iow. This time, the problem is not on our side but on iow's, and I contend that we should keep the linear trees in the tests (if only to make sure that these sort of trees remain explicitly supported). I have submitted [a PR](https://github.com/biocore/improved-octo-waddle/pull/48) to iow which fixes the relevant problem.

Even though this build is currently failing, I imagine that it's probably better to merge this PR in sooner rather than later (I'm aware this is me being a hypocrite considering the backlog of things I need to review, sorry...) -- having the installation documentation up-to-date will prevent user confusion.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,39 @@ and barplots).
 
 Empress is available as either a standalone program or a QIIME 2 plugin. The standalone version will generate a folder with the HTML/JS/CSS files necessary to view the plot while the QIIME 2 version will generate a `.qzv` Visualization that can be viewed on [https://view.qiime2.org/](https://view.qiime2.org/) or by using `qiime tools view`.
 
+### Installation troubleshooting
+
+#### For Windows Subsystem for Linux (WSL) users
+
+**Problem:** Empress' installation process depends on the
+[`binutils`](https://en.wikipedia.org/wiki/GNU_Binutils) package already being
+installed. Although most Unix operating systems come with `binutils` installed
+by default, we've noticed that some Linux distributions / setups (including
+WSL) do not have `binutils` installed.
+
+**Solution:** if, when installing Empress, you encounter long errors
+that include something
+like `ld: cannot find -lbitarr`,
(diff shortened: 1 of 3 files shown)


ElDeveloper:
@fedarko, while the issue in IOW is sorted out, would we be able to pin the version requirement?

fedarko:
@ElDeveloper Just pinned the iow version requirement to 0.1.3 for now. This fixes the build (McHelper is broken which causes the "Main CI + McHelper" build to fail, but all of the installations / tests pass).

I think 0.1.3 is the only iow version that will work with EMPress (see https://pypi.org/project/iow/#history): earlier versions of iow will not have [the updates Erfan added in 2020](https://github.com/biocore/improved-octo-waddle/pull/28), and later versions of iow will have the problematic linear-tree check.

Thanks! :evergreen_tree: :deciduous_tree: :christmas_tree:

ElDeveloper:
Thanks so much @fedarko

fedarko:
Thanks @ElDeveloper!!!

